PR #1, opened 2019-06-20
Manifests to configure IdP and create Users in OpenShift 4

alexeykazakov:
This is more for documentation purposes right now.

### DIFF
--- a/deploy/user.yaml
+++ b/deploy/user.yaml
@@ -1,0 +1,23 @@
+apiVersion: user.openshift.io/v1
+groups: null
+identities: null
+kind: User
+metadata:
+  name: ${USERNAME}
+---
+apiVersion: user.openshift.io/v1
+kind: Identity
+metadata:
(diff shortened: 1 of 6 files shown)


deploy/user.yaml
```diff
@@ -0,0 +1,23 @@
...
+providerUserName: ${USER_ID_FROM_SUB_CLAIM}
+---
+apiVersion: user.openshift.io/v1
+kind: UserIdentityMapping
+metadata:
+  name: rhd:${USER_ID_FROM_SUB_CLAIM}
+user:
+  name: ${USERNAME}
+identity:
+  name: rhd:${USER_ID_FROM_SUB_CLAIM}
```
nurali-techie: what is `USER_ID_FROM_SUB_CLAIM` here ??

alexeykazakov: User ID from the access token issued by the RHD Identity Provider. Should be taken from the “sub” claim from the token.
See https://github.com/codeready-toolchain/member-operator/pull/19/files

PR #4, opened 2019-06-24
feat(CRT-150): client caching controller for clusters represented by KubeFedCluster CRs

MatousJobanek:
This PR brings a controller that watches `KubeFedCluster` CRs and caches clients (created based on the CRs) in a map. These cached clients can be used in both operators `host` and `member` for mutual communication.

Changes introduced in this PR: 
* a new controller that checks  `KubeFedCluster` CRs. The controller is implemented using `NewGenericInformerWithEventHandler` taken from KubeFed project
* a caching mechanism in `cluster.cache.go` that keeps clients of the KubeFed clusters in a map
* a service in `cluster.service.go` that takes care of the transition of `KubeFedCluster` data between the controller and the cache


pkg/cluster/service.go
```diff
@@ -0,0 +1,154 @@
...
+	return s.Log.
+		WithValues("Request.Namespace", cluster.Namespace, "Request.Name", cluster.Name)
+}
+
+func (s *KubeFedClusterService) buildClusterConfig(fedCluster *v1beta1.KubeFedCluster, fedNamespace string) (*rest.Config, error) {
+	clusterName := fedCluster.Name
+
+	apiEndpoint := fedCluster.Spec.APIEndpoint
+	if apiEndpoint == "" {
+		return nil, errors.Errorf("The api endpoint of cluster %s is empty", clusterName)
```
alexeykazakov:
```suggestion
		return nil, errors.Errorf("the api endpoint of cluster %s is empty", clusterName)
```

pkg/cluster/service.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	secretName := fedCluster.Spec.SecretRef.Name
+	if secretName == "" {
+		return nil, errors.Errorf("Cluster %s does not have a secret name", clusterName)
```
alexeykazakov:
```suggestion
		return nil, errors.Errorf("cluster %s does not have a secret name", clusterName)
```

pkg/cluster/service.go
```diff
@@ -0,0 +1,154 @@
...
+	secret := &v1.Secret{}
+	name := types.NamespacedName{Namespace: fedNamespace, Name: secretName}
+	err := s.Client.Get(context.TODO(), name, secret)
+	if err != nil {
+		return nil, err
+	}
+
+	token, tokenFound := secret.Data[util.TokenKey]
+	if !tokenFound || len(token) == 0 {
+		return nil, errors.Errorf("The secret for cluster %s is missing a non-empty value for %q", clusterName, util.TokenKey)
```
alexeykazakov:
```suggestion
		return nil, errors.Errorf("the secret for cluster %s is missing a non-empty value for %q", clusterName, util.TokenKey)
```

pkg/cluster/service_test.go
```diff
@@ -0,0 +1,216 @@
...
+			Namespace: "test-namespace",
+			Labels:    labels,
+		},
+		Status: status,
+	}, secret
+}
+
+func TestAddKubeFedClusterAsMember(t *testing.T) {
+	// given
+	defer gock.OffAll()
```
alexeykazakov:
```suggestion
	defer gock.Off()
```
It's safer to use `gock.Off()` instead. Go runs tests in different packages in parallel by default. Killing all unmatched requests can cause race condition between tests from different packages and can cause test failures.

pkg/cluster/service_test.go
```diff
@@ -0,0 +1,216 @@
...
+				assert.Equal(t, labels["namespace"], fedCluster.OperatorNamespace)
+			}
+			assert.Equal(t, status, *fedCluster.ClusterStatus)
+		})
+	}
+}
+
+func TestAddKubeFedClusterAsHost(t *testing.T) {
+	// given
+	defer gock.OffAll()
```
alexeykazakov:
```suggestion
	defer gock.Off()
```

pkg/cluster/service_test.go
```diff
@@ -0,0 +1,216 @@
...
+				assert.Equal(t, labels["namespace"], fedCluster.OperatorNamespace)
+			}
+			assert.Equal(t, status, *fedCluster.ClusterStatus)
+		})
+	}
+}
+
+func TestAddKubeFedClusterFailsBecauseOfMissingSecret(t *testing.T) {
+	// given
+	defer gock.OffAll()
```
alexeykazakov:
```suggestion
	defer gock.All()
```

pkg/cluster/service_test.go
```diff
@@ -0,0 +1,216 @@
...
+
+	// then
+	fedCluster, ok := cluster.GetFedCluster("east")
+	require.False(t, ok)
+	assert.Nil(t, fedCluster)
+}
+
+func TestAddKubeFedClusterFailsBecauseOfEmptySecret(t *testing.T) {
+	// given
+	defer gock.OffAll()
```
alexeykazakov:
```suggestion
	defer gock.Off()
```

pkg/cluster/service_test.go
```diff
@@ -0,0 +1,216 @@
...
+
+	// then
+	fedCluster, ok := cluster.GetFedCluster("east")
+	require.False(t, ok)
+	assert.Nil(t, fedCluster)
+}
+
+func TestUpdateKubeFedCluster(t *testing.T) {
+	// given
+	defer gock.OffAll()
```
alexeykazakov:
```suggestion
	defer gock.Off()
```

pkg/cluster/service_test.go
```diff
@@ -0,0 +1,216 @@
...
+	fedCluster, ok := cluster.GetFedCluster("east")
+	require.True(t, ok)
+	assert.Equal(t, cluster.Host, fedCluster.Type)
+	assert.Equal(t, "toolchain-host-operator", fedCluster.OperatorNamespace)
+	assert.Equal(t, statusFalse, *fedCluster.ClusterStatus)
+}
+
+func TestDeleteKubeFedCluster(t *testing.T) {
+	// given
+	defer gock.OffAll()
```
alexeykazakov:
```suggestion
	defer gock.Off()
```

pkg/controller/kubefedcluster_caching.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+func StartCachingController(mgr manager.Manager, stopChan <-chan struct{}) error {
+	cntrlName := "controller_kubefedcluster_with_cache"
+	clusterCacheService := cluster.KubeFedClusterService{
+		Client: mgr.GetClient(),
+		Log:    logf.Log.WithName(cntrlName),
+	}
+
+	_, clusterController, err := util.NewGenericInformerWithEventHandler(
```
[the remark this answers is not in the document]
alexeykazakov: After disusing it with Matous it turned out that K8s already takes care of it. When controller starts K8s starts a reconcile cycle for all the primary resources. So, we should be fine here.

pkg/cluster/service.go
```diff
@@ -0,0 +1,154 @@
...
+// it's used for adding/updating/deleting
+type KubeFedClusterService struct {
+	Client client.Client
+	Log    logr.Logger
+}
+
+// AddKubeFedCluster takes the KubeFedCluster CR object,
+// creates FedCluster with a kube client and stores it in a cache
+func (s *KubeFedClusterService) AddKubeFedCluster(obj interface{}) {
+	cluster, ok := obj.(*v1beta1.KubeFedCluster)
```
xcoulon: these first lines are repeated in every "exposed" methods of this service. Maybe you could factorize them?

pkg/cluster/service.go
```diff
@@ -0,0 +1,154 @@
...
+
+	secretName := fedCluster.Spec.SecretRef.Name
+	if secretName == "" {
+		return nil, errors.Errorf("Cluster %s does not have a secret name", clusterName)
+	}
+	secret := &v1.Secret{}
+	name := types.NamespacedName{Namespace: fedNamespace, Name: secretName}
+	err := s.Client.Get(context.TODO(), name, secret)
+	if err != nil {
+		return nil, err
```
xcoulon: do you need to wrap the error with a contextual message?

pkg/cluster/cache_whitebox_test.go
```diff
@@ -0,0 +1,105 @@
...
+	// given
+	defer resetClusterCache()
+	fedCluster := newTestFedCluster("testCluster", v1.ConditionTrue)
+
+	// when
+	clusterCache.addFedCluster(fedCluster)
+
+	// then
+	assert.Len(t, clusterCache.clusters, 1)
+	assert.Equal(t, *fedCluster, *(clusterCache.clusters["testCluster"]))
```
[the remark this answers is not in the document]
MatousJobanek: actually, I can do it... I wanted to verify the actual values, but it's right that testing the pointers directly does the same :-) 0f895fb

pkg/cluster/cache_whitebox_test.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func TestGetCluster(t *testing.T) {
```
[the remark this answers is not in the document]
MatousJobanek: good idea 6335143

pkg/cluster/cache_whitebox_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// when
+	returnedFedCluster, ok := clusterCache.getFedCluster("testCluster")
+
+	// then
+	assert.True(t, ok)
+	assert.Equal(t, *fedCluster, *returnedFedCluster)
+}
+
+func TestExportedGetCluster(t *testing.T) {
```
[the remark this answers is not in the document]
MatousJobanek: well, they are different methods - that's why I created different tests... But I can refactor it slightly - is this better: b192c13 ?

pkg/cluster/service.go
MatousJobanek: I was thinking about it when I was writing it, but I would safe max one line per method comparing to 7 new lines of the new function. That's why I decided to not do that...  But I see that it doesn't look good, so I changed it: 9147ad1

pkg/cluster/service.go
MatousJobanek: done 717dcc1

PR #8, opened 2019-06-26
add makefiles

xcoulon:


make/format.mk
```diff
@@ -0,0 +1,18 @@
+
+GOFORMAT_FILES := $(shell find  . -name '*.go')
```
MatousJobanek: exclude the vendor directory

PR #9, opened 2019-06-26
Add make build

alexeykazakov:
`make build` doesn't give us much since there is no executables provided by this repo. But it might be useful to have some target which we could use for quick compilation error check rather than always relaying on the test execution (which can be longer).

@xcoulon thoughts?


xcoulon:
> `make build` doesn't give us much since there is no executables provided by this repo. But it might be useful to have some target which we could use for quick compilation error check rather than always relaying on the test execution (which can be longer).
> 
> @xcoulon thoughts?

yeah, we can do that indeed

PR #11, opened 2019-06-28
Add helper functions to get host and member clusters

alexeykazakov:
We discussed with Shane some helper/utility functions which will get us access to the Host and Member clusters.
We can't relay on the first cluster or on all available clusters. Because it may happen that some physical cluster acts us a member and as a host. So, such cluster will have mixed types of clusters in the cache.
So, using the existing `GetFirstCluster()` is not safe.

@MatousJobanek @sbryzak, WDYT?

TODO:
- [x] Tests

### DIFF
--- a/pkg/cluster/cache.go
+++ b/pkg/cluster/cache.go
@@ -46,13 +46,16 @@ func (c *kubeFedClusterClients) getFedCluster(name string) (*FedCluster, bool) {
 	return cluster, ok
 }
 
-func (c *kubeFedClusterClients) getFirstFedCluster() (*FedCluster, bool) {
+func (c *kubeFedClusterClients) getFedClustersByType(clusterType Type) []*FedCluster {
 	c.RLock()
 	defer c.RUnlock()
+	clusters := make([]*FedCluster, 0, len(c.clusters))
 	for _, cluster := range c.clusters {
-		return cluster, true
+		if cluster.Type == clusterType {
+			clusters = append(clusters, cluster)
+		}
 	}
-	return nil, false
+	return clusters
 }
 
 // GetFedCluster returns a kube client for the cluster (with the given name) and info if the client exists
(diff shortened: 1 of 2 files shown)


alexeykazakov:
Should have submitted it as a draft... :)

alexeykazakov:
Added tests. Ready for review.

sbryzak:
This PR addresses [this issue](https://jira.coreos.com/browse/CRT-140).

pkg/cluster/cache.go
```diff
@@ -46,23 +46,36 @@ func (c *kubeFedClusterClients) getFedCluster(name string) (*FedCluster, bool) {
...
+func GetHostCluster() (*FedCluster, bool) {
+	clusters := clusterCache.getFedClustersByType(Host)
+	if len(clusters) == 0 {
+		return nil, false
+	}
+	return clusters[0], true
+}
+
+// GetMemberClusters returns the kube clients for the host clusters from the cache of the clusters
+func GetMemberClusters() []*FedCluster {
```
MatousJobanek: Please change the function so it takes a name of the required member cluster and returns the single cluster:
```go
func GetFedCluster(name string) (*FedCluster, bool)
```

MatousJobanek: ...so it is easier to use in host operator - for example:
```go
fedCluster, ok := cluster.GetMemberCluster(account.TargetCluster)
if !ok {
	return fmt.Errorf("the fedCluster %s not found in the registry", account.TargetCluster)
}
```

alexeykazakov: We need a function to get all the available member clusters when we approving/provisioning the user from host. It's for the logic which will check all the cluster's capacity and decide which one to choose to provision the user to.

For the case you mention we can still use the existing `GetFedCluster(name)`, can't we?

MatousJobanek: yeah, I thought that you replaced the function - sorry for confusion... :-)

PR #19, opened 2019-07-12
Update README.adoc

tinakurian:
Adding instructions on how to run and connect host and member operators.

**Please note**: UserProvisionRequest has been renamed to UserSignup in codeready-toolchain/api, however, this name change has not yet been merged in codeready-toolchain/host-operator (PR:  https://github.com/codeready-toolchain/host-operator/pull/27) in order to try out the new steps added to the README.adoc you must use codeready-toolchain/api revision 5b2c0d42be1b085bdeea3fd899b9f5d89692be69 until codeready-toolchain/host-operator#27 has been merged. 

You will also need to do apply the UserProvisionRequest manually for the time being as follows: 
`oc apply -f deploy/crds/toolchain_v1alpha1_userprovisionrequest.yaml`

Relates to: https://jira.coreos.com/browse/CRT-178

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -6,7 +6,7 @@ image:https://codecov.io/gh/codeready-toolchain/toolchain-common/branch/master/g
 
 This repo is for controllers, libs, scripts, make files, etc to be shared between host and member operators.
 
-== Setting Up and Connecting Host and Member Operators
+== Setting Up and Connecting Host and Member Clusters
 
 In a new terminal, navigate to codeready-toolchain/host-operator directory and execute the following commands
 ```
--- a/README.adoc
+++ b/README.adoc
@@ -36,7 +36,7 @@ $ make up-local
 Now that we are setup, we can add resources and watch the logs on the member-operator! To do so, open a new terminal and run the following commands
 ```
 $ oc project toolchain-host-operator
-$ oc apply -f '/path/to/masteruserreord.yaml'
+$ oc apply -f '/path/to/masteruserrecord.yaml'
 $ oc project toolchain-member-operator
 $ oc apply -f '/path/to/useraccount.yaml'
 ```
--- a/README.adoc
+++ b/README.adoc
@@ -13,7 +13,6 @@ In a new terminal, execute the following commands
 $ cd $GOPATH/github.com/codeready-toolchain/host-operator
 $ minishift start --profile host
 $ make up-local
-^C
 ```
 
 ```
@@ -28,8 +27,10 @@ $ make up-local
 
 Now that we are setup, we can add resources and watch the logs on the member-operator! To do so, open a new terminal and run the following commands
 ```
+$ cd $GOPATH/github.com/codeready-toolchain/host-operator
 $ oc project toolchain-host-operator
 $ oc apply -f '/path/to/masteruserrecord.yaml'
+$ cd $GOPATH/github.com/codeready-toolchain/member-operator
 $ oc project toolchain-member-operator
 $ oc apply -f '/path/to/useraccount.yaml'
 ```
--- a/README.adoc
+++ b/README.adoc
@@ -46,6 +46,8 @@ $ oc apply -f '/path/to/useraccount.yaml'
 ```
 
 **Example masteruserrecord.yaml:**
+
+_Note:_ The `targetCluster` field must contain the correct IP address of the member cluster.
 ```
 apiVersion: toolchain.dev.openshift.com/v1alpha1
 kind: MasterUserRecord
--- a/README.adoc
+++ b/README.adoc
@@ -41,6 +41,61 @@ $ oc project toolchain-member-operator
 $ oc apply -f '/path/to/useraccount.yaml'
 ```
 
+**Example masteruserrecord.yaml:**
+```
+apiVersion: toolchain.dev.openshift.com/v1alpha1
+kind: MasterUserRecord
+metadata:
+  name: example
+spec:
+  disabled: false
+  deprovisioned: false
+  userID: 86505192-a386-11e9-ad56-525400ad2b23
+  userAccounts:
+  - targetCluster: member-192-168-42-61-8443
+    syncIndex: 86505192-a386-11e9-ad56-525400ad2b23
+    spec:
+      nsLimit: admin
+      userID: 86505192-a386-11e9-ad56-525400ad2b23
+      nsTemplateSet:
+        tierName: basic
+        namespaces:
+        - type: ide
+          revision: abcdef
+        - type: cicd
+          revision: abcdef
+status:
+  conditions:
+  - type: Ready
+    status: True
+  userAccounts:
+  - targetCluster: target-member-cluster
+    syncIndex: 86505192-a386-11e9-ad56-525400ad2b23
+```
+
+**Example useraccount.yaml:**
+```
+apiVersion: toolchain.dev.openshift.com/v1alpha1
+kind: UserAccount
+metadata:
+  name: example
+spec:
+  userID: 86505192-a386-11e9-ad56-525400ad2b23
+  disabled: false
+  nsLimit: admin
+  nsTemplateSet:
+    tierName: basic
+    namespaces:
+    - type: ide
+      revision: abcdef
+    - type: cicd
+      revision: abcdef
+status:
+  conditions:
+  - type: Ready
+    status: True
+```
+
 == Configuring developers.redhat.com Authentication in OpenShift 4 Cluster
 
 Create a secret with Identity Provider credentials:
--- a/README.adoc
+++ b/README.adoc
@@ -22,6 +22,7 @@ $ minishift start --profile member
 $ make up-local
 ^C
 $ make add-member-to-host
+$ make add-host-to-member
 $ make up-local
 ```
 
--- a/README.adoc
+++ b/README.adoc
@@ -8,24 +8,28 @@ This repo is for controllers, libs, scripts, make files, etc to be shared betwee
 
 == Setting Up and Connecting Host and Member Clusters
 
-In a new terminal, navigate to codeready-toolchain/host-operator directory and execute the following commands
+In a new terminal, execute the following commands
 ```
+$ cd $GOPATH/github.com/codeready-toolchain/host-operator
 $ minishift start --profile toolchain
 $ make up-local
 ^C
 ```
-In the same terminal, navigate to codeready-toolchain/member-operator directory and execute the following commands
+
 ```
+$ cd $GOPATH/github.com/codeready-toolchain/member-operator
 $ make up-local
 ^C
 $ make add-member-to-host
 ```
-In the same terminal, navigate back to codeready-toolchain/host-operator directory and execute the following commands
+
 ```
+$ cd $GOPATH/github.com/codeready-toolchain/host-operator
 make add-host-to-member
 ```
-In the same terminal, navigate back to codeready-toolchain/member-operator directory and execute the following commands
+
 ```
+$ cd $GOPATH/github.com/codeready-toolchain/member-operator
 $ make up-local
 ```
 
--- a/README.adoc
+++ b/README.adoc
@@ -30,7 +30,7 @@ $ make add-host-to-member
 $ make build
 $ make vendor
 $ export OPERATOR_NAMESPACE=toolchain-member-operator
-$ operator-sdk up local --namespace=toolcha-member-operator --verbose
+$ operator-sdk up local --namespace=toolchain-member-operator --verbose
 ```
 
 Now that we are setup, we can add resources and watch the logs on the member-operator! To do so, open a new terminal and run the following commands
--- a/README.adoc
+++ b/README.adoc
@@ -11,25 +11,17 @@ This repo is for controllers, libs, scripts, make files, etc to be shared betwee
 In a new terminal, execute the following commands
 ```
 $ cd $GOPATH/github.com/codeready-toolchain/host-operator
-$ minishift start --profile toolchain
+$ minishift start --profile host
 $ make up-local
 ^C
 ```
 
 ```
 $ cd $GOPATH/github.com/codeready-toolchain/member-operator
+$ minishift start --profile member
 $ make up-local
 ^C
 $ make add-member-to-host
-```
-
-```
-$ cd $GOPATH/github.com/codeready-toolchain/host-operator
-make add-host-to-member
-```
-
-```
-$ cd $GOPATH/github.com/codeready-toolchain/member-operator
 $ make up-local
 ```
 
--- a/README.adoc
+++ b/README.adoc
@@ -47,7 +47,7 @@ $ oc apply -f '/path/to/useraccount.yaml'
 
 **Example masteruserrecord.yaml:**
 
-_Note:_ The `targetCluster` field must contain the correct IP address of the member cluster.
+_Note:_ The `targetCluster` field must contain the correct member cluster name.
 ```
 apiVersion: toolchain.dev.openshift.com/v1alpha1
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -6,6 +6,37 @@ image:https://codecov.io/gh/codeready-toolchain/toolchain-common/branch/master/g
 
 This repo is for controllers, libs, scripts, make files, etc to be shared between host and member operators.
 
+== Setting Up and Connecting Host and Member Operators
+
+In a new terminal, navigate to codeready-toolchain/host-operator directory and execute the following commands
+```
+$ minishift start --profile toolchain
```
xcoulon: why do we need to specify a profile here?

README.adoc
```diff
@@ -6,6 +6,37 @@ image:https://codecov.io/gh/codeready-toolchain/toolchain-common/branch/master/g
...
+$ make up-local
```
xcoulon: ideally, we should run something like `make setup` so we don't have to `ctrl+c` afterwards

README.adoc
```diff
@@ -6,6 +6,37 @@ image:https://codecov.io/gh/codeready-toolchain/toolchain-common/branch/master/g
...
+^C
+```
+In the same terminal, navigate to codeready-toolchain/member-operator directory and execute the following commands
+```
+$ make up-local
+^C
```
xcoulon: same here

README.adoc
```diff
@@ -6,6 +6,37 @@ image:https://codecov.io/gh/codeready-toolchain/toolchain-common/branch/master/g
...
+$ make add-member-to-host
```
xcoulon: there is no such `add-member-to-host` goal in `toolchain-common`

README.adoc
```diff
@@ -6,6 +6,37 @@ image:https://codecov.io/gh/codeready-toolchain/toolchain-common/branch/master/g
...
+```
+In the same terminal, navigate back to codeready-toolchain/host-operator directory and execute the following commands
+```
+make add-host-to-member
```
xcoulon: same here: this goal does not exist

README.adoc
MatousJobanek: in addition, the script works (for now) only with the specific names of the profiles: https://github.com/codeready-toolchain/toolchain-common#scripts

README.adoc
```diff
@@ -6,6 +6,37 @@ image:https://codecov.io/gh/codeready-toolchain/toolchain-common/branch/master/g
 
 This repo is for controllers, libs, scripts, make files, etc to be shared between host and member operators.
 
+== Setting Up and Connecting Host and Member Operators
```
dipak-pawar: I believe, we are connecting clusters here, not operators. can we reword here?

README.adoc
```diff
@@ -6,6 +6,37 @@ image:https://codecov.io/gh/codeready-toolchain/toolchain-common/branch/master/g
...
+```
+In the same terminal, navigate back to codeready-toolchain/member-operator directory and execute the following commands
+```
+$ make up-local
+```
+
+Now that we are setup, we can add resources and watch the logs on the member-operator! To do so, open a new terminal and run the following commands
+```
+$ oc project toolchain-host-operator
+$ oc apply -f '/path/to/masteruserreord.yaml'
```
dipak-pawar: ~`/path/to/masteruserreord.yaml`~ -> `/path/to/masteruserrecord.yaml`.

Also I would give some masteruserrecord.yaml from example for reference

README.adoc
```diff
@@ -6,6 +6,37 @@ image:https://codecov.io/gh/codeready-toolchain/toolchain-common/branch/master/g
...
+^C
+$ make add-member-to-host
+```
+In the same terminal, navigate back to codeready-toolchain/host-operator directory and execute the following commands
+```
+make add-host-to-member
+```
+In the same terminal, navigate back to codeready-toolchain/member-operator directory and execute the following commands
+```
+$ make up-local
```
dipak-pawar: Is there any reason why are we doing `make up-local` on member-operator two times?

README.adoc
```diff
@@ -6,6 +6,37 @@ image:https://codecov.io/gh/codeready-toolchain/toolchain-common/branch/master/g
...
 
+== Setting Up and Connecting Host and Member Operators
+
+In a new terminal, navigate to codeready-toolchain/host-operator directory and execute the following commands
+```
+$ minishift start --profile toolchain
+$ make up-local
+^C
+```
+In the same terminal, navigate to codeready-toolchain/member-operator directory and execute the following commands
```
nurali-techie: From my POV, no need to say in same_terminal and navigate etc.

Instead,
```
$ cd $GOPATH/github.com/codeready-toolchain/member-operator
$ make up-local
^c 
```
Rather explaining in sentence .. do it using `cd` command.

nurali-techie:
@tinakurian FYI, I set up host & member cluster using minishift using below steps (whatever I can recall).

```
$ minishift start --profile host
$ cd xxx/host-operator
$ make up-local
$ ctrl+c

$ minishift start --profile member
$ cd xxx/member-operator
$ make up-local
$ ctrl+c

$ cd xxx/host-operator OR $ cd xxx/member-operator
$ make add-host-to-member
$ make add-member-to-host
```
- so there are two minishift profile.  one for `host` and one for `member`
- make up-local will apply required CRD, role, rb, sa etc

README.adoc
tinakurian: @xcoulon right! it's sitting in the host/member repositories. So you'd have to switch to that repository to execute this. I had mentioned it above in the blurb before the code snippet - maybe there is another way to highlight it, so it is more clear. Easily gets lost among the rest of the words i think.

README.adoc
tinakurian: @xcoulon right! it's sitting in the host/member repositories. So you'd have to switch to that repository to execute this. I had mentioned it above in the blurb before the code snippet - maybe there is another way to highlight it, so it is more clear. Easily gets lost among the rest of the words i think.

README.adoc
tinakurian: @dipak-pawar The second time is to get it up and running so we can actually watch the logs now that it is aware of the host. You could alternatively, use two terminals and then not need to start and stop them. Would that be better in the instructions?

README.adoc
tinakurian: @dipak-pawar Sure! That sounds like a good idea. Will add that!

README.adoc
tinakurian: @nurali-techie That sounds much more clear! Will make that change. Thanks :-)

README.adoc
tinakurian: @MatousJobanek I was able to run them both under the same profile, for instance run both member and host under a member profile or vice versa, so i assumed the profile name didn't matter. I will verify this though!

tinakurian: @xcoulon I'm not entirely sure, the readme on both the member and host repositories specifies you have to. Will test it out and let you know if it's needed though!

README.adoc
tinakurian: Fixed this to be more clear based off of @nurali-techie's comment!

README.adoc
tinakurian: @dipak-pawar Fixed the typo - good catch :-)

tinakurian:
> @tinakurian FYI, I set up host & member cluster using minishift using below steps (whatever I can recall).
> 
> ```
> $ minishift start --profile host
> $ cd xxx/host-operator
> $ make up-local
> $ ctrl+c
> 
> $ minishift start --profile member
> $ cd xxx/member-operator
> $ make up-local
> $ ctrl+c
> 
> $ cd xxx/host-operator OR $ cd xxx/member-operator
> $ make add-host-to-member
> $ make add-member-to-host
> ```
> 
> * so there are two minishift profile.  one for `host` and one for `member`
> * make up-local will apply required CRD, role, rb, sa etc

@nurali-techie Ohh! I was wondering if i needed to do both of these steps: 
> $ make add-host-to-member
> $ make add-member-to-host

Thanks for clarifying. I will update the readme to reflect this. 

Are two profiles really necessary? It seems like i was able to get them running using one. But I can add creating both in either case. Will update! Thanks for the review!

README.adoc
tinakurian: @xcoulon @MatousJobanek so the profile name does need to be either host or member. I've updated the doc to reflect this!

tinakurian:
> @tinakurian `add-host-to-member` is missing. Please include that. Rest looks good. Approving

@nurali-techie Done! Thanks :-)

tinakurian:
@nurali-techie When you get a chance, would you be able to add the lgtm tag?

README.adoc
```diff
@@ -6,6 +6,89 @@ image:https://codecov.io/gh/codeready-toolchain/toolchain-common/branch/master/g
...
+  disabled: false
+  nsLimit: admin
+  nsTemplateSet:
+    tierName: basic
+    namespaces:
+    - type: ide
+      revision: abcdef
+    - type: cicd
+      revision: abcdef
+status:
```
nurali-techie: I guess one last thing, when we use `oc apply` to create CR.  The CR file should not contain `status` field.  I guess, it will be ignored.  `status` is populated by reconciler.  Please remove `status` from both example yaml file.

README.adoc
```diff
@@ -6,6 +6,89 @@ image:https://codecov.io/gh/codeready-toolchain/toolchain-common/branch/master/g
...
+$ make add-host-to-member
+$ make up-local
+```
+
+Now that we are setup, we can add resources and watch the logs on the member-operator! To do so, open a new terminal and run the following commands
+```
+$ oc project toolchain-host-operator
+$ oc apply -f '/path/to/masteruserrecord.yaml'
+$ oc project toolchain-member-operator
+$ oc apply -f '/path/to/useraccount.yaml'
```
MatousJobanek: This commands sequence won't work - don't forget that you use two minishift profiles ;-)

MatousJobanek: in addition, in the previous code snippets, you started only the member operator, so you cannot access the host operator as it is not running

README.adoc
MatousJobanek: @tinakurian could you please address this ^^^ comment? Starting and killing the operator just for the sake of deploying resources is not the best way of showing the functionality.

tinakurian: @xcoulon @MatousJobanek This is definitely something we could add. The task was really to document how we can get both host and member running and talking to each other. I can create a separate task to create a script in order to address this - this will also help avoid scope creep

README.adoc
MatousJobanek: > You could alternatively, use two terminals and then not need to start and stop them.

I don't think so. In your instructions, you use two clusters (two profiles) and I'm not sure if there is an easy way you could reach them without switching the profiles/context. If it is possible to do that, then you should describe it in the description. 

> The second time is to get it up and running so we can actually watch the logs now that it is aware of the host.

This is fine for the member, but then if you want to do the same for the host you would need:
1. to stop the process
2. change the profile
3. change the directory
4. run `up-local` for the host. 

Wouldn't be better - instead of starting/stopping the process - to just deploy the operator and let it running inside of the cluster? Then, you could just change the profiles and namespaces

README.adoc
MatousJobanek: The main task is to improve the user experience for ppl that are new to the project. If there is a room for improvement to make the experience better, then we should do it. In addition, the targets should have been added in host/member projects, so you can do it in parallel.

README.adoc
tinakurian: @MatousJobanek i have updated the instructions already to avoid stopping the host cluster. But the member does need to be started and stopped in order to run add to host and add to member scripts - as mentioned by @nurali-techie here: https://github.com/codeready-toolchain/toolchain-common/pull/19#issuecomment-512085257

README.adoc
tinakurian: @MatousJobanek I'm not sure we are talking about the same story? This was the story which i created related to this work: https://jira.coreos.com/browse/CRT-178. The goal of this story was really for me to learn how to get both host and member running and talking to each other and also to document the process. Again, not apposed to improving the process, just think it should be done as part of a different story/task. I didn't have this in mind when i wrote the story and allocated points to it.

README.adoc
tinakurian: @nurali-techie Removed - thanks for letting me know!

README.adoc
MatousJobanek: OK, if you don't want to change the makefile targets (as part of this task) then I do respect your decision, but please, don't use the `make up-local` for deploying the resources. `make up-local` is a target to be used for running operator locally, not for deploying resources. instead of this, you can use a combination of targets:
`make login-as-admin create-namespace deploy-rbac deploy-crd`
but this is already documented in both operator and member projects:
https://github.com/codeready-toolchain/host-operator#development
https://github.com/codeready-toolchain/member-operator#development

README.adoc
MatousJobanek: the operators don't have to be started in order to create KubeFedCluster the CRs, they can be started later. The Nurali's steps are technically correct, but I wouldn't show setting up clusters by starting/stopping operators without enough explanation

README.adoc
```diff
@@ -6,6 +6,80 @@ image:https://codecov.io/gh/codeready-toolchain/toolchain-common/branch/master/g
...
+apiVersion: toolchain.dev.openshift.com/v1alpha1
+kind: MasterUserRecord
+metadata:
+  name: example
+spec:
+  disabled: false
+  deprovisioned: false
+  userID: 86505192-a386-11e9-ad56-525400ad2b23
+  userAccounts:
+  - targetCluster: member-192-168-42-61-8443
```
MatousJobanek: btw, just a detail - it would be good to add a description to this field as well - that the `targetCluster` should contain the correct IP address of the member cluster

README.adoc
tinakurian: @MatousJobanek I understand now :-) Done!! I tested out the steps in the order they are documented and it worked perfectly. As you suggested, it is more clear to list out the steps used instead of using make up-local. Especially better since we aren't starting the operator when it doesn't need to be started. Thanks for the advice and tips!

README.adoc
tinakurian: @xcoulon Removed!

README.adoc
tinakurian: @MatousJobanek Good point! Added a note to this section to address that! Thanks!

README.adoc
tinakurian: @MatousJobanek Done! This is much clearer - thanks for the suggestion!

README.adoc
tinakurian: @MatousJobanek Done! Agreed - much clearer now.

README.adoc
```diff
@@ -6,6 +6,91 @@ image:https://codecov.io/gh/codeready-toolchain/toolchain-common/branch/master/g
...
```
sbryzak: This should actually be the _name_ of the target cluster, not the IP address.  We use the name to lookup the cluster info, and in fact there is a function (see [here](https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/cluster/cache.go#L67)) that accepts the cluster name, which is used to get a client for the cluster.

tinakurian: @sbryzak Thanks for clarifying! I have updated the comment to describe the field as the **name**. I see in your PR this field contains the name of the cluster we would be pointing at: https://github.com/codeready-toolchain/host-operator/pull/27/files#diff-1696f26dd78d9fa4906ecb5d5c0911c3R7. I guess this looks a bit different since it is using minishift.

tinakurian:
@MatousJobanek @sbryzak @nurali-techie is this PR ready to be merged at this point? It looks like it needs the lgtm label

PR #20, opened 2019-07-15
upload coverage reports using OpenShift CI env vars

xcoulon:


.gitignore
```diff
@@ -3,6 +3,8 @@ coverage.txt
 bin/
 tmp/
 vendor/
+env
+bash
```
xcoulon: I can remove those entries if it's problematic. I downloaded them from https://codecov.io/bash and https://codecov.io/env so I could better understand what flags to use in the makefile

PR #28, opened 2019-08-13
feat(CRT-165,CRT-202): add skeleton for KubeFedCluster e2e test & add test doubles for running e2e test with both operators

MatousJobanek:
This PR adds:
* function for setting up e2e test that uses both operators (host and member)
* doubles that help with waiting for the availability of KubeFedCluster CRs
* skeleton for KubeFedCluster e2e test that will be used in both repositories host and member

### DIFF
--- a/pkg/test/e2e/awaitility.go
+++ b/pkg/test/e2e/awaitility.go
@@ -83,9 +83,9 @@ func (a *SingleAwaitility) WaitForKubeFedCluster(expNs string, clusterType clust
 	})
 }
 
-// WaitForKubeFedClusterWithName waits until there is a KubeFedCluster with the given name
+// WaitForKubeFedClusterConditionWithName waits until there is a KubeFedCluster with the given name
 // and with the given ClusterCondition (if it the condition is nil, then it skips this check)
-func (a *SingleAwaitility) WaitForKubeFedClusterWithName(name string, condition *v1beta1.ClusterCondition) error {
+func (a *SingleAwaitility) WaitForKubeFedClusterConditionWithName(name string, condition *v1beta1.ClusterCondition) error {
 	timeout := Timeout
 	if condition != nil {
 		timeout = (util.DefaultClusterHealthCheckPeriod + 5) * time.Second
--- a/pkg/test/e2e/awaitility.go
+++ b/pkg/test/e2e/awaitility.go
@@ -1,0 +1,136 @@
+package e2e
+
+import (
+	"context"
+	"github.com/codeready-toolchain/toolchain-common/pkg/cluster"
+	framework "github.com/operator-framework/operator-sdk/pkg/test"
+	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/kubefed/pkg/apis/core/common"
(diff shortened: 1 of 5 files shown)


pkg/test/e2e/kubefed.go
```diff
@@ -0,0 +1,174 @@
...
+	t.Run("create new KubeFedCluster with correct data and expect to be ready", func(t *testing.T) {
+		// given
+		newName := "new-ready-" + string(kubeFedClusterType)
+		newFedCluster := NewKubeFedCluster(singleAwait.Ns, newName, current.Spec.CABundle,
+			current.Spec.APIEndpoint, current.Spec.SecretRef.Name, labels)
+
+		// when
+		err := awaitility.Client.Create(context.TODO(), newFedCluster, CleanupOptions(ctx))
+
+		// then the KubeFedCluster should not be ready
```
alexeykazakov: It **should** be ready, right? You basically created a new fed cluster resource which points to the same expected-to-be-already-available cluster (current.Spec.APIEndpoint), right? So, the comment is not accurate?

pkg/test/e2e/kubefed.go
```diff
@@ -0,0 +1,174 @@
...
+		newName := "new-offline-" + string(kubeFedClusterType)
+		newFedCluster := NewKubeFedCluster(singleAwait.Ns, newName, current.Spec.CABundle,
+			"https://1.2.3.4:8443", current.Spec.SecretRef.Name, labels)
+
+		// when
+		err := awaitility.Client.Create(context.TODO(), newFedCluster, CleanupOptions(ctx))
+
+		// then
+		require.NoError(t, err)
+		err = singleAwait.WaitForKubeFedClusterWithName(newFedCluster.Name, &v1beta1.ClusterCondition{
```
alexeykazakov: I wonder if we should name `WaitForKubeFedClusterWithName` function something like `WaitForKubeFedClusterConditionWithName`? A bit long name but we are waiting for the specific condition rather than for the cluster itself.

pkg/test/e2e/awaitility.go
```diff
@@ -0,0 +1,136 @@
...
+		return false, err
+	})
+}
+
+// WaitForKubeFedClusterWithName waits until there is a KubeFedCluster with the given name
+// and with the given ClusterCondition (if it the condition is nil, then it skips this check)
+func (a *SingleAwaitility) WaitForKubeFedClusterWithName(name string, condition *v1beta1.ClusterCondition) error {
+	timeout := Timeout
+	if condition != nil {
+		timeout = (util.DefaultClusterHealthCheckPeriod + 5) * time.Second
```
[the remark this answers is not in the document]
MatousJobanek: the DefaultClusterHealthCheckPeriod is currently 10 seconds and to update the ClusterCondition it usually takes 11-12 seconds. I put the 3 additional seconds just as a cushion to not fail just because it didn't make to change the condition in time. I can lower the cushion to 3 seconds, but that won't help much with making the test faster. What would help is to have the periods customizable - I created a Jira issue for that: https://jira.coreos.com/browse/CRT-216

pkg/test/e2e/kubefed.go
MatousJobanek: yes, you are right. I was struggling with the test configuration for some time and I have changed the tests as well as their order several times, but I forgot to change the comments :-). 
Thanks for pointing this out - fixed: 13307bf

pkg/test/e2e/kubefed.go
MatousJobanek: Good point - renamed bf3ebc5

PR #30, opened 2019-08-19
feat(test): add status update mock to fake client

MatousJobanek:
add status update mock to fake client so I can mock calls that update the status

### DIFF
--- a/pkg/test/client.go
+++ b/pkg/test/client.go
@@ -13,15 +13,24 @@ import (
 // NewFakeClient creates a fake K8s client with ability to override specific Create/Update/Delete functions
 func NewFakeClient(t *testing.T, initObjs ...runtime.Object) *FakeClient {
 	client := fake.NewFakeClientWithScheme(scheme.Scheme, initObjs...)
-	return &FakeClient{client, t, nil, nil, nil}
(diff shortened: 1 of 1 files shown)


nurali-techie:
@MatousJobanek for me, until I know one usecase of this new function, it's difficult to review.

MatousJobanek:
@nurali-techie it's chicken-egg problem - until it's in master I cannot provide a working solution (I mean working in CI).

>  until I know one usecase of this new function, it's difficult to review.

I think that if you take a look at our logic of updating statuses, the already mocked methods and their usage, you can immediately imagine some of the use cases ;-).
Anyway - here is one of the cases why it's needed: https://github.com/codeready-toolchain/host-operator/pull/23/commits/6f8aa52f30b742777428443ae7c7e02675af308b#diff-37787ca0ef3b20842b503a57aeae391fR155-R157

nurali-techie:
@MatousJobanek thanks, I was expecting where it's used (consuming PR only).  So for me, when we add something new in `common`, we would want to use in host/member.  So if both PR are available to see then it's enough to review.  It's similar to adding new stuff in `api` project and sharing corresponding host/member PR :-)

PR #32, opened 2019-08-21
Add Get and List to fake client

alexeykazakov:
This PR introduces `Get()` and `List()` mock functions to the mock client.

Also it adds `ctx context.Context` param to all mock functions because sometimes it's needed in our tests. This change breaks API but it's pretty easy to fix the existing tests to support `ctx` param.

### DIFF
--- a/pkg/test/client_test.go
+++ b/pkg/test/client_test.go
@@ -1,0 +1,121 @@
+package test
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/api/core/v1"
+	errs "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


pkg/test/client_test.go
```diff
@@ -0,0 +1,121 @@
...
+	require.NotNil(t, fclient)
+
+	assert.Nil(t, fclient.MockGet)
+	assert.Nil(t, fclient.MockList)
+	assert.Nil(t, fclient.MockUpdate)
+	assert.Nil(t, fclient.MockDelete)
+	assert.Nil(t, fclient.MockCreate)
+	assert.Nil(t, fclient.MockStatusUpdate)
+
+	key := types.NamespacedName{"somenamespace", "somename"}
```
golangcibot: composites: `github.com/codeready-toolchain/toolchain-common/vendor/k8s.io/apimachinery/pkg/types.NamespacedName` composite literal uses unkeyed fields (from `govet`)

PR #34, opened 2019-08-29
chore: move custom assertions and doubles from host-operator repo

MatousJobanek:
This is the first step of the whole effort to move e2e tests to a separated repository
This PR only moves custom assertions and test doubles from host-operator to toolchain-common repo


pkg/test/masteruserrecord/assertion.go
```diff
@@ -0,0 +1,98 @@
+package murtest
```
xcoulon: why do you name this package `murtest` and not `masteruserrecord_test` as we would normally do?

pkg/test/masteruserrecord/assertion.go
```diff
@@ -0,0 +1,98 @@
...
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
+	"github.com/codeready-toolchain/toolchain-common/pkg/test"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"testing"
+)
+
+type MurAssertion struct {
```
xcoulon: could simply be `Assertion`

pkg/test/useraccount/assertion.go
```diff
@@ -0,0 +1,66 @@
+package uatest
```
xcoulon: same question about this pkg name

pkg/test/useraccount/doubles.go
```diff
@@ -0,0 +1,40 @@
...
+			Name:      mur.Name,
+			Namespace: test.MemberOperatorNs,
+		},
+		Spec: mur.Spec.UserAccounts[0].Spec,
+	}
+	ModifyUa(ua, modifiers...)
+	return ua
+}
+
+func ModifyUa(ua *toolchainv1alpha1.UserAccount, modifiers ...UaModifier) {
```
xcoulon: should we make this function private?

pkg/test/masteruserrecord/doubles.go
```diff
@@ -0,0 +1,145 @@
...
+						Revision: "123abc",
+						Template: "",
+					},
+				},
+			},
+		},
+	}
+}
+
+func ModifyMur(mur *toolchainv1alpha1.MasterUserRecord, modifiers ...MurModifier) {
```
xcoulon: should we make this function private?

pkg/test/masteruserrecord/doubles.go
```diff
@@ -0,0 +1,145 @@
...
+	for _, modify := range modifiers {
+		modify(mur)
+	}
+}
+
+func ModifyUaInMur(mur *toolchainv1alpha1.MasterUserRecord, targetCluster string, modifiers ...UaInMurModifier) {
```
xcoulon: should we make this function private?

pkg/test/masteruserrecord/assertion.go
MatousJobanek: because it's too long. in addition, it's not a test but assertion - you shouldn't use `_test` suffix in that case

pkg/test/useraccount/assertion.go
MatousJobanek: and same answer :-)

pkg/test/masteruserrecord/assertion.go
MatousJobanek: done 4101afa

pkg/test/useraccount/doubles.go
MatousJobanek: Why? It's used in the test: https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/masteruserrecord/sync_test.go#L85

pkg/test/masteruserrecord/doubles.go
MatousJobanek: This one is not used, indeed. But I guess that for consistency it would be good to keep it exported as `ModifyUa` is - so it can be used in a test.

pkg/test/masteruserrecord/doubles.go
MatousJobanek: this function is used heavily - eg here: https://github.com/codeready-toolchain/host-operator/pull/59/files#diff-93c8b7a9cc0875ddd6d3002d3d0b4253R212

pkg/test/masteruserrecord/assertion.go
xcoulon: @MatousJobanek I missed the fact that it's a utility pkg for the tests, and we can't use the `_test` suffix for that. But I'd like to have the pkg name that matches the directory name, really. I think it's a good convention. We can always use an alias when importing the pkg somewhere else.

pkg/test/useraccount/assertion.go
xcoulon: please see https://github.com/codeready-toolchain/toolchain-common/pull/34#discussion_r319404170 about the package name

pkg/test/useraccount/doubles.go
xcoulon: ah ok, I thought it was only used in the `NewUserAccountFromMur` func. 
but since it's in the `useraccounts` pkg, could we just rename it to `Modify`?

pkg/test/masteruserrecord/doubles.go
xcoulon: ok, makes sense. Since it's in the `masteruserrecord` pkg, could you rename this func to `Modify`? (no need to repeat the `Mur` here)

pkg/test/masteruserrecord/doubles.go
xcoulon: ok

pkg/test/useraccount/doubles.go
MatousJobanek: done 63785f9

pkg/test/masteruserrecord/doubles.go
MatousJobanek: done 63785f9

PR #42, opened 2019-10-04
decode template util func added for used in other test

nurali-techie:


pkg/test/template/util_test.go
```diff
@@ -0,0 +1,75 @@
...
+	err := apis.AddToScheme(s)
+	require.NoError(t, err)
+	codecFactory := serializer.NewCodecFactory(s)
+	decoder := codecFactory.UniversalDeserializer()
+
+	t.Run("valid_template", func(t *testing.T) {
+		// test
+		tmpl, err := DecodeTemplate(decoder, validTmplContent)
+
+		assert.NoError(t, err)
```
xcoulon: you should use 
```suggestion
		require.NoError(t, err)
``` 
here instead

pkg/test/template/util_test.go
```diff
@@ -0,0 +1,75 @@
...
+		assert.NotNil(t, tmpl)
```
xcoulon: same here: use 
```suggestion
		require.NotNil(t, tmpl)
```

pkg/test/template/util.go
```diff
@@ -0,0 +1,25 @@
...
+		return nil, fmt.Errorf("decoder cannot be nil")
+	}
+	obj, _, err := decoder.Decode(tmplContent, nil, nil)
+	if err != nil {
+		return nil, errs.Wrapf(err, "unable to decode template")
+	}
+	tmpl, ok := obj.(*templatev1.Template)
+	if !ok {
+		return nil, fmt.Errorf("unable to convert object type %T to Template, must be a v1.Template", obj)
+	}
```
xcoulon: actually, you can do it in a simpler way, without having to convert the returned object into a `*templatev1.Template`:

```suggestion
tmplObj := &templatev1.Template{}
_, _, err = g.decoder.Decode(tmpl, nil, tmplObj)
if err != nil {
	return nil, errs.Wrapf(err, "unable to decode template")
}
```

pkg/test/template/util_test.go
```diff
@@ -0,0 +1,75 @@
...
+		assert.Equal(t, "mytemplate", tmpl.GetName())
+		assert.NotNil(t, tmpl.Objects)
```
xcoulon: same here:
```suggestion
		require.NotNil(t, tmpl.Objects)
```

pkg/test/template/util_test.go
```diff
@@ -0,0 +1,75 @@
...
+		assert.Equal(t, 1, len(tmpl.Objects))
+	})
+
+	t.Run("invalid_template", func(t *testing.T) {
+		// test
+		tmpl, err := DecodeTemplate(decoder, invalidTmplContent)
+
+		assert.Error(t, err)
```
xcoulon:
```suggestion
		require.Error(t, err)
```

pkg/test/template/util_test.go
```diff
@@ -0,0 +1,75 @@
...
+		assert.Nil(t, tmpl)
+		assert.Contains(t, err.Error(), "unable to decode template")
+	})
+
+	t.Run("nil_decoder", func(t *testing.T) {
+		// test
+		tmpl, err := DecodeTemplate(nil, invalidTmplContent)
+
+		assert.Error(t, err)
```
xcoulon:
```suggestion
		require.Error(t, err)
```

pkg/test/template/util_test.go
```diff
@@ -0,0 +1,75 @@
...
+		assert.Nil(t, tmpl)
+		assert.Contains(t, err.Error(), "decoder cannot be nil")
+	})
+
+	t.Run("nil_template", func(t *testing.T) {
+		// test
+		tmpl, err := DecodeTemplate(decoder, nil)
+
+		assert.Error(t, err)
```
xcoulon:
```suggestion
		require.Error(t, err)
```

pkg/test/template/util.go
nurali-techie: @xcoulon with this simpler way I don't thing this wrapper function required in `common/test`.  We can directly call `decoder.Decode(tmplContent, nil, tmplObj)` in our test_code wherever required .. WDYS ??

xcoulon: right, the function body is pretty straightforward, indeed, and this func here adds no real benefit, so we can probably close this PR.

nurali-techie: Thanks for simpler way suggestion :-)  Closing this PR.

PR #48, opened 2019-11-02
Switch to Kube 1.14

alexeykazakov:
Requires https://github.com/codeready-toolchain/toolchain-common/pull/49

Related host PR - https://github.com/codeready-toolchain/host-operator/pull/98
member PR - https://github.com/codeready-toolchain/member-operator/pull/89


pkg/test/client.go
```diff
@@ -42,37 +50,51 @@ func (c *FakeClient) Get(ctx context.Context, key client.ObjectKey, obj runtime.
...
 
 func (c *FakeClient) Status() client.StatusWriter {
+	m := mockStatusUpdate{}
+	if c.MockStatusUpdate == nil && c.MockStatusPatch == nil {
+		return c.Client.Status()
+	}
+	if c.MockStatusUpdate != nil {
+		m.mockUpdate = c.MockStatusUpdate
+	}
 	if c.MockStatusUpdate != nil {
```
xcoulon: shouldn't it be
```suggestion
	if c. MockStatusPatch != nil {
		m. mockPatch = c. MockStatusPatch
	}
```

xcoulon: (be careful though, I'm not sure how this suggestion will affect the current code, in terms of lines to change)

alexeykazakov: Yes. Good catch. Fixed.

pkg/test/client.go
```diff
@@ -42,37 +50,51 @@ func (c *FakeClient) Get(ctx context.Context, key client.ObjectKey, obj runtime.
...
 func (c *FakeClient) Status() client.StatusWriter {
+	m := mockStatusUpdate{}
+	if c.MockStatusUpdate == nil && c.MockStatusPatch == nil {
+		return c.Client.Status()
+	}
 	if c.MockStatusUpdate != nil {
-		return &mockStatusUpdate{mockUpdate: c.MockStatusUpdate}
+		m.mockUpdate = c.MockStatusUpdate
+	}
+	if c. MockStatusPatch != nil {
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
	if c.MockStatusPatch != nil {
```

xcoulon: ooops, I guess it's my bad @alexeykazakov :/

pkg/test/client.go
```diff
@@ -13,26 +13,34 @@ import (
 // NewFakeClient creates a fake K8s client with ability to override specific Get/List/Create/Update/StatusUpdate/Delete functions
 func NewFakeClient(t *testing.T, initObjs ...runtime.Object) *FakeClient {
 	client := fake.NewFakeClientWithScheme(scheme.Scheme, initObjs...)
-	return &FakeClient{client, t, nil, nil, nil, nil, nil, nil}
+	return &FakeClient{client, t, nil, nil, nil, nil, nil, nil, nil, nil, nil}
```
MatousJobanek: I'm curious - we do we need to specify all the nil values when nil is the default value?
```suggestion
	return &FakeClient{Client: client, T: t}
```

pkg/test/client_test.go
```diff
@@ -81,41 +128,57 @@ func TestNewClient(t *testing.T) {
...
 	t.Run("mock Status Update", func(t *testing.T) {
 		defer func() { fclient.MockStatusUpdate = nil }()
-		fclient.MockStatusUpdate = func(ctx context.Context, obj runtime.Object) error {
+		fclient.MockStatusUpdate = func(ctx context.Context, obj runtime.Object, opts ...client.UpdateOption) error {
 			return expectedErr
 		}
 		assert.EqualError(t, fclient.MockStatusUpdate(context.TODO(), &v1.Secret{}), expectedErr.Error())
 	})
+
+	t.Run("mock Status Patch", func(t *testing.T) {
```
MatousJobanek: There is missing the same subtest for the `MockPatch` method

pkg/test/client.go
alexeykazakov: Fixed in https://github.com/codeready-toolchain/toolchain-common/pull/48/commits/306f81888e72953684d26874f19edca862f54f94

pkg/test/client_test.go
alexeykazakov: Good catch! Fixed in https://github.com/codeready-toolchain/toolchain-common/pull/48/commits/306f81888e72953684d26874f19edca862f54f94

PR #51, opened 2019-11-05
ignore actual conditions if no expectation is provided

xcoulon:


xcoulon:
closing as we are not going to do that, finally :(

PR #52, opened 2019-11-06
Update KubeFedCluster CRD

alexeykazakov:
KubeFed folks changed KubeFedCluster CRD group from `core.kubefed.k8s.io` to `core.kubefed.io`

The new CRD was generated by https://github.com/codeready-toolchain/api/pull/57

Required for:
https://github.com/codeready-toolchain/toolchain-e2e/pull/32
https://github.com/codeready-toolchain/member-operator/pull/89
https://github.com/codeready-toolchain/host-operator/pull/98


scripts/update-kubefed-crd.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
```
xcoulon: would it be simpler if we use something like `go-bindata`, as we've already done in other parts of the codebase, rather than having another shell script to maintain?

pkg/cluster/kubefedcluster.go
```diff
@@ -0,0 +1,28 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// EnsureKubeFedClusterCrd creates a KubeFedCluster CRD in the cluster.
+// If the creation returns an error that is of the type "AlreadyExists" then the error is ignored,
+// if the error is of another type then it is returned
+func EnsureKubeFedClusterCrd(scheme *runtime.Scheme, client client.Client) error {
```
xcoulon: I know it's a minor thing, but I believe that this func should be named `EnsureKubeFedClusterCRD` instead, since `CRD` is an acronym.

scripts/update-kubefed-crd.sh
MatousJobanek: Yes, I agree with Xavier. I believe that embedding and using the CRD as an asset will be much easier. 
I simply copy-pasted the content of the KubeFedCluster CRD to a constant because I didn't expect that the CRD would change during the time - I kind of considered it as a final CRD. 
In other words, I should have done it by using the `go-bindata` - my bad, sorry.

make/generate.mk
```diff
@@ -0,0 +1,7 @@
+.PHONY: generate-kubefed-crd
+generate-kubefed-crd: vendor
```
MatousJobanek: I'm wondering - cannot that be part of the `api` repo? In the same way as is done for host/member-operator repos. We do have the dependency to KubeFed project as well, so it should be doable...

scripts/update-kubefed-crd.sh
alexeykazakov: OK. Makes sense.
Moved to go-bindata assets and moved the generation to API - https://github.com/codeready-toolchain/api/pull/57

pkg/cluster/kubefedcluster.go
alexeykazakov: Done.

make/generate.mk
alexeykazakov: Makes sense.
Moved to API - https://github.com/codeready-toolchain/api/pull/57

pkg/cluster/kubefedcluster_assets.go
```diff
@@ -0,0 +1,353 @@
+// Code generated for package cluster by go-bindata DO NOT EDIT. (@generated)
```
xcoulon: ok, so here we include the generated file in the git repo because `toolchain-common` is a library for the operators. However, we should have a `make` goal to generate this file, and I can't see it in your PR. Or did I miss something? 

Alternatively, we could try to use [`go generate` comments](https://blog.golang.org/generate) to trigger the call to `go-bindata` during the operator build, and thus avoid committing this file in the repo...

alexeykazakov: The generation is part of api repo `make generate` target. Check out the linked PR.
And we can’t generate it in build time. We need to fully control any possible changes in CRDs which can be introduced by newer kubefed. New CRD must go though PR reviews.

xcoulon: ok, thanks for the clarifications @alexeykazakov

PR #54, opened 2019-11-07
chore: update KubeFed group in add-cluster.sh script

MatousJobanek:


MatousJobanek:
We all missed that ;-)

PR #58, opened 2019-11-14
add kubeconfig option to avoid prompt

dipak-pawar:
Currently, we are working on infrastructure setup script, which will do the following
- Creates host and member cluster
- Setup IDP
- Create all admin users
- Deploy operators
- Setup KubeFed

For the last point, we need two different cluster credentials kubeconfig to be provided to switch the context and login to respective cluster using context.

This script assumes that if user is passing kubeconfig then it needs to have context with `host-admin` and `member-admin`

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -124,7 +124,7 @@ $ oc apply -f deploy/user.yaml
 === add-cluster.sh
 
 The CodeReady Toolchain architecture contains two types of clusters `host` and `member`.
-To connect these two clusters together it is necessary to run a script link:scripts/add-cluster.sh[] that takes multiple flags like, `--type`, `--member-ns`, `--host-ns`, `--single-cluster`. However you can see all these options by using `./scripts/add-cluster.sh -h`
+To connect these two clusters together it is necessary to run a script link:scripts/add-cluster.sh[] that takes multiple flags like, `--type`, `--member-ns`, `--host-ns`, `--single-cluster`, `--minishift`. However you can see all these options by using `./scripts/add-cluster.sh -h`
 
 ==== host and member clusters using host and member profiles on minishift
 Make sure that you have started minishift as different profiles. You can use following commands for it.
--- a/README.adoc
+++ b/README.adoc
@@ -218,8 +218,8 @@ in respective namespace with all requires resources like Service Account, (Clust
     . creates a secret in `toolchain-member-operator` namespace with the SA token taken from the `toolchain-host-operator` ns
     . creates `KubeFedCluster` in `toolchain-member-operator` namespace CR representing the added `host`
 
-==== Using a different clusters for host and member operators
-Make sure to have two different clusters ready and available. To setup Toolchain, you'll have to run `host-operator` and `registration-service` on host cluster and `member-operator` on member cluster.
+==== Using different clusters for host and member operators
+Make sure you have two different clusters ready and available where the `host-operator` should be running in the host cluster and `member-operator` in the member cluster.
 
 To do either you can pass kube-config using `-kc` flag with kubeconfig having `host-admin` and `member-admin` contexts or if you don't have kubeconfig you can provide required detail of cluster when promted for cluster URL and token.
 
--- a/README.adoc
+++ b/README.adoc
@@ -124,7 +124,7 @@ $ oc apply -f deploy/user.yaml
 === add-cluster.sh
 
 The CodeReady Toolchain architecture contains two types of clusters `host` and `member`.
-To connect these two clusters together it is necessary to run a script link:scripts/add-cluster.sh[] that takes multiple flags like, `--type`, `--member-ns`, `--host-ns`, `--single-cluster`, `--minishift`, `--kube-config`. However you can see all these options by using `./scripts/add-cluster.sh -h`
+To connect these two clusters together it is necessary to run a script link:scripts/add-cluster.sh[] that takes multiple flags like, `--type`, `--member-ns`, `--host-ns`, `--single-cluster`, `--kube-config`. However you can see all these options by using `./scripts/add-cluster.sh -h`
 
 ==== host and member clusters using host and member profiles on minishift
 Make sure that you have started minishift as different profiles. You can use following commands for it.
--- a/README.adoc
+++ b/README.adoc
@@ -218,6 +218,54 @@ in respective namespace with all requires resources like Service Account, (Clust
     . creates a secret in `toolchain-member-operator` namespace with the SA token taken from the `toolchain-host-operator` ns
     . creates `KubeFedCluster` in `toolchain-member-operator` namespace CR representing the added `host`
 
+==== Using a different clusters for host and member operators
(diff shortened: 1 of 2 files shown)


scripts/add-cluster.sh
```diff
@@ -99,13 +109,13 @@ SA_TOKEN=`oc get secret ${SA_SECRET} -n ${OPERATOR_NS}  -o json | jq -r '.data["
 SA_CA_CRT=`oc get secret ${SA_SECRET} -n ${OPERATOR_NS} -o json | jq -r '.data["ca.crt"]'`
 
 # this env variable is set in openshift-ci environment.
-# openshift ci has long name for cluster i.e.> 63 characters if read from config which is not allowed by k8s/openshift
-if [[ -z ${OPENSHIFT_BUILD_NAMESPACE} ]]; then
+# openshift 4 has long name for cluster i.e.> 63 characters if read from config which is not allowed by k8s/openshift
+if [[ ${MINISHIFT} == "true" ]]; then
```
alexeykazakov: I think it's breaks the local deployments when using `make test-e2e` and `make test-e2e-local` from e2e repo when running against minishift. I don't think we set minishift flag in this case.
We would need to use something to automatically detect the version of OS here instead of relaying on the flag.

MatousJobanek: we should be able to easily check if the OC is version 3 or 4: 
```
IS_OS3=`curl -k -XGET -H "Authorization: Bearer $(oc whoami -t 2>/dev/null)" $(oc config view --minify -o jsonpath='{.clusters[0].cluster.server}')/version/openshift 2>/dev/null | grep paths`
```
if the OpenShift is of version 4, then the variable is empty (in OS4 the `/version/openshift` endpoint is missing), if it is of version 3 (in other words is minishift) then the var is not empty

MatousJobanek: in other words - we could easily remove the parameter `--minishift` and replace it by:
```
MINISHIFT=`curl -k -XGET -H "Authorization: Bearer $(oc whoami -t 2>/dev/null)" $(oc config view --minify -o jsonpath='{.clusters[0].cluster.server}')/version/openshift 2>/dev/null | grep paths`
```

MatousJobanek: Btw, why do we need to change the `if` check? For now, it was working fine with the 48h clusters (that are OS4), so what is the difference in the case of dev clusters?

scripts/add-cluster.sh
```diff
@@ -64,7 +74,7 @@ while test $# -gt 0; do
             *)
                echo "$1 is not a recognized flag!"
                user_help
-               exit -1
+               exit 0
```
MatousJobanek: Why? It should return non-zero value when there is provided unsupported flag

scripts/add-cluster.sh
```diff
@@ -8,18 +8,23 @@ user_help () {
...
         if [[ ${MINISHIFT} != "true" ]]; then
+          if [[ -z ${KUBECONFIG} ]]; then
             echo "Write the server url for the cluster type: $1:"
             read -p '> ' CLUSTER_URL
             echo "Provide the token to log in to the cluster ${CLUSTER_URL}:"
             read -sp '> ' CLUSTER_TOKEN
             echo ""
             oc login --token=${CLUSTER_TOKEN} --server=${CLUSTER_URL}
+          else
+            oc config use-context "$1-admin"
```
MatousJobanek: Please, add also a description to the readme how to set the contexts so they are properly used by the script - what are the expected names, etc...

scripts/add-cluster.sh
dipak-pawar: > Btw, why do we need to change the if check? For now, it was working fine with the 48h clusters (that are OS4), so what is the difference in the case of dev clusters?

With the current logic, we are using `OPENSHIFT_BUILD_NAMESPACE`  env variable, which is set only in openshift-ci environment. If we are running 48 hrs dev cluster, we don't have `OPENSHIFT_BUILD_NAMESPACE`  variable set and I don't want that to be set explicitly by user even not needed

MatousJobanek: I understand the situation with `OPENSHIFT_BUILD_NAMESPACE` env variable and I didn't ask you about it. 
The reason why we have different ways of getting  `API_ENDPOINT` and `JOINING_CLUSTER_NAME` vas is (from the comment in the script):
`# openshift ci has long name for cluster i.e.> 63 characters if read from config which is not allowed by k8s/openshift`
which is the case for openshift-ci, but not for dev clusters. In other words - why we cannot take the values from the kubeconfig also for the dev clusters?

Anyway, it's fine if we change the if statement to use kubeconfig only for OS3 minishift, but use the curl command that I posted above, so we don't have to change the toolchain-e2e test setup. I was just curious why do we need to do that - if there is another reason...

dipak-pawar: it was only reason which is added in the comment, but as we are getting cluster name from infrastucture CRD for openshift 4, let's use it for openshift 4 to be consistent.

Also it won't change if we read it from config for dev clusters or openshift-ci and both are creating new openshift 4 cluster.

scripts/add-cluster.sh
dipak-pawar: Addressed in  d4f3d0d

scripts/add-cluster.sh
dipak-pawar: Addressed in d4f3d0d

scripts/add-cluster.sh
dipak-pawar: Addressed in d4f3d0d

README.adoc
```diff
@@ -218,6 +218,54 @@ in respective namespace with all requires resources like Service Account, (Clust
     . creates a secret in `toolchain-member-operator` namespace with the SA token taken from the `toolchain-host-operator` ns
     . creates `KubeFedCluster` in `toolchain-member-operator` namespace CR representing the added `host`
 
+==== Using a different clusters for host and member operators
+Make sure to have two different clusters ready and available. To setup Toolchain, you'll have to run `host-operator` and `registration-service` on host cluster and `member-operator` on member cluster.
```
MatousJobanek: I don't think that we need to mention the registration service here as there is no relation to the script. In addition, the reg-service will be part of the host operator
```suggestion
Make sure you have two different clusters ready and available where the `host-operator` should be running in the host cluster and `member-operator` in the member cluster.
```

scripts/add-cluster.sh
```diff
@@ -7,19 +7,23 @@ user_help () {
...
     echo "-s,  --single-cluster running both operators on single cluster"
-    echo "-m,  --minishift      in case of using minishift"
+    echo "-kc,  --kube-config   kubeconfig for managing multiple clusters"
     exit 0
 }
 
 login_to_cluster() {
     if [[ ${SINGLE_CLUSTER} != "true" ]]; then
-        if [[ ${MINISHIFT} != "true" ]]; then
+        if [[ -z "$IS_MINISHIFT" ]]; then
```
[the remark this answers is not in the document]
alexeykazakov: It looks like it's the only @MatousJobanek's comment left to address.

PR #61, opened 2019-11-20
Switch to Cluster in UserAccount Embeded Status

alexeykazakov:
Required for https://github.com/codeready-toolchain/host-operator/pull/103

### DIFF
--- a/pkg/test/masteruserrecord/assertion.go
+++ b/pkg/test/masteruserrecord/assertion.go
@@ -73,6 +73,15 @@ func (a *Assertion) AllUserAccountsHaveStatusSyncIndex(syncIndex string) *Assert
 	return a
 }
 
+func (a *Assertion) AllUserAccountsHaveCluster(expected toolchainv1alpha1.Cluster) *Assertion {
+	err := a.loadUaAssertion()
+	require.NoError(a.t, err)
+	for _, ua := range a.masterUserRecord.Status.UserAccounts {
+		assert.Equal(a.t, expected, ua.Cluster)
+	}
+	return a
+}
+
 func (a *Assertion) AllUserAccountsHaveCondition(expected toolchainv1alpha1.Condition) *Assertion {
 	err := a.loadUaAssertion()
 	require.NoError(a.t, err)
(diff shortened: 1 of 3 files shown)


pkg/test/masteruserrecord/assertion.go
```diff
@@ -73,6 +73,15 @@ func (a *Assertion) AllUserAccountsHaveStatusSyncIndex(syncIndex string) *Assert
 	return a
 }
 
+func (a *Assertion) AllUserAccountsHaveCluster(expected toolchainv1alpha1.Cluster) *Assertion {
+	err := a.loadUaAssertion()
+	require.NoError(a.t, err)
+	for _, ua := range a.masterUserRecord.Status.UserAccounts {
+		assert.Equal(a.t, expected, ua.Cluster)
```
xcoulon: why don't you use something like `ua.Cluster.Name == targetCluster || ua.TargetCluster == targetCluster` as above?

pkg/test/masteruserrecord/assertion.go
```diff
@@ -73,6 +73,15 @@ func (a *Assertion) AllUserAccountsHaveStatusSyncIndex(syncIndex string) *Assert
 	return a
 }
 
+func (a *Assertion) AllUserAccountsHaveCluster(expected toolchainv1alpha1.Cluster) *Assertion {
```
MatousJobanek: Does it make sense that all `UserAccounts` in one MUR have the same cluster? it's not really realistic case - `UserAccounts` should be provisioned in different clusters...

pkg/test/masteruserrecord/assertion.go
alexeykazakov: Because we need to compare the entire cluster (which is a stuct now, not just a string). Comparing cluster names only is not enough. And we don't care about deprecated `TargetCluster` here because this new assertion is used in the new code only.

xcoulon: ok, so the deprecated `TargetCluster` field will be empty for both `expected` and `ua` objects in this case, then. Makes sense.

pkg/test/masteruserrecord/assertion.go
alexeykazakov: You are right. It won't be realistic when we support multiple UserAccounts. So that assertion won't be useful for checking MURs with multiple UserAccounts. But we are not going to support multiple accounts any time soon. And for now it's still useful to have such an assertion.

I wanted to provide something quick and simple for the current usecases and we can introduce another more useful assertion for multiple accounts when we need it.

alexeykazakov:
Thanks for the quick review @MatousJobanek @xcoulon !

PR #63, opened 2019-11-26
Added temporary fix for token timestamp issue

sbryzak:
Signed-off-by: Shane Bryzak <sbryzak@gmail.com>


pkg/test/auth/tokenmanager.go
```diff
@@ -144,9 +144,30 @@ func (tg *TokenManager) Key(kid string) (*rsa.PrivateKey, error) {
 	return key, nil
 }
 
+
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion

```

alexeykazakov:
But this PR will affect tests only.
What about at least dev env?
Also, isn’t 5 milliseconds too small leeway.

pkg/test/auth/tokenmanager.go
```diff
@@ -144,9 +144,29 @@ func (tg *TokenManager) Key(kid string) (*rsa.PrivateKey, error) {
...
+	c.StandardClaims.ExpiresAt += leeway
+	err := c.StandardClaims.Valid()
+	c.StandardClaims.ExpiresAt -= leeway
+	return err
+}
+
 // GenerateToken generates a default token.
 func (tg *TokenManager) GenerateToken(identity Identity, kid string, extraClaims ...ExtraClaim) *jwt.Token {
 	token := jwt.New(jwt.SigningMethodRS256)
+	token.Claims = &MyClaims{&jwt.StandardClaims{ExpiresAt: time.Now().Unix() - 1}}
```
tinakurian: i think this is the same as the "exp" field which is specified down below: https://github.com/codeready-toolchain/toolchain-common/pull/63/files#diff-c2bd6592ef1ff7e697a51858a44abfaeR176

tinakurian:
> But this PR will affect tests only. What about at least dev env? Also, isn’t 5 milliseconds too small leeway.

How do other applications interacting with RHD handle this? I would have assumed all the machines run on the same time so that everyone wouldn't need this type of workaround

sbryzak:
I've increased the leeway to 500ms, do you think that is too high?  I still have problems with setting the claims, it turns out it's not so easy to convert jwt StandardClaims to MapClaims.  As for the dev environment, are we able to reproduce this issue there?  I would say the problem lies in [this line of code](https://github.com/codeready-toolchain/registration-service/blob/master/pkg/auth/tokenparser.go#L38), would you concur with that?

sbryzak:
I'm also assuming that it is [this code](https://github.com/dgrijalva/jwt-go/blob/master/claims.go#L49) (from jwt-go) that is the cause of the issue:

```
	if c.VerifyNotBefore(now, false) == false {
		vErr.Inner = fmt.Errorf("token is not valid yet")
		vErr.Errors |= ValidationErrorNotValidYet
	}
```

If that is the case, then parsing the token will be an issue, unless we ignore `ValidationErrorNotValidYet` errors when parsing.

alexeykazakov:
Tina, weren’t you one who reported that time async issue between your
machine (minishift) and the SSO? :)

The other one was Abhishek.

While I don’t think we will have this issue when running the reg services
on any major cloud provider it can be an issue in dev env as it was for you
and Abhishek. It’s OK if we have some manual workaround for dev environment
though but I’m not sure we do. There is just one Signup button which does
both login and POST /signup so we can’t control any delays here...
But maybe on other click on Signup would work actually... since the user is
already logged in and the token is available to be used for POST.

alexeykazakov:
Can you Tina work with Abhishek and see if landing page works for you if
you click on Signup multiple times (if it fails the first time)? If it does
then I think we are ok if we fix it for the tests only.
Regarding leeway for the tests I think Tina can test it on her machine to
confirm is 500ms is enough. I don’t see any problem if we have it up to a
few seconds.

pkg/test/auth/tokenmanager.go
sbryzak: Good catch, yes it is the same claim.  I'm wondering now though if the `ExpiresAt` claim is the actual problem here.  Isn't the problem that the token is being issued at a time that appears to the parser to be in the future?  If that is the case, wouldn't it really be the `iat` (issued at) claim that we need to apply the leeway to?  I've been working off the basis of [this comment](https://github.com/dgrijalva/jwt-go/issues/257#issuecomment-494597343) and made the changes accordingly, however on further reflection this doesn't really make sense now in the context of our issue.  What do you think?

tinakurian:
> Tina, weren’t you one who reported that time async issue between your machine (minishift) and the SSO? :) The other one was Abhishek. While I don’t think we will have this issue when running the reg services on any major cloud provider it can be an issue in dev env as it was for you and Abhishek. It’s OK if we have some manual workaround for dev environment though but I’m not sure we do. There is just one Signup button which does both login and POST /signup so we can’t control any delays here... But maybe on other click on Signup would work actually... since the user is already logged in and the token is available to be used for POST.

Yes, this was only while using minishift - Abhishek had the same problem probably because he was running using minishift as well. 

Sure... I can work with Michael maybe since it was his code we merged..? 

Sure.. i can test on my machine, the tests failed consistently for me... after setting the IAT back by 60 seconds they stopped failing.

pkg/test/auth/tokenmanager.go
tinakurian: Yeah, so when i was attempting to fix this problem changing the `exp` field didn't seem to have any impact on fixing it. It was more the `iat` field - i had set it back by 60 seconds and that seemed to solve it for me. So i think this code contains the right idea, it just needs to do it for `iat` instead of `exp.`

tinakurian:
> I've increased the leeway to 500ms, do you think that is too high? I still have problems with setting the claims, it turns out it's not so easy to convert jwt StandardClaims to MapClaims. As for the dev environment, are we able to reproduce this issue there? I would say the problem lies in [this line of code](https://github.com/codeready-toolchain/registration-service/blob/master/pkg/auth/tokenparser.go#L38), would you concur with that?

I'm not too sure if 500ms would be too little - for me 60s seemed to work, but we can try with 500ms and see what happens!

pkg/test/auth/tokenmanager.go
sbryzak: That makes way more sense to me too.  I'll keep working on trying to fix the StandardClaims / MapClaims incompatibility and make the fix to iat instead of exp.

pkg/test/auth/tokenmanager.go
```diff
@@ -144,29 +144,67 @@ func (tg *TokenManager) Key(kid string) (*rsa.PrivateKey, error) {
...
+  This section is a temporary fix until formal leeway support is available in the next jwt-go release
+
+ *****************************************************/
+
+const leeway = 500
+
+type MyClaims struct {
+	jwt.StandardClaims
+	jwt.MapClaims
+	IdentityID  string `json:"uuid,omitempty"`
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
	IdentityID        string `json:"uuid,omitempty"`
```

tinakurian: What if we did this instead: 
```
type MyMapClaims struct {
	jwt.MapClaims
}
```

pkg/test/auth/tokenmanager.go
```diff
@@ -69,21 +69,21 @@ type ExtraClaim func(token *jwt.Token)
 // WithEmailClaim sets the `email` claim in the token to generate
 func WithEmailClaim(email string) ExtraClaim {
 	return func(token *jwt.Token) {
-		token.Claims.(jwt.MapClaims)["email"] = email
+		token.Claims.(*MyClaims).Email = email
 	}
 }
 
 // WithIATClaim sets the `iat` claim in the token to generate
 func WithIATClaim(iat time.Time) ExtraClaim {
```
tinakurian: We can probably just remove this now since you've provided an alternative!

sbryzak: I think we still need this, without it there's no easy way in our tests to override the default claim values assigned to the token.

PR #64, opened 2019-11-29
chore(CRT-356): move template.Processor logic from member and support comparison of objects

MatousJobanek:
* move `template.Processor` logic from member so it can be used in host repo
* modified processor*.go files by:
  * adding "last used configuration" to an annotation - the annotation `toolchain.dev.openshift.com/last-applied-configuration` will contain the marshaled content of the resource that is being applied (the same thing as kubectl does with `kubectl.kubernetes.io/last-applied-configuration`
  * using the annotation for objects comparison so it's not updated when the object hasn't been changed - there was also kept the option to force updates no matter if the object is the same or not
  * using `runtime.Unstructured` for casting (instead of `unstructured.Unstructured`) as I came across this [comment](https://github.com/openshift/cluster-version-operator/issues/32#issuecomment-427679322)
* the filter*.go files were only moved - without any changes introduced

will be used in host-operator for applying the registration-service deployment template as well as for initial creation of the RegistrationService CR during the operator startup


pkg/template/processor.go
```diff
@@ -0,0 +1,193 @@
...
+	templatev1 "github.com/openshift/api/template/v1"
+	"github.com/openshift/library-go/pkg/template/generator"
+	"github.com/openshift/library-go/pkg/template/templateprocessing"
+	"github.com/pkg/errors"
+	errs "github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
```
golangcibot: SA1019: Package sigs.k8s.io/controller-runtime/pkg/runtime/log is deprecated: use pkg/log  (from `staticcheck`)

pkg/template/processor.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+	return string(newYaml)
+}
+
+func marshalObjectContent(newResource runtime.Object) []byte {
+	var newYaml []byte
+	var err error
+	newRes, ok := newResource.(runtime.Unstructured)
+	if ok {
+		newYaml, err = yaml.Marshal(newRes.UnstructuredContent())
```
MatousJobanek: I just realized that I should rather use json and not yaml for marshaling - I'll change it on Monday.

xcoulon:
> using runtime.Unstructured for casting (instead of unstructured.Unstructured) as I came across this comment

thanks, I knew there were 2 pkgs for the `Unstructured` type but I wasn't aware that `runtime` was the one to use! I believe we use `unstructured` in other places, too.

pkg/template/processor.go
```diff
@@ -0,0 +1,193 @@
+package template
+
+import (
+	"context"
+	"fmt"
+	"gopkg.in/yaml.v2"
```
xcoulon: we should probably reorganize the imports here (but that's something minor)

pkg/template/filter_test.go
```diff
@@ -0,0 +1,217 @@
+package template_test
+
+import (
+	"github.com/codeready-toolchain/toolchain-common/pkg/template"
```
xcoulon: we should probably reorganize the imports here (but that's something minor)

pkg/template/processor.go
```diff
@@ -0,0 +1,193 @@
...
+	metaNew, err := meta.Accessor(newResource)
+	if err != nil {
+		return false, errors.Wrapf(err, "cannot get metadata from %+v", newResource)
+	}
+
+	// creates a deepcopy of the new resource to be used to check if it already exists
+	existing := newResource.DeepCopyObject()
+
+	// set current object as annotation
+	//newWithoutAnnotation := newResource.DeepCopyObject()
```
xcoulon: could you remove this commented out line?

pkg/template/processor.go
```diff
@@ -0,0 +1,193 @@
...
+	annotations := metaNew.GetAnnotations()
+	newConfiguration := getNewConfiguration(newResource)
```
xcoulon: you don't really need to assign the result of `getNewConfiguration(newResource)` (that's a minor thing, though)

pkg/template/processor.go
MatousJobanek: changed in 7a69b5a

pkg/template/processor.go
```diff
@@ -0,0 +1,193 @@
...
+	if err != nil {
+		return false, errors.Wrapf(err, "cannot get metadata from %+v", newResource)
+	}
+
+	// check if it was changed or not
+	return originalGeneration == metaNewAfterUpdate.GetGeneration(), nil
+}
+
+func getNewConfiguration(newResource runtime.Object) string {
+	newYaml := marshalObjectContent(newResource)
```
xcoulon: ok, so the variable here should be named `newJSON`, now ;)

MatousJobanek:
> thanks, I knew there were 2 pkgs for the Unstructured type but I wasn't aware that runtime was the one to use! I believe we use unstructured in other places, too.

there are only two places where I found it:
* in the [toolchain-operator](https://github.com/codeready-toolchain/toolchain-operator/blob/master/pkg/toolchain/create_resource.go#L16)
* in the tests that are moved from member as part of this PR - let me change that here too :-)

pkg/template/processor.go
```diff
@@ -0,0 +1,193 @@
...
+	if len(newYaml) == 0 {
+		return fmt.Sprintf("%v", newResource)
+	}
+	return string(newYaml)
+}
+
+func marshalObjectContent(newResource runtime.Object) []byte {
+	var newJson []byte
+	var err error
+	newRes, ok := newResource.(runtime.Unstructured)
```
xcoulon: what about combining the 2 lines here in `if newRes, ok := newResource.(runtime.Unstructured); ok` instead?

pkg/template/processor.go
```diff
@@ -0,0 +1,193 @@
...
+	if ok {
+		newJson, err = json.Marshal(newRes.UnstructuredContent())
+		if err != nil {
+			log.Error(err, "unable to marshal the object", "object", newRes.UnstructuredContent())
```
xcoulon: aren't you supposed to return an error instead of just logging it?

pkg/template/processor.go
xcoulon: I'm not sure the `marshalObjectContent` alone makes sense. Shouldn't we move its code here in `getNewConfiguration`?

pkg/template/processor.go
```diff
@@ -0,0 +1,193 @@
...
+	newConfiguration := getNewConfiguration(newResource)
+	if annotations == nil {
+		annotations = map[string]string{}
+	}
+	annotations[LastAppliedConfigurationAnnotationKey] = newConfiguration
+	metaNew.SetAnnotations(annotations)
+
+	// gets current object (if exists)
+	namespacedName := types.NamespacedName{Namespace: metaNew.GetNamespace(), Name: metaNew.GetName()}
+	if err := p.cl.Get(context.TODO(), namespacedName, existing); err != nil {
```
xcoulon: shouldn't we check that the error was `IsNotFound` after the call to `Get` and before the call to `Create`?

pkg/template/processor_test.go
```diff
@@ -0,0 +1,807 @@
...
+		p := template.NewProcessor(cl, s)
+		tmpl, err := decodeTemplate(decoder,
+			createTemplate(withObjects(namespace, rolebinding), withParams(usernameParam, commitParam)))
+		require.NoError(t, err)
+		objs, err := p.Process(tmpl, values)
+		require.NoError(t, err)
+
+		// when adding labels and an owner reference
+		obj := objs[0]
+		nsObj, ok := obj.Object.(*unstructured.Unstructured)
```
xcoulon: should you cast to `runtime.Unstructured` here?

pkg/template/processor_test.go
```diff
@@ -0,0 +1,807 @@
...
+		assertNamespaceExists(t, cl, user)
+		assertRoleBindingExists(t, cl, user)
+
+	})
+
+	t.Run("should update existing role binding", func(t *testing.T) {
+		// given
+		cl := test.NewFakeClient(t)
+		cl.MockCreate = func(ctx context.Context, obj runtime.Object, opts ...client.CreateOption) error {
+			if rb, ok := obj.(*unstructured.Unstructured); ok {
```
xcoulon: should you cast to `runtime.Unstructured` here?

pkg/template/processor_test.go
```diff
@@ -0,0 +1,807 @@
...
+				rb.SetResourceVersion("1")
+				return cl.Client.Create(ctx, obj, opts...)
+			}
+			return cl.Client.Create(ctx, obj, opts...)
+		}
+		cl.MockUpdate = func(ctx context.Context, obj runtime.Object, opts ...client.UpdateOption) error {
+			if u, ok := obj.(*unstructured.Unstructured); ok {
```
xcoulon: should you cast to `runtime.Unstructured` here?

pkg/template/processor_test.go
```diff
@@ -0,0 +1,807 @@
...
+	assert.Equal(t, expected, actual.Object)
+}
+
+type expectedObj struct {
+	template string
+	username string
+	commit   string
+}
+
+func newObject(template, username, commit string) (*unstructured.Unstructured, error) {
```
xcoulon: should you return a `runtime.Unstructured` here?

pkg/test/client.go
```diff
@@ -2,8 +2,8 @@ package test
 
 import (
 	"context"
+	"github.com/codeready-toolchain/api/pkg/apis"
```
xcoulon: we should probably reorganize the imports

MatousJobanek:
> in the tests that are moved from member as part of this PR - let me change that here too :-)

ok, I changed it only in the case of casting - I believe that the creation is fine 657e55c

pkg/template/processor.go
MatousJobanek: imports, imports, bloody imports :hankey: :gun: 
could you please remind me what is the expected format again? I always forgot that as I only use "the format option" in IDE

pkg/template/processor.go
MatousJobanek: zombie :ghost:  - removed fa28d50

pkg/template/processor.go
MatousJobanek: Yes, I do because it's then used here https://github.com/codeready-toolchain/toolchain-common/pull/64/files#diff-ae1e62ddb6744c2036a1b470ddc6a527R132

pkg/template/processor.go
MatousJobanek: zombie :ghost: - renamed f66ba6b

pkg/template/processor.go
MatousJobanek: done ef439ab

MatousJobanek: I changed it a bit more: 8720488

pkg/template/processor.go
MatousJobanek: yes I'm :-) see 8720488

pkg/template/processor.go
MatousJobanek: I believe that after these changes 8720488 it makes  sense to keep it in separated function

pkg/template/processor.go
MatousJobanek: well, I had it originally in the first version of my implementation, but then I realized why it should try to create the resource anyway - no matter what was the reason for the error. 
I do remember that in the tenant service I got an error of missing RBAC for resources that were not found. However, I'm not sure if that was because of OpenShift 3, because I used http requests instead of the go client or because of any other reason...

pkg/template/processor_test.go
MatousJobanek: ah, I missed this one - fixed here 521794c but didn't use any `*.Unstructured` at all ;-)

pkg/template/processor_test.go
MatousJobanek: that was already replaced here 657e55c  - you reviewed old code

pkg/template/processor_test.go
MatousJobanek: that was already replaced here 657e55c  - you reviewed old code

pkg/template/processor_test.go
MatousJobanek: that was already replaced here 657e55c  - you reviewed old code

pkg/template/processor.go
xcoulon: it should be something like this:
```
import (
    <sdtlib package>
    <sdtlib package>
    ...
    <blank line>
    <our package>
    <our package>
    ...
    <third party package>
    <third party package>
    ...
)
```

pkg/template/processor.go
xcoulon: ah ok, I was missing the rest of the func body :/

pkg/template/processor.go
MatousJobanek: does it look better? 70877f8

pkg/template/processor.go
alexeykazakov: I'm not sure about that particular case you got in tenant service and in Kube in general but it's considered to be a good practice to return Auth errors instead of 404 (or any other NotFound error) in case of missing resources for clients which doesn't have permissions to read the resource. It's done for security reasons so if you don't have permissions to read resources you won't be able to check if the resource exists or not.

alexeykazakov: But if the client does have enough permissions but still gets RBAC errors just because the resource is missing then it looks like a bug :)

MatousJobanek: Yes, that was exactly the case when I got the auth error n tenant service, but the user we use in there should have enough permissions. However, it could be a different behavior for users and SA. 
Anyway, I changed the implementation back to the original version so it's trying to create the resource only when it gets "NotFound" - see 914a786

xcoulon: ok, thanks @MatousJobanek

PR #71, opened 2020-01-03
fix(CRT-397): use standard informer for KubeFedCluster_cache controller

MatousJobanek:
Hopefully, it should limit the number of errors in cluster logs
There are two PRs opened to verify that it's working for both operators host and member
https://github.com/codeready-toolchain/member-operator/pull/112
https://github.com/codeready-toolchain/host-operator/pull/129

### DIFF
--- a/pkg/controller/kubefedcluster_caching.go
+++ b/pkg/controller/kubefedcluster_caching.go
@@ -2,7 +2,10 @@ package controller
 
 import (
 	"github.com/codeready-toolchain/toolchain-common/pkg/cluster"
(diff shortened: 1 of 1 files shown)


MatousJobanek:
After trying it out several times over and over again for a few hours I figured out that the GenericInformer is not the reason for the annoying errors in our logs. 
See these PRs for the real fix:
https://github.com/codeready-toolchain/member-operator/pull/113
https://github.com/codeready-toolchain/host-operator/pull/130

PR #74, opened 2020-02-06
chore: move ApplySingle to separated struct

MatousJobanek:
move ApplySingle to separated struct so I can use:
```go
applyClient := applyCl.NewApplyClient(cl, scheme)
applyClient.CreateOrUpdateObject(.....)
```


pkg/client/client.go
```diff
@@ -0,0 +1,135 @@
...
+// NewApplyClient returns a new ApplyClient
+func NewApplyClient(cl client.Client, scheme *runtime.Scheme) *ApplyClient {
+	return &ApplyClient{cl: cl, scheme: scheme}
+}
+
+// CreateOrUpdateObject creates the object if is missing and if the owner object is provided, then it's set as a controller reference.
+// If the objects exists then when the spec content has changed (based on the content of the annotation in the original object) then it
+// is automatically updated. If it looks to be same then based on the value of forceUpdate param it updates the object or not.
+// The return boolean says if the object was either created or updated
+func (p ApplyClient) CreateOrUpdateObject(obj runtime.Object, forceUpdate bool, owner v1.Object) (bool, error) {
```
xcoulon: do we really need to wrap the call to `createOrUpdateObj` in `CreateOrUpdateObject`?

MatousJobanek: any other idea?

PR #87, opened 2020-03-09
feat: add assertions to check tier in MUR

MatousJobanek:
add assertions to check tier in MUR

### DIFF
--- a/pkg/test/masteruserrecord/assertion.go
+++ b/pkg/test/masteruserrecord/assertion.go
@@ -92,6 +92,42 @@ func (a *Assertion) AllUserAccountsHaveCondition(expected toolchainv1alpha1.Cond
 	return a
 }
 
+func (a *Assertion) AllUserAccountsHaveTier(tierName string, namespaces []toolchainv1alpha1.NSTemplateTierNamespace) *Assertion {
+	err := a.loadUaAssertion()
+	require.NoError(a.t, err)
+	for _, ua := range a.masterUserRecord.Spec.UserAccounts {
+		a.userAccountHasTier(ua, tierName, namespaces)
+	}
+	return a
+}
+
+func (a *Assertion) UserAccountHasTier(targetCluster, tierName string, namespaces []toolchainv1alpha1.NSTemplateTierNamespace) *Assertion {
+	err := a.loadUaAssertion()
+	require.NoError(a.t, err)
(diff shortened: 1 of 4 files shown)


pkg/test/masteruserrecord/assertion.go
```diff
@@ -92,6 +92,42 @@ func (a *Assertion) AllUserAccountsHaveCondition(expected toolchainv1alpha1.Cond
...
+func (a *Assertion) AllUserAccountsHaveTier(tierName string, namespaces []toolchainv1alpha1.NSTemplateTierNamespace) *Assertion {
+	err := a.loadUaAssertion()
+	require.NoError(a.t, err)
+	for _, ua := range a.masterUserRecord.Spec.UserAccounts {
+		a.userAccountHasTier(ua, tierName, namespaces)
+	}
+	return a
+}
+
+func (a *Assertion) UserAccountHasTier(targetCluster, tierName string, namespaces []toolchainv1alpha1.NSTemplateTierNamespace) *Assertion {
```
golangcibot: `(*Assertion).UserAccountHasTier` - `namespaces` is unused (from `unparam`)

pkg/test/masteruserrecord/assertion.go
```diff
@@ -92,6 +92,42 @@ func (a *Assertion) AllUserAccountsHaveCondition(expected toolchainv1alpha1.Cond
 	return a
 }
 
+func (a *Assertion) AllUserAccountsHaveTier(tierName string, namespaces []toolchainv1alpha1.NSTemplateTierNamespace) *Assertion {
```
alexeykazakov: Why not just pass NSTemplateTier here?
Something like:
```suggestion
func (a *Assertion) AllUserAccountsHaveTier(tier NSTemplateTier) *Assertion {
```

MatousJobanek: yes, I was thinking about providing the whole `NSTemplateTier` object, but I decided for two params just for the sake of the simplicity of calling this assertion method. The reason is that setting the name in the `NSTemplateTier` is not obvious and requires creation of `ObjectMeta` object that requires additional lines of code.

MatousJobanek: ok, looking at all usages of the assertion methods I have to admit that I can easily pass the whole `NSTemplateTier` object. So changed 262d7bb

PR #88, opened 2020-03-13
fix test on updating resource

xcoulon:

### DIFF
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -95,7 +95,7 @@ func TestApplySingle(t *testing.T) {
 			require.NoError(t, err)
 			assert.True(t, createdOrChanged) // resource was updated on the server, so returned value if `true`
 			updateGeneration := modifiedObj.(*corev1.Service).GetGeneration()
-			assert.Equal(t, updateGeneration, originalGeneration+1)
+			assert.Equal(t, originalGeneration+1, updateGeneration)
 		})
 
 		t.Run("when using forceUpdate=false, it should update when spec is different", func(t *testing.T) {
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -63,10 +63,10 @@ func TestApplySingle(t *testing.T) {
 		t.Run("when using forceUpdate=true, it should not update when specs are same", func(t *testing.T) {
 			// given
 			cl, _ := newClient(t, s)
-			obj := defaultService.DeepCopyObject()
+			obj := defaultService.DeepCopy()
 			_, err := cl.CreateOrUpdateObject(obj, true, nil)
 			require.NoError(t, err)
-			originalGeneration := obj.(*corev1.Service).GetGeneration()
+			originalGeneration := obj.GetGeneration()
 
 			// when updating with the same obj again
 			createdOrChanged, err := cl.CreateOrUpdateObject(obj, true, nil)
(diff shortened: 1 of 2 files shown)


pkg/client/client_test.go
```diff
@@ -164,6 +185,40 @@ func TestApplySingle(t *testing.T) {
 	})
 }
 
+func newClient(t *testing.T, s *runtime.Scheme) (*applyCl.ApplyClient, *test.FakeClient) {
+	cli := NewFakeClient(t)
+	cli.MockCreate = func(ctx context.Context, obj runtime.Object, opts ...client.CreateOption) error {
+		// force the Generation to `1` for newly created objects
+		if obj, ok := obj.(*unstructured.Unstructured); ok {
```
MatousJobanek: This won't work - you cannot cast to `unstructured.Unstructured` - it will always return false.
you should use the `meta.Accessor` instead:
```go
ob, err := meta.Accessor(obj)
if err != nil {
	return err
}
ob.SetGeneration(1)
```

pkg/client/client_test.go
```diff
@@ -57,25 +60,47 @@ func TestApplySingle(t *testing.T) {
...
-			createdOrChanged, err := cl.CreateOrUpdateObject(modifiedService.DeepCopyObject(), true, nil)
+			// when updating with the same obj again
+			createdOrChanged, err := cl.CreateOrUpdateObject(obj, true, nil)
 
 			// then
 			require.NoError(t, err)
-			assert.True(t, createdOrChanged)
+			assert.False(t, createdOrChanged) // resource was not updated on the server, so returned value is `false`
+			updateGeneration := obj.(*corev1.Service).GetGeneration()
+			assert.Equal(t, updateGeneration, originalGeneration)
```
MatousJobanek: it's the other way around - the first parameter should be expected value and the second the actual value:
```suggestion
			assert.Equal(t, originalGeneration, updateGeneration)
```

pkg/client/client_test.go
```diff
@@ -57,25 +60,47 @@ func TestApplySingle(t *testing.T) {
...
+			// when updating with the modified obj
+			modifiedObj := modifiedService.DeepCopyObject()
+			modifiedObj.(*corev1.Service).ObjectMeta.Generation = obj.(*corev1.Service).GetGeneration()
+			createdOrChanged, err := cl.CreateOrUpdateObject(modifiedObj, true, nil)
+
+			// then
+			require.NoError(t, err)
+			assert.True(t, createdOrChanged) // resource was updated on the server, so returned value if `true`
+			updateGeneration := modifiedObj.(*corev1.Service).GetGeneration()
+			assert.Equal(t, updateGeneration, originalGeneration+1)
```
MatousJobanek: same here: 
```suggestion
			assert.Equal(t, originalGeneration+1, updateGeneration)
```

pkg/client/client_test.go
```diff
@@ -57,25 +60,47 @@ func TestApplySingle(t *testing.T) {
...
+		t.Run("when using forceUpdate=true, it should not update when specs are same", func(t *testing.T) {
 			// given
-			cli := NewFakeClient(t)
-			cl := applyCl.NewApplyClient(cli, s)
-			_, err := cl.CreateOrUpdateObject(defaultService.DeepCopyObject(), true, nil)
+			cl, _ := newClient(t, s)
+			obj := defaultService.DeepCopyObject()
+			_, err := cl.CreateOrUpdateObject(obj, true, nil)
 			require.NoError(t, err)
+			originalGeneration := obj.(*corev1.Service).GetGeneration()
```
MatousJobanek: if we use just `defaultService.DeepCopy()` then we don't have to cast it back here and just use `obj.GetGeneration()`

pkg/client/client_test.go
```diff
@@ -164,6 +185,40 @@ func TestApplySingle(t *testing.T) {
...
+	}
+	cli.MockUpdate = func(ctx context.Context, obj runtime.Object, opts ...client.UpdateOption) error {
+		// compare the specs (only) and only increment the generation if something changed
+		// (the server will check the object metadata, but we're skipping this here)
+		if svc, ok := obj.(*corev1.Service); ok {
+			existing := corev1.Service{}
+			if err := cli.Get(ctx, types.NamespacedName{Namespace: svc.GetNamespace(), Name: svc.GetName()}, &existing); err != nil {
+				return err
+			}
+			if !reflect.DeepEqual(existing.Spec, svc.Spec) { // Service has a `spec` field
```
MatousJobanek: may I have one proposal? could we change this to:
```suggestion
			if !reflect.DeepEqual(existing.Spec, svc.Spec) || !reflect.DeepEqual(existing.ObjectMeta.Labels, svc.ObjectMeta.Labels) { // Service has a `spec` field
```
and add a new test case that is proposed below (or maybe above :-))?

pkg/client/client_test.go
```diff
@@ -57,25 +60,47 @@ func TestApplySingle(t *testing.T) {
...
+			createdOrChanged, err := cl.CreateOrUpdateObject(obj, true, nil)
 
 			// then
 			require.NoError(t, err)
-			assert.True(t, createdOrChanged)
+			assert.False(t, createdOrChanged) // resource was not updated on the server, so returned value is `false`
+			updateGeneration := obj.(*corev1.Service).GetGeneration()
+			assert.Equal(t, updateGeneration, originalGeneration)
+		})
+
```
MatousJobanek:
```suggestion

		t.Run("when using forceUpdate=true, it will update even when specs are same and return true as labels have changed", func(t *testing.T) {
			// given
			cl, _ := newClient(t, s)
			service := defaultService.DeepCopy()
			_, err := cl.CreateOrUpdateObject(service, true, nil)
			require.NoError(t, err)
			originalGeneration := service.GetGeneration()
			service.SetLabels(map[string]string{"new": "label"})

			// when updating with changed labels
			createdOrChanged, err := cl.CreateOrUpdateObject(service, true, nil)

			// then
			require.NoError(t, err)
			assert.True(t, createdOrChanged) // resource was updated on the server, so returned value is `true`
			updateGeneration := service.GetGeneration()
			assert.Equal(t, originalGeneration+1, updateGeneration)
		})
```

pkg/client/client_test.go
xcoulon: ah thanks! and yes, I saw it failed, but I thought I used something else instead 🤦‍♂

xcoulon: fixed in cd909f5

pkg/client/client_test.go
xcoulon: no, we can't: `DeepCopy` returns a `runtime.Object` interface which has no suc `GetGeneration()` func :/

pkg/client/client_test.go
xcoulon: what would be the benefit of this test? I just wanted to mimic the server behavior but in a simpler form.

pkg/client/client_test.go
MatousJobanek: see my proposal for the new test - I don't have to cast anything there ;-)

xcoulon: ahhh, I did not pay enough attention to the methods: you suggested `DeepCopy()` instead of `DeepCopyObject()` (which returns a `runtime.Object` and not a `corev1.Service`)

xcoulon: done in 92b60fa

pkg/client/client_test.go
MatousJobanek: > what would be the benefit of this test? 

I wanted to have a way to test case when the `CreateOrUpdateObject` based on the changed generation value

> I just wanted to mimic the server behavior but in a simpler form.

yes, you are right. I thought that the generation is incremented also for changes in labels and annotations, but I found out that it is incremented only for changes in spec. 
So keep it as it is and you can ignore my proposal ;-)

xcoulon:
> Nice! 👍

thanks, and thanks for the help and feedback, too ;)

PR #89, opened 2020-03-17
Update Generation in Fake client

alexeykazakov:
Some existing tests relay on resource Generation changes which is currently are ignored by the fake client.
This PR https://github.com/codeready-toolchain/toolchain-common/pull/88/files#diff-fb87476fa66849c8514ba700e715b4b6R108
fixed a bug in `ApplyClient` which actually broke some of our tests (sounds weird I know).
Let's add generation updates in our fake client so it's available in all our tests.

### DIFF
--- a/pkg/test/client.go
+++ b/pkg/test/client.go
@@ -148,7 +148,7 @@ func cleanObject(obj runtime.Object) (runtime.Object, error) {
 		return nil, err
 	}
 
-	for k, _ := range m {
+	for k := range m {
 		if k != "metadata" && k != "kind" && k != "apiVersion" {
 			m[k] = nil
 		}
--- a/pkg/test/client.go
+++ b/pkg/test/client.go
@@ -105,8 +105,19 @@ func (c *FakeClient) Update(ctx context.Context, obj runtime.Object, opts ...cli
 	if err != nil {
 		return err
 	}
+	updatingMap, err := toMap(obj)
+	if err != nil {
+		return err
+	}
+	updatingMap["metadata"] = nil
+	updatingMap["status"] = nil
+	updatingMap["kind"] = nil
+	updatingMap["apiVersion"] = nil
 
-	current := obj.DeepCopyObject()
+	current, err := cleanObject(obj)
+	if err != nil {
+		return err
+	}
 	if err := c.Client.Get(ctx, types.NamespacedName{Namespace: updatingMeta.GetNamespace(), Name: updatingMeta.GetName()}, current); err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


alexeykazakov:
On the other hand it's probably too tricky to implement updating generation properly for arbitrary objects...
Let's make it as WIP for now to see if we can really make it generic enough.

pkg/test/client.go
```diff
@@ -87,6 +98,27 @@ func (c *FakeClient) Update(ctx context.Context, obj runtime.Object, opts ...cli
...
+		}
+	} else if cm, ok := obj.(*corev1.ConfigMap); ok {
+		existing := corev1.ConfigMap{}
+		if err := c.Client.Get(ctx, types.NamespacedName{Namespace: cm.GetNamespace(), Name: cm.GetName()}, &existing); err != nil {
+			return err
+		}
+		if !reflect.DeepEqual(existing.Data, cm.Data) { // ConfigMap has a `data` field
+			cm.SetGeneration(existing.GetGeneration() + 1)
+		}
+	}
```
MatousJobanek: This should work and should be more generic:
```go
updatingMeta, err := meta.Accessor(obj)
if err != nil {
	return err
}

current := obj.DeepCopyObject()
if err := c.Client.Get(ctx, types.NamespacedName{Namespace: updatingMeta.GetNamespace(), Name: updatingMeta.GetName()}, current); err != nil {
	return err
}
currentMeta, err := meta.Accessor(current)
if err != nil {
	return err
}

updatingContent, err := json.Marshal(obj)
if err != nil {
	return err
}
updatingMap := map[string]interface{}{}
if err := json.Unmarshal(updatingContent, &updatingMap); err != nil {
	return err
}
currentContent, err := json.Marshal(current)
if err != nil {
	return err
}
currentMap := map[string]interface{}{}
if err := json.Unmarshal(currentContent, &currentMap); err != nil {
	return err
}

if updatingMap["spec"] != nil {
	if !reflect.DeepEqual(updatingMap["spec"], currentMap["spec"]) {
		updatingMeta.SetGeneration(currentMeta.GetGeneration() + 1)
	}
} else if updatingMap["data"] != nil {
	if !reflect.DeepEqual(updatingMap["data"], currentMap["data"]) {
		updatingMeta.SetGeneration(currentMeta.GetGeneration() + 1)
	}
}
```

alexeykazakov: Thanks! It doesn't cover all the use cases but works most of them.
Updated in https://github.com/codeready-toolchain/toolchain-common/pull/89/commits/e2334fbfe888e9c8daddf1610b4bc730e3240e21

MatousJobanek: we could maybe cover more cases with this version
```go
	updatingMap["metadata"] = nil
	updatingMap["status"] = nil
	updatingMap["kind"] = nil
	updatingMap["apiVersion"] = nil

	currentMap["metadata"] = nil
	currentMap["status"] = nil
	currentMap["kind"] = nil
	currentMap["apiVersion"] = nil

	if !reflect.DeepEqual(updatingMap, currentMap) {
		updatingMeta.SetGeneration(currentMeta.GetGeneration() + 1)
	}
```

pkg/test/client.go
```diff
@@ -87,9 +98,78 @@ func (c *FakeClient) Update(ctx context.Context, obj runtime.Object, opts ...cli
...
 
+func cleanObject(obj runtime.Object) (runtime.Object, error) {
+	newObj := obj.DeepCopyObject()
+
+	m, err := toMap(newObj)
+	if err != nil {
+		return nil, err
+	}
+
+	for k, _ := range m {
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
	for k := range m {
```

pkg/test/client.go
alexeykazakov: Yes, it's a good idea! I also added some cleanup for the object we use an empty one for loading the current resource to avoid the situation when we have some leftovers: https://github.com/codeready-toolchain/toolchain-common/pull/89/commits/c4811ce8458f9c9cfabc627dafe31c46fd1515f3

PR #91, opened 2020-03-30
add NSTemplateSet fluent assertions

xcoulon:
also, rename `loadUaAssertion()` func to `loadMasterUserRecord()` in `pkg/test/masteruserrecord/assertion.go`

### DIFF
--- a/pkg/test/masteruserrecord/assertion.go
+++ b/pkg/test/masteruserrecord/assertion.go
@@ -19,7 +19,7 @@ type Assertion struct {
 	t                test.T
 }
 
-func (a *Assertion) loadUaAssertion() error {
+func (a *Assertion) loadMasterUserRecord() error {
 	if a.masterUserRecord != nil {
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


xcoulon:
> Looks good. Thanks for the fix of the wrong copy-pasted `loadUaAssertion` method :-)
> 
> Just one idea - I guess that you are going to use the new assertion for the verification of the new `ClusterResources` field/template - don't you wanna reuse the functions that are in MUR assertion and create a method:
> 
> ```go
> func (a *Assertion) HasTierTemplates(targetCluster string, expectations ...NsTemplateSetSpecExp) *Assertion {
> 	err := a.loadNSTemplateSet()
> 	require.NoError(a.t, err)
> 	expectedTmplSetSpec := &toolchainv1alpha1.NSTemplateSetSpec{}
> 	for _, modify := range expectations {
> 		modify(expectedTmplSetSpec)
> 	}
> 	assert.Equal(a.t, *expectedTmplSetSpec, a.nsTmplSet.Spec)
> 	return a
> }
> ```
> 
> or directly (for the cases when you would refer to values taken from `UserAccount`)
> 
> ```go
> func (a *Assertion) HasSpec(targetCluster string, expSpec toolchainv1alpha1.NSTemplateSetSpec) *Assertion {
> 	err := a.loadNSTemplateSet()
> 	require.NoError(a.t, err)
> 	assert.Equal(a.t, expSpec, a.nsTmplSet.Spec)
> 	return a
> }
> ```
> 
> both of them could be handy :-)

thanks for the suggestion @MatousJobanek! For now, I just wanted to refactor the existing tests with these new assertions, and I'll probably add more when working on the tests related to the cluster resources, indeed.

PR #96, opened 2020-04-07
refactor: return create/update status when applying tmpl objects

xcoulon:
This will be useful to determine if objects were created or updated by our operators.

### DIFF
--- a/pkg/template/processor_test.go
+++ b/pkg/template/processor_test.go
@@ -249,10 +249,11 @@ func TestProcessAndApply(t *testing.T) {
 		require.NoError(t, err)
 
 		// when
-		err = p.Apply(objs)
+		createdOrUpdated, err := p.Apply(objs)
 
 		// then
 		require.NoError(t, err)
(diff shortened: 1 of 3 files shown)


pkg/template/processor_test.go
```diff
@@ -340,6 +344,27 @@ func TestProcessAndApply(t *testing.T) {
...
+		require.NoError(t, err)
+		created, err := p.Apply(objs)
+		require.NoError(t, err)
+		assert.True(t, created)
+		assertNamespaceExists(t, cl, user)
+		assertRoleBindingExists(t, cl, user)
+
+		// when apply the same template again
+		updated, err := p.Apply(objs)
+		require.NoError(t, err)
```
MatousJobanek, line 364: please, add the `// then` comment
```suggestion

		// then
		require.NoError(t, err)
```

xcoulon: ah yes! thanks!

xcoulon:
thanks for the review and approval, @MatousJobanek 🙌

PR #99, opened 2020-04-17
feat: when no FedCluster is found then refresh the cache

MatousJobanek:
When no FedCluster of the given type is found, then refresh the cache. This is for the cases 
* when there is a rolling update 
* when a pod panicked and a new one started 

so the member operator can immediately start operating with the already existing resources. 
Otherwise, the provisioning/update may fail with an error: `unable to connect to the host cluster: unknown cluster`

paired PRs:
https://github.com/codeready-toolchain/host-operator/pull/185
https://github.com/codeready-toolchain/member-operator/pull/145

### DIFF
--- a/pkg/cluster/cache_whitebox_test.go
+++ b/pkg/cluster/cache_whitebox_test.go
@@ -1,6 +1,7 @@
 package cluster
 
 import (
+	"sync"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


pkg/cluster/cache_whitebox_test.go
```diff
@@ -175,6 +175,25 @@ func TestGetClustersByType(t *testing.T) {
 			assert.Contains(t, clusters, member2)
 		})
 
+		t.Run("found after refreshing the cache", func(t *testing.T) {
+			// given
+			defer resetClusterCache()
+			member := newTestFedCluster("member", Member, ready)
+			called := false
+			clusterCache.refreshCache = func() {
```
xcoulon, line 183: should you also reset the `clusterCache.refreshCache` in `resetClusterCache()` ?

MatousJobanek: this is done by `defer resetClusterCache()` - check the function https://github.com/codeready-toolchain/toolchain-common/blob/21d2a6b78d83291625f7dcc54cca5b26f6bf954f/pkg/cluster/cache_whitebox_test.go#L420-L422
it creates a brand new instance of the `clusterCache`

MatousJobanek:
> let's add a good test which will spin up a few go routines which will try to mess with the cache (read, add, delete) very intensely (thousands of attempts) for a short period of time.

done 401ed68

PR #100, opened 2020-04-24
feat: move apply to client and assign labels when applying

MatousJobanek:
proposed changes:
* moved apply method from processor to client - it makes more sense to have it there
* also moved the related tests
* added "labels" parameter to the apply method so we can let the client take care of it and we can make the NSTemplateSet controller a bit simpler :-) 
* decreased a bit the concurrent test of fedcluster cache - it was too much and my PC had problems with the number of concurrent goroutines

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -134,3 +134,39 @@ func (p ApplyClient) createObj(newResource runtime.Object, metaNew v1.Object, ow
 	}
 	return p.cl.Create(context.TODO(), newResource)
 }
+
+// Apply applies the objects, ie, creates or updates them on the cluster
+// returns `true, nil` if at least one of the objects was created or modified,
+// `false, nil` if nothing changed, and `false, err` if an error occurred
+func (p ApplyClient) Apply(objs []runtime.RawExtension, newLabels map[string]string) (bool, error) {
+	createdOrUpdated := false
+	for _, rawObj := range objs {
+		obj := rawObj.Object
+		if obj == nil {
+			continue
+		}
+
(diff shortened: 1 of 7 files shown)


MatousJobanek:
> This is a breaking change. I guess you are working on Host/Member PRs to adopt this change? :)

done:
https://github.com/codeready-toolchain/member-operator/pull/148
https://github.com/codeready-toolchain/host-operator/pull/190

pkg/client/client.go
```diff
@@ -134,3 +134,39 @@ func (p ApplyClient) createObj(newResource runtime.Object, metaNew v1.Object, ow
...
+// `false, nil` if nothing changed, and `false, err` if an error occurred
+func (p ApplyClient) Apply(objs []runtime.RawExtension, newLabels map[string]string) (bool, error) {
+	createdOrUpdated := false
+	for _, rawObj := range objs {
+		obj := rawObj.Object
+		if obj == nil {
+			continue
+		}
+
+		acc, err := meta.Accessor(obj)
```
xcoulon, line 149: we could probably skip that part if `newLabels` is empty

MatousJobanek: ah, I merged the PR before you added the comment - sorry for not waiting a bit longer :pray: 
I can either open another PR with the change or add the change next time when I'll be changing the common repo. 
And thanks a lot for your review :-) :+1:

xcoulon: no worries, this is just an optimisation, so it can be addressed in a separate PR, indeed ;)

PR #111, opened 2020-06-15
Assert labels on MUR accounts

xcoulon:
Verify that the MUR has labels for associated
with the given tier. There are 2 labels for expect:
- the tier name
- a hash based on the template refs in the tier

Updates https://issues.redhat.com/browse/CRT-683


pkg/test/masteruserrecord/master_user_record.go
```diff
@@ -16,46 +19,106 @@ import (
...
 			UserAccounts: []toolchainv1alpha1.UserAccountEmbedded{newEmbeddedUa(test.MemberClusterName)},
 		},
 	}
 	Modify(mur, modifiers...)
 	return mur
 }
 
+// templateTierNameLabel returns the label key to specify the tier templates in use on the given cluster
+func templateTierNameLabel(cluster string) string {
+	return toolchainv1alpha1.LabelKeyPrefix + cluster + "-templates-tier"
```
alexeykazakov, line 94: The same question as in https://github.com/codeready-toolchain/host-operator/pull/221#discussion_r439080801
Do we really have to use cluster names in these labels?

xcoulon: I responded in https://github.com/codeready-toolchain/host-operator/pull/221#discussion_r440610354 and I'll apply the necessary changes here as well ;)

pkg/test/masteruserrecord/master_user_record.go
```diff
@@ -16,46 +19,113 @@ import (
...
+func templateTierHashLabelKey(tierName string) string {
+	return toolchainv1alpha1.LabelKeyPrefix + tierName + "-tier-hash"
+}
+
+type templateRefs struct {
+	Namespaces       []string `json:"namespaces"`
+	ClusterResources string   `json:"clusterresource,omitempty"`
+}
+
+// ComputeTemplateRefsHash computes the hash of the `.spec.namespaces[].templateRef` + `.spec.clusteResource.TemplateRef`
```
alexeykazakov, line 101:
```suggestion
// computeTemplateRefsHash computes the hash of the `.spec.namespaces[].templateRef` + `.spec.clusteResource.TemplateRef`
```

pkg/test/masteruserrecord/master_user_record.go
```diff
@@ -1,70 +1,155 @@
...
+	murs := make([]runtime.Object, size)
+	for i := 0; i < size; i++ {
+		murs[i] = NewMasterUserRecord(t, fmt.Sprintf(nameFmt, i), modifiers...)
+	}
+	return murs
+}
+
+func NewMasterUserRecord(t *testing.T, userName string, modifiers ...MurModifier) *toolchainv1alpha1.MasterUserRecord {
+	userID := uuid.NewV4().String()
+	hash, _ := computeTemplateRefsHash(DefaultNSTemplateTier) // we can assume the JSON marshalling will always work
```
MatousJobanek, line 84: if we can ignore the error here, why don't we ignore the error also in `Account`? I know that this is predefined but still, it looks weird. 
In addition, we have `*testing.T` instance here, so we can easily check the error also in this function.
Just to be clear - I don't wanna force you to do it (the probability of the error is really low), I just have a bad feeling when we ignore an error somewhere - that's why I shared my thought/concern.

pkg/test/templateupdaterequest/templateupdaterequest.go
```diff
@@ -0,0 +1,84 @@
+package templateupdaterequest
```
MatousJobanek, line 1: I'm curious - why do we have these functions/file here and not in host-operator repo?
I'm asking since I expect that the function should be used only in host-operator - having it here makes any future modification more difficult.

pkg/test/templateupdaterequest/templateupdaterequest.go
```diff
@@ -0,0 +1,84 @@
...
+
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
+	"github.com/codeready-toolchain/toolchain-common/pkg/test"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// NewTemplateUpdateRequest creates a specified number of TemplateRequestUpdate objects, with options
```
MatousJobanek, line 15:
```suggestion
// NewTemplateUpdateRequest creates a new instance of TemplateUpdateRequest with the given name and modifies it with the given options
```

pkg/test/templateupdaterequest/templateupdaterequest.go
```diff
@@ -0,0 +1,84 @@
...
+			},
+		},
+	}
+	for _, opt := range options {
+		opt.applyToTemplateUpdateRequest(r)
+	}
+	return r
+}
+
+// NewTemplateUpdateRequests creates a specified number of TemplateRequestUpdate objects, with options
```
MatousJobanek, line 32:
```suggestion
// NewTemplateUpdateRequests creates a specified number of TemplateUpdateRequest objects, with options
```

pkg/test/masteruserrecord/assertion.go
```diff
@@ -194,3 +200,21 @@ func (a *Assertion) DoesNotHaveFinalizer() *Assertion {
...
+func (a *Assertion) DoesNotHaveLabel(key string) *Assertion {
+	err := a.loadUaAssertion()
+	require.NoError(a.t, err)
+	assert.NotContains(a.t, a.masterUserRecord.Labels, key)
+	return a
+}
+
+// HasLabel verifies that the MasterUserRecord has
+// a label with the given key
+func (a *Assertion) HasLabel(key string) *Assertion {
```
MatousJobanek, line 215: it would be also good to have an assertion:
```go
 func (a *Assertion) HasLabelWithValue(key, value string) *Assertion {
```
that would verify also the value. This could be used here https://github.com/codeready-toolchain/host-operator/pull/221/files#diff-dd63ae5743a88e29925009afded0e819R85 to verify that the label still exists with the same value as was before

MatousJobanek: never mind, I've just realized that the value is verified as part of `UserAccountHasTier`

pkg/test/masteruserrecord/master_user_record.go
xcoulon: I believe I ignored the error at first for the sake for simplicity (ie, not returning an error, especially for test data), and I only added the `t *testing` param afterwards, so yes, I can to the checking now

xcoulon: done in 97d17cc

pkg/test/templateupdaterequest/templateupdaterequest.go
xcoulon: because we have similar funcs for MasterUserRecord. But yes, it's harder to maintain.

pkg/test/masteruserrecord/assertion.go
xcoulon: ok, marking as resolved for now then

pkg/test/templateupdaterequest/templateupdaterequest.go
MatousJobanek: The MUR and UserAccount functions/assertions are here because they are used in both repos host-operator and member-operator. The rest of such functions stays in the repos where they are used. 
I don't really see any benefit in having it here.

pkg/test/masteruserrecord/master_user_record.go
```diff
@@ -74,36 +159,89 @@ func ModifyUaInMur(mur *toolchainv1alpha1.MasterUserRecord, targetCluster string
...
+				UserAccountSpecBase: toolchainv1alpha1.UserAccountSpecBase{
+					NSTemplateSet: templates,
+				},
+			},
+		})
+		// set the labels for the tier templates in use
+		hash, err := computeTemplateRefsHash(tier)
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
xcoulon: I don't understand what's wrong here: I'm configuring the MasterUserRecord with an account which uses an outdated tier. When the tier will be updated, I can select this particular MasterUserRecord because its "hash" label value changed. What's wrong with that?

pkg/test/templateupdaterequest/templateupdaterequest.go
xcoulon: sigh... ok, I'll move this package in the host-operator then 😬

xcoulon: done in 100644

pkg/test/masteruserrecord/master_user_record.go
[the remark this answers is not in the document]
MatousJobanek: never mind - I didn't have the context of the usage in host-operator. I didn't realize that you would add a different tier than is used in the first UserAccount

xcoulon:
> Thanks a lot for moving the `templateupdaterequest.go` to the host-operator - it will make our lives easier.
> 👍

well, thanks for suggesting it. I felt the pain last week but I was too buried to undertake the change... now it's done ;)

PR #113, opened 2020-06-19
additional MUR modifier and extra condition

xcoulon:


xcoulon:
> Looks good. But since you are modifying some existing test assertions make sure you don't break any existing unit tests in both operators.

oops, I pushed-force a commit which only contained an update on the `api` module. I did not realize that you already reviewed the PR :|

pkg/test/masteruserrecord/master_user_record.go
```diff
@@ -189,19 +189,41 @@ func TargetCluster(targetCluster string) MurModifier {
...
+// AdditionalAccount sets an additional account on the MasterUserRecord
+func AdditionalAccount(cluster string, tier toolchainv1alpha1.NSTemplateTier, options ...AccountOption) MurModifier {
 	return func(mur *toolchainv1alpha1.MasterUserRecord) error {
-		// set the user account
 		templates := nstemplateSetFromTier(tier)
-		mur.Spec.UserAccounts = append(mur.Spec.UserAccounts, toolchainv1alpha1.UserAccountEmbedded{
-			TargetCluster: "member-cluster",
+		ua := toolchainv1alpha1.UserAccountEmbedded{
+			TargetCluster: cluster,
+			SyncIndex:     "1", // default value
```
MatousJobanek, line 214: Could you please align the values? 
https://github.com/codeready-toolchain/toolchain-common/blob/3a94d31a9796c5c5fb2efe5f42bfaf0618924d0f/pkg/test/masteruserrecord/master_user_record.go#L137

MatousJobanek: or, if you slightly change the `newEmbeddedUa` function (to take the given NSTemplateSet), then you could use it here... Then we would have only one place where we initialize the UserAccount

pkg/test/masteruserrecord/master_user_record.go
```diff
@@ -189,19 +189,41 @@ func TargetCluster(targetCluster string) MurModifier {
...
+type AccountOption func(ua *toolchainv1alpha1.UserAccountEmbedded)
 
+func SyncIndex(index string) AccountOption {
+	return func(ua *toolchainv1alpha1.UserAccountEmbedded) {
+		ua.SyncIndex = index
+	}
+}
+
+// Account sets the first account on the MasterUserRecord
+func Account(cluster string, tier toolchainv1alpha1.NSTemplateTier, options ...AccountOption) MurModifier {
```
MatousJobanek, line 201: May I know why you don't use the already existing type and functions?
There is already the `UaInMurModifier` type and related functions - is it possible to reuse them or transform the existing ones to your way? 
My point is that it looks like an overlap and duplication to me

xcoulon: it's a `MurModifier` because I want to be able to do something like this:
```
murtest.NewMasterUserRecords(t, 10, "user-%d", murtest.Account("cluster1", oldNSTemplateTier))
```
where I create 10 MURs with an account using the given tier.

pkg/test/masteruserrecord/master_user_record.go
xcoulon: `SyncIndex` value changed in 2486906

pkg/test/masteruserrecord/master_user_record.go
MatousJobanek: I was refering to `AccountOption`

xcoulon: > I was refering to `AccountOption`

Damn'! I always get confused with the block of code which is showed in the reviews 😬

xcoulon: > There is already the `UaInMurModifier` type and related functions - is it possible to reuse them or transform the existing ones to your way?
> My point is that it looks like an overlap and duplication to me

ok, let me see what I can do about it.

xcoulon: done in 8463a2b: using `UaInMurModifier` instead

go.mod
```diff
@@ -41,4 +41,6 @@ replace (
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20200204173128-addea2498afe // avoids case-insensitive import collision: "github.com/googleapis/gnostic/openapiv2" and "github.com/googleapis/gnostic/OpenAPIv2"
 )
 
+replace github.com/codeready-toolchain/api => github.com/xcoulon/api v0.0.0-20200630081021-4d177ed6d634
```
MatousJobanek, line 44: A reminder :-)

xcoulon: removed in https://github.com/codeready-toolchain/toolchain-common/commit/fdaf9a060f38f5785b1af3518f0c9a16e96757f5

PR #117, opened 2020-07-21
add func to count same occurrences of conditions …

xcoulon:
also, add funcs to set/update conditions without duplicate checks

### DIFF
--- a/pkg/condition/condition.go
+++ b/pkg/condition/condition.go
@@ -54,6 +54,19 @@ func IsFalseWithReason(conditions []toolchainv1alpha1.Condition, conditionType t
 	return found && c.Status == apiv1.ConditionFalse && c.Reason == reason
 }
 
+// Count counts the conditions that match the given type/status/reason
+func Count(conditions []toolchainv1alpha1.Condition, conditionType toolchainv1alpha1.ConditionType, status apiv1.ConditionStatus, reason string) int {
+	count := 0
+	for _, c := range conditions {
+		if c.Type == toolchainv1alpha1.TemplateUpdateRequestComplete &&
+			c.Status == status &&
+			c.Reason == reason {
+			count++
+		}
+	}
+	return count
+}
+
 func addOrUpdateStatusCondition(conditions []toolchainv1alpha1.Condition, newCondition toolchainv1alpha1.Condition) ([]toolchainv1alpha1.Condition, bool) {
 	newCondition.LastTransitionTime = metav1.Now()
 
--- a/pkg/condition/condition.go
+++ b/pkg/condition/condition.go
@@ -22,22 +22,13 @@ func AddOrUpdateStatusConditions(conditions []toolchainv1alpha1.Condition, newCo
 	return conditions, atLeastOneUpdated
 }
 
-// AddStatusConditions adds the given conditions *without* checking for duplicate types
-// Sets the `LastTransitionTime` to `metav1.Now()` for each given condition
+// AddStatusConditions adds the given conditions *without* checking for duplicate types (as opposed to `AddOrUpdateStatusConditions`)
+// Also, it sets the `LastTransitionTime` to `metav1.Now()` for each given condition if needed
 func AddStatusConditions(conditions []toolchainv1alpha1.Condition, newConditions ...toolchainv1alpha1.Condition) []toolchainv1alpha1.Condition {
 	for _, cond := range newConditions {
-		cond.LastTransitionTime = metav1.Now()
-		conditions = append(conditions, cond)
-	}
-	return conditions
-}
-
-// SetStatusConditions appends all the given conditions  without checking for duplicate types
-// Sets the `LastTransitionTime` to `metav1.Now()` for each given condition
-func SetStatusConditions(newConditions ...toolchainv1alpha1.Condition) []toolchainv1alpha1.Condition {
-	conditions := []toolchainv1alpha1.Condition{}
-	for _, cond := range newConditions {
-		cond.LastTransitionTime = metav1.Now()
+		if cond.LastTransitionTime.IsZero() {
+			cond.LastTransitionTime = metav1.Now()
+		}
 		conditions = append(conditions, cond)
 	}
 	return conditions
--- a/pkg/condition/condition.go
+++ b/pkg/condition/condition.go
@@ -70,7 +70,7 @@ func IsFalseWithReason(conditions []toolchainv1alpha1.Condition, conditionType t
 func Count(conditions []toolchainv1alpha1.Condition, conditionType toolchainv1alpha1.ConditionType, status apiv1.ConditionStatus, reason string) int {
 	count := 0
 	for _, c := range conditions {
-		if c.Type == toolchainv1alpha1.TemplateUpdateRequestComplete &&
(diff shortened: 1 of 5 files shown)


pkg/condition/condition.go
```diff
@@ -54,6 +75,19 @@ func IsFalseWithReason(conditions []toolchainv1alpha1.Condition, conditionType t
 	return found && c.Status == apiv1.ConditionFalse && c.Reason == reason
 }
 
+// Count counts the conditions that match the given type/status/reason
+func Count(conditions []toolchainv1alpha1.Condition, conditionType toolchainv1alpha1.ConditionType, status apiv1.ConditionStatus, reason string) int {
+	count := 0
+	for _, c := range conditions {
+		if c.Type == toolchainv1alpha1.TemplateUpdateRequestComplete &&
```
rajivnathan, line 82:
```suggestion
		if c.Type == conditionType &&
```
I'm assuming we want to check against `conditionType` here

pkg/condition/condition.go
```diff
@@ -22,6 +22,27 @@ func AddOrUpdateStatusConditions(conditions []toolchainv1alpha1.Condition, newCo
 	return conditions, atLeastOneUpdated
 }
 
+// AddStatusConditions adds the given conditions *without* checking for duplicate types
```
rajivnathan, line 25: Minor thing, but do we mean `fields` instead of `types` here? The `addOrUpdateStatusCondition` function checks the status, reason and message fields in addition to the type.

alexeykazakov: This function adds the new conditions even if the status already have some condition with the same type. The current `AddOrUpdateConditions()` do not append the condition if the status already has a condition with the same type. It will update the existing condition instead. But I guess Xavier needs to add multiple "Update" conditions to Tier resources.
I guess we need more detailed description of the function. We could also name it `AppendStatusConditions()` but I'm fine with just `AddStatusConditions()` too.

rajivnathan: I'm fine with `AddStatusConditions()` too. What I was trying to say is that this function does not care about any of the fields really, not just the type. eg. if you call this function and the type, status, reason and message are all the same it will still update the condition with at least the new LastTransitionTime. So maybe if we update the comment to something like `AddStatusConditions adds the given conditions *without* checking for any duplicate fields` it's more clear?

alexeykazakov: But only the type of the condition matters here, doesn't it? How you can possible check if there any duplicated fields if you just keep appending the given conditions? I mean what could be an alternative here? You can't ignore type duplication but still avoid having multiple condition with let's say the same `message` or 'lastTransaction`. It wouldn't make any sense, would it? So, mentioning it would be IMO redundant if not confusing.

The bottom line is that the type of the condition is served as an identification of unique conditions in most cases. But in some cases (Update condition in Tier) conditions don't have identifications. So we just keep appending new ones ignoring the existing type instances.

pkg/condition/condition.go
```diff
@@ -22,6 +22,27 @@ func AddOrUpdateStatusConditions(conditions []toolchainv1alpha1.Condition, newCo
...
+// Sets the `LastTransitionTime` to `metav1.Now()` for each given condition
+func AddStatusConditions(conditions []toolchainv1alpha1.Condition, newConditions ...toolchainv1alpha1.Condition) []toolchainv1alpha1.Condition {
+	for _, cond := range newConditions {
+		cond.LastTransitionTime = metav1.Now()
+		conditions = append(conditions, cond)
+	}
+	return conditions
+}
+
+// SetStatusConditions appends all the given conditions  without checking for duplicate types
```
alexeykazakov, line 35: Well.. this function doesn't append anything to anything. It's simple returns a copy of the same conditions but with LastTransitionTIme set to now.
Why would you need that? Couldn't you just set the last transition time on the client side?
Or at least could we maybe call this function differently? `UpdateLastTransitionTime()` or something like this?

pkg/condition/condition.go
rajivnathan: I thought it would not add multiple conditions that have the same type at the Kubernetes API level since type is the merge key but it seems that's not the case. So then yes, I agree with you. Sorry, it was only my confusion here. At least I learned something new :)

alexeykazakov: It's actually a good point. I didn't think about merging keys. So I guess it would be safer to change the list type to something else for those conditions even if it works now.

pkg/condition/condition.go
xcoulon: let me refactor the code, I did not notice how poorly written this func is 🤦‍♂️

pkg/condition/condition.go
xcoulon: @rajivnathan indeed you're making a very good point about the merge keys. I will see how to change that.

For the context: my intent here is to provide a func that does NOT work as its sibling `AddOrUpdateStatusConditions` (which makes sure that there is no duplicate condition based on the `Type` field).  The reason is that I want to record all failed attempts to update a MasterUserRecord, so we have multiple status conditions of the same type (with potentially the same reason and message, but it's not mandatory). And after a given number of attempts, the update process should stop for this particular MasterUserRecord.

xcoulon: (note that I did not expect that this change would trigger such a long discussion 😂 )

xcoulon:
> Can we add some tests for the new functions, please? :)

sure ;)

pkg/condition/condition.go
xcoulon: on second thought, this func has so little value that I'm going to drop it and set the `LastTransitionTime` on the caller side instead.

xcoulon: removed in 21363d1

xcoulon:
> > Can we add some tests for the new functions, please? :)
> 
> sure ;)

done in 21363d1

pkg/condition/condition.go
xcoulon: I've slightly clarified the comments in 21363d1. Please let me know if it's better now

xcoulon: > @rajivnathan indeed you're making a very good point about the merge keys. I will see how to change that.

actually, it's a VERY good point! I'm saying this because until now I did not see how to verify https://github.com/codeready-toolchain/host-operator/pull/243 in our e2e tests unless we have a reliable way to trigger failure (oh the irony!). 
Yet, we may have some "surprises" if the `merge key` removes the duplicate entries in the status conditions. So for now, I suggest I add a basic e2e test to verify that we can store multiple conditions with the same `type` in the TemplateUpdateRequest.

alexeykazakov: I think we should use `// +listType=atomic` and no patching strategy or merging keys.
See an example in MUR:
https://github.com/codeready-toolchain/api/blob/master/pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go#L97

pkg/condition/condition.go
xcoulon: good catch, thanks @rajivnathan!

pkg/condition/condition.go
xcoulon: done in https://github.com/codeready-toolchain/api/pull/170

PR #118, opened 2020-07-23
Use toolchainv1alpha1.Condition type for toolchain component status

rajivnathan:
Related issue: https://issues.redhat.com/browse/CRT-729

Use toolchainv1alpha1.Condition type for kubefed conditions so that all toolchain components can report their status using the same condition types for a uniform implementation.

Errors do not need to be returned with the conditions because they are already included in the conditions.

Related PRs:
https://github.com/codeready-toolchain/member-operator/pull/181
https://github.com/codeready-toolchain/host-operator/pull/254

### DIFF
--- a/pkg/status/componentconditions.go
+++ b/pkg/status/componentconditions.go
@@ -34,7 +34,7 @@ func NewComponentErrorCondition(reason, msg string) *toolchainv1alpha1.Condition
 }
 
 // ValidateComponentConditionReady checks whether the provided conditions signal that the component is ready, returns an error otherwise
-func ValidateComponentConditionReady(conditions []toolchainv1alpha1.Condition) error {
+func ValidateComponentConditionReady(conditions ...toolchainv1alpha1.Condition) error {
 	c, found := condition.FindConditionByType(conditions, toolchainv1alpha1.ConditionReady)
 	if !found {
 		return fmt.Errorf("a ready condition was not found")
--- a/pkg/status/componentconditions.go
+++ b/pkg/status/componentconditions.go
@@ -1,10 +1,13 @@
 package status
 
 import (
-	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"fmt"
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
+	"github.com/codeready-toolchain/toolchain-common/pkg/condition"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func NewComponentReadyCondition(reason string) *toolchainv1alpha1.Condition {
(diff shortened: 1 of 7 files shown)


pkg/status/componentconditions.go
```diff
@@ -29,3 +32,15 @@ func NewComponentErrorCondition(reason, msg string) *toolchainv1alpha1.Condition
 		LastUpdatedTime:    &currentTime,
 	}
 }
+
+// ValidateComponentConditionReady checks whether the provided conditions signal that the component is ready, returns an error otherwise
+func ValidateComponentConditionReady(conditions ...toolchainv1alpha1.Condition) error {
```
alexeykazakov, line 37: Please add tests for this function.

pkg/status/deployment_test.go
```diff
@@ -17,56 +16,65 @@ import (
...
+			}
+			test.AssertConditionsMatchAndRecentTimestamps(t, conditions, expected)
+		})
+
 		t.Run("deployment does not exist", func(t *testing.T) {
 			fakeClient := test.NewFakeClient(t)
-			conditions, err := GetDeploymentStatusConditions(fakeClient, "test-deployment", test.HostOperatorNs)
+			conditions := GetDeploymentStatusConditions(fakeClient, "test-deployment", test.HostOperatorNs)
+			err := ValidateComponentConditionReady(conditions...)
 			require.Error(t, err)
```
alexeykazakov, line 39: I think instead of indirectly testing ValidateComponentConditionReady in these tests we could just create dedicated tests for ValidateComponentConditionReady() in a separate componentconditions_test.go file with full coverage.

pkg/status/kubefed.go
```diff
@@ -30,31 +28,23 @@ type KubefedAttributes struct {
...
-		return &badClusterStatus, fmt.Errorf(notFoundMsg)
+		return []toolchainv1alpha1.Condition{*NewComponentErrorCondition(toolchainv1alpha1.ToolchainStatusClusterConnectionNotFoundReason, ErrMsgClusterConnectionNotFound)}
 	}
-	clusterStatus := *fedCluster.ClusterStatus.DeepCopy()
 
 	// check conditions of cluster connection
 	if !kubefed_util.IsClusterReady(fedCluster.ClusterStatus) {
-		return &clusterStatus, fmt.Errorf("the cluster connection is not ready")
+		for _, c := range fedCluster.ClusterStatus.Conditions {
+			if c.Type == "Ready" && c.Message != nil {
```
alexeykazakov, line 42: Shouldn't we check the status of the Ready Condition instead of Message? If it's True or not.
We still need to make sure that message is not nil before using it. But we should not relay on the message as the status of the condition.
Check the status. And if it's not true then use the message if it's not nil. If it's nil then use our own message.

pkg/status/componentconditions.go
rajivnathan: Added in dcc42a88fe4f39f0dc61d436066a786a3c4c4215

pkg/status/deployment_test.go
rajivnathan: I added the tests but I think it's still useful to test the error here as well so I'll keep these checks here if that's okay.

pkg/status/kubefed.go
rajivnathan: Actually the `kubefed_util.IsClusterReady` check above already checks the status of the ready condition so this block will only execute in the case when Ready is false. So here I'm being extra cautious and checking if the kubefed message exists so we can reuse it or if we want to provide a generic message if one does not exist.

rajivnathan:
@alexeykazakov thanks for the review! I already had the PRs for the other repos ready and was planning to put them up immediately after this PR but you make a good point that it could block other PRs which needs to put something new into common so I will make sure to put all PRs up together in the future.

pkg/status/kubefed.go
alexeykazakov: OK. But then if status is false but there is no message, shouldn't we provide our own message instead?

rajivnathan: Yup, we do. Just outside this for loop we have
```
genericErrMsg := "the cluster connection is not ready"
		return []toolchainv1alpha1.Condition{*NewComponentErrorCondition(toolchainv1alpha1.ToolchainStatusClusterConnectionNotReadyReason, genericErrMsg)}
```
So we provide the message `"the cluster connection is not ready"`

alexeykazakov: Right. I see it now.

PR #119, opened 2020-07-27
Load secrets and configmaps

tinakurian:
moves loading configMap and secret into common. This code will be used in both the host and member operator.

Jira: https://issues.redhat.com/browse/CRT-743
Relates to: <host-operator PR>

### DIFF
--- a/pkg/configuration/load_configuration.go
+++ b/pkg/configuration/load_configuration.go
@@ -85,7 +85,7 @@ func LoadFromConfigMap(prefix, resourceKey string, cl client.Client) error {
 
 	// get configMap data and set environment variables
 	for key, value := range configMap.Data {
-		configKey := CreateOperatorEnvVarKey(prefix, key)
+		configKey := createOperatorEnvVarKey(prefix, key)
 		err := os.Setenv(configKey, value)
 		if err != nil {
 			return err
@@ -107,11 +107,11 @@ func getResourceName(key string) string {
 	return resourceName
 }
 
-// CreateOperatorEnvVarKey creates env vars based on resource data.
+// createOperatorEnvVarKey creates env vars based on resource data.
 // Returns env var key.
 //
 // prefix: represents the operator prefix (HOST_OPERATOR/MEMBER_OPERATOR)
 // key: is the value to convert into an env var key
-func CreateOperatorEnvVarKey(prefix, key string) string {
+func createOperatorEnvVarKey(prefix, key string) string {
 	return prefix + "_" + (strings.ToUpper(strings.ReplaceAll(strings.ReplaceAll(key, ".", "_"), "-", "_")))
 }
--- a/pkg/configuration/load_configuration.go
+++ b/pkg/configuration/load_configuration.go
@@ -14,16 +14,15 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 )
 
-// LoadFromSecret retrieves an operator secret and sets a map
-// in order to override default configurations.
-// If no secret is found, then configuration will use defaults.
-// Returns error if WATCH_NAMESPACE is not set, if the resource GET request failed
-// (for other reasons apart from isNotFound) and if setting env vars fails.
+// LoadFromSecret retrieves an operator secret, loads all keys and values from the secret
+// and stores them in a map. This map is then returned by the function.
+// The function doesn't take into account any default values - this has to be
+// handled while getting the values in the configuration object.
 //
 // resourceKey: is the env var which contains the secret resource name.
 // cl: is the client that should be used to retrieve the secret.
-func LoadFromSecret(resourceKey string, cl client.Client) (map[string][]byte, error) {
-	var secretData = make(map[string][]byte)
+func LoadFromSecret(resourceKey string, cl client.Client) (map[string]string, error) {
+	var secretData = make(map[string]string)
 
 	// get the secret name
 	secretName := getResourceName(resourceKey)
@@ -47,7 +46,11 @@ func LoadFromSecret(resourceKey string, cl client.Client) (map[string][]byte, er
 		logf.Log.Info("secret is not found")
 	}
 
-	return secret.Data, nil
+	for key, value := range secret.Data {
+		secretData[key] = string(value)
+	}
+
+	return secretData, nil
 }
 
 // LoadFromConfigMap retrieves the host operator configmap and sets environment
--- a/pkg/configuration/load_configuration.go
+++ b/pkg/configuration/load_configuration.go
@@ -14,26 +14,28 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 )
 
-// LoadFromSecret retrieves an operator secret and sets environment
-// variables in order to override default configurations.
+// LoadFromSecret retrieves an operator secret and sets a map
+// in order to override default configurations.
 // If no secret is found, then configuration will use defaults.
 // Returns error if WATCH_NAMESPACE is not set, if the resource GET request failed
 // (for other reasons apart from isNotFound) and if setting env vars fails.
 //
 // prefix: represents the operator prefix (HOST_OPERATOR/MEMBER_OPERATOR)
 // resourceKey: is the env var which contains the secret resource name.
-// cl: is the client that should be used to retrieve the configmap.
-func LoadFromSecret(prefix, resourceKey string, cl client.Client) error {
+// cl: is the client that should be used to retrieve the secret.
+func LoadFromSecret(prefix, resourceKey string, cl client.Client) (map[string]string, error) {
+	var secretData = make(map[string]string)
+
 	// get the secret name
 	secretName := getResourceName(resourceKey)
 	if secretName == "" {
-		return nil
+		return secretData, nil
 	}
 
 	// get the secret
 	namespace, err := k8sutil.GetWatchNamespace()
 	if err != nil {
-		return err
+		return secretData, err
 	}
 
 	secret := &v1.Secret{}
@@ -41,7 +43,7 @@ func LoadFromSecret(prefix, resourceKey string, cl client.Client) error {
 	err = cl.Get(context.TODO(), namespacedName, secret)
 	if err != nil {
 		if !errs.IsNotFound(err) {
-			return err
+			return secretData, err
 		}
 		logf.Log.Info("secret is not found")
 	}
@@ -49,13 +51,10 @@ func LoadFromSecret(prefix, resourceKey string, cl client.Client) error {
 	// get secrets and set environment variables
 	for key, value := range secret.Data {
 		secretKey := createOperatorEnvVarKey(prefix, key)
-		err := os.Setenv(secretKey, string(value))
-		if err != nil {
-			return err
-		}
+		secretData[secretKey] = string(value)
 	}
 
-	return nil
+	return secretData, nil
 }
 
 // LoadFromConfigMap retrieves the host operator configmap and sets environment
--- a/pkg/configuration/load_configuration.go
+++ b/pkg/configuration/load_configuration.go
@@ -23,8 +23,8 @@ import (
 // prefix: represents the operator prefix (HOST_OPERATOR/MEMBER_OPERATOR)
 // resourceKey: is the env var which contains the secret resource name.
 // cl: is the client that should be used to retrieve the secret.
-func LoadFromSecret(prefix, resourceKey string, cl client.Client) (map[string]string, error) {
-	var secretData = make(map[string]string)
+func LoadFromSecret(resourceKey string, cl client.Client) (map[string][]byte, error) {
+	var secretData = make(map[string][]byte)
 
 	// get the secret name
 	secretName := getResourceName(resourceKey)
@@ -48,13 +48,7 @@ func LoadFromSecret(prefix, resourceKey string, cl client.Client) (map[string]st
 		logf.Log.Info("secret is not found")
 	}
 
-	// get secrets and set environment variables
-	for key, value := range secret.Data {
-		secretKey := CreateOperatorEnvVarKey(prefix, key)
-		secretData[secretKey] = string(value)
-	}
-
-	return secretData, nil
+	return secret.Data, nil
 }
 
 // LoadFromConfigMap retrieves the host operator configmap and sets environment
--- a/pkg/configuration/load_configuration.go
+++ b/pkg/configuration/load_configuration.go
@@ -1,4 +1,4 @@
-package controller
+package configuration
 
 import (
 	"context"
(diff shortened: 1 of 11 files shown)


pkg/controller/load_configuration.go
```diff
@@ -0,0 +1,96 @@
...
+	"strings"
+
+	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+// LoadFromSecret retrieves an operator secret
```
alexeykazakov, line 15: Can we add more details to the function description? What exactly this function does. What are the given params, etc. In what case an error is returned and what cases are ignored (no secret name is specified or exists, etc).
This function now lives in `common` so it now requires a better documentation :)

pkg/controller/load_configuration.go
```diff
@@ -0,0 +1,96 @@
...
+	namespace, err := k8sutil.GetWatchNamespace()
+	if err != nil {
+		return err
+	}
+
+	secret := &v1.Secret{}
+	namespacedName := types.NamespacedName{Namespace: namespace, Name: secretName}
+	err = cl.Get(context.TODO(), namespacedName, secret)
+	if err != nil {
+		return err
```
alexeykazakov, line 33: We should ignore NotFound errors. But we want to log a message saying that there is no Secret so we don't load anything from it.

pkg/controller/load_configuration.go
```diff
@@ -0,0 +1,96 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// LoadFromConfigMap retrieves the host operator configMap
+func LoadFromConfigMap(prefix, resourceKey string, cl client.Client) error {
```
alexeykazakov, line 49: Same here: better documentation.

pkg/controller/load_configuration.go
```diff
@@ -0,0 +1,96 @@
...
+	// get the configMap
+	namespace, err := k8sutil.GetWatchNamespace()
+	if err != nil {
+		return err
+	}
+
+	configMap := &v1.ConfigMap{}
+	namespacedName := types.NamespacedName{Namespace: namespace, Name: configMapName}
+	err = cl.Get(context.TODO(), namespacedName, configMap)
+	if err != nil {
```
alexeykazakov, line 65: Ignore NotFoundError.

pkg/controller/load_configuration.go
```diff
@@ -0,0 +1,96 @@
...
+
+	return nil
+}
+
+// getResourceName gets the resource name via env var
+func getResourceName(key string) string {
+	// get the resource name
+	resourceName := os.Getenv(key)
+	if resourceName == "" {
+		logf.Log.Info(key + " is not set")
```
alexeykazakov, line 86: Let's provide more information here. What key and what we are doing here.

pkg/controller/load_configuration.go
tinakurian: this is done in each operator, but i guess i can move it here, i just thought it would be better to have it in the operator so we have more control over the error at that point

pkg/controller/load_configuration.go
tinakurian: done here: 26869a4875e8152516ff11b763471d23defe1670

pkg/controller/load_configuration.go
tinakurian: done here: 26869a4875e8152516ff11b763471d23defe1670

pkg/controller/load_configuration.go
tinakurian: done here: 26869a4875e8152516ff11b763471d23defe1670

pkg/controller/load_configuration.go
tinakurian: done here: 26869a4875e8152516ff11b763471d23defe1670

pkg/controller/load_configuration.go
tinakurian: done here: 26869a4875e8152516ff11b763471d23defe1670

tinakurian:
> Looks good!
> But can we also add tests for:
> 
> 1. No WATCH_NAMESPACE set
> 2. No config/secret name is set
> 3. There is some not error which is not NotFoundError.

done here: 029f1c85e69d027b16ce672cd0b1d69208bc8e2f and here: 55b5f8cbddc6cc4e96dcb524e7044ea66c3c1a48

pkg/controller/load_configuration_test.go
```diff
@@ -42,25 +117,80 @@ func TestLoadFromConfigMap(t *testing.T) {
...
+		// test that the secret was not found since no secret name was set
+		testTest := os.Getenv("HOST_OPERATOR_SPECIAL_KEY")
+		assert.Equal(t, "", testTest)
+	})
+	t.Run("env overwrite", func(t *testing.T) {
+		// given
+		restore := test.SetEnvVarAndRestore(t, "HOST_OPERATOR_SECRET_NAME", "test-secret")
+		defer restore()
+
+		secret := &v1.Secret{
```
alexeykazakov, line 167: I guess you could make code a bit more readable if you create a couple of functions like `createSecret(key, value string)/createCM()` and reuse it everywhere. But it's minor.

tinakurian: done here: 037f319d194e6770d99e7ccf422bc49047cf5e3e

pkg/controller/load_configuration.go
```diff
@@ -0,0 +1,120 @@
...
+		if !errs.IsNotFound(err) {
+			return err
+		}
+		logf.Log.Info("secret is not found")
+	}
+
+	// get secrets and set environment variables
+	for key, value := range secret.Data {
+		secretKey := createOperatorEnvVarKey(prefix, key)
+		err := os.Setenv(secretKey, string(value))
```
MatousJobanek, line 52: @tinakurian 
We need to be sure that none of the operators prints the secret variables to the log.
host-operator seems to be OKish: https://github.com/codeready-toolchain/host-operator/blob/master/cmd/manager/main.go#L277-L289
but member-operator is not safe: https://github.com/codeready-toolchain/member-operator/blob/master/cmd/manager/main.go#L254-L260 this has to be fixed, otherwise, the secrets will be printed to the log.

Btw, do we really need to set them as env variables? Why don't you store them as variables of the configuration object? This shouldn't be so hard (even if you want to make it generic) and it will be much safer. I have some weird feeling when we set the secrets as normal variables.

How I would do it? There are  a few options, but the easiest way is to store them in a map:
```go
type Config struct {
	host  *viper.Viper
	secretValues map[string]string
}
```
by default, the map would already contain the keys and the default values.
then then `LoadFromSecret` would take the map and assign the values there (if not defined then it keeps the default value). 
The respective methods (such as `GetMailgunDomain()`) would then read the value from the map

pkg/controller/load_configuration_test.go
```diff
@@ -0,0 +1,238 @@
...
+		require.Error(t, err)
+		assert.Equal(t, "oopsie woopsie", err.Error())
+
+		// test env vars are parsed and created correctly
+		testTest := os.Getenv("MEMBER_OPERATOR_TEST_KEY_ONE")
+		assert.Equal(t, testTest, "")
+	})
+	t.Run("env overwrite", func(t *testing.T) {
+		// given
+		restore := test.SetEnvVarAndRestore(t, "MEMBER_OPERATOR_CONFIG_MAP_NAME", "test-config")
```
MatousJobanek, line 80: it's just a detail, but we should also set and restore the `MEMBER_OPERATOR_TEST_KEY` var
```suggestion
		restore := test.SetEnvVarsAndRestore(t,
			test.Env("MEMBER_OPERATOR_CONFIG_MAP_NAME", "test-config"),
			test.Env("MEMBER_OPERATOR_TEST_KEY", ""))
```

pkg/controller/load_configuration_test.go
```diff
@@ -0,0 +1,238 @@
...
+		require.Error(t, err)
+		assert.Equal(t, "oopsie woopsie", err.Error())
+
+		// test env vars are parsed and created correctly
+		testTest := os.Getenv("HOST_OPERATOR_TEST_KEY_SECRET")
+		assert.Equal(t, testTest, "")
+	})
+	t.Run("env overwrite", func(t *testing.T) {
+		// given
+		restore := test.SetEnvVarAndRestore(t, "HOST_OPERATOR_SECRET_NAME", "test-secret")
```
MatousJobanek, line 160: same here:
```suggestion
		restore := test.SetEnvVarsAndRestore(t,
			test.Env("HOST_OPERATOR_SECRET_NAME", "test-secret"),
			test.Env("HOST_OPERATOR_TEST_KEY", ""))
```

pkg/controller/load_configuration.go
alexeykazakov: Good points @MatousJobanek .

tinakurian: @MatousJobanek i don't think member-operator has secrets at the moment? just a configmap

tinakurian:
> Why did you add the files into the `controller` package? this doesn't make any sense to me - it doesn't have anything in common with controller - this is a config of the whole operator.
> Could you please move it into a `configuration` package?

yeah, i had actually already spoken to @alexeykazakov this morning about this :-) just haven't gotten around to updating this PR, will do this soon..

pkg/controller/load_configuration_test.go
tinakurian: fixed here: 8cba0fc9d72606ec22c4ef1d91cab9288ab19460

pkg/controller/load_configuration_test.go
tinakurian: fixed here: 9bd69e9343e58d1f672002e024e566533a3aecdf

pkg/controller/load_configuration.go
tinakurian: @MatousJobanek i have changed this to use a map as you suggested, please check this PR: https://github.com/codeready-toolchain/host-operator/pull/255. I've also updated this PR to reflect those changes: deab6ebe13dfe4de125b50d196c3b414be22630a

MatousJobanek: Awesome, thanks a lot for this :+1: 
So, from the code, it looks like you delegate the "default values management" to the `Get*` methods, right?

pkg/configuration/load_configuration.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+// LoadFromSecret retrieves an operator secret and sets a map
+// in order to override default configurations.
+// If no secret is found, then configuration will use defaults.
```
MatousJobanek, line 19: Could you please fix the comment? 
this function doesn't set a map and also it doesn't care about default values.
you could clarify here how the default values should be used, what is in the map etc...

pkg/configuration/load_configuration.go
```diff
@@ -14,48 +14,47 @@ import (
...
 	}
 
 	// get secrets and set environment variables
 	for key, value := range secret.Data {
 		secretKey := createOperatorEnvVarKey(prefix, key)
-		err := os.Setenv(secretKey, string(value))
-		if err != nil {
-			return err
-		}
+		secretData[secretKey] = string(value)
```
MatousJobanek: btw, do we still need to use the env var format? I mean to have it capitalized, with the prefix and with underscores...?
We don't use it as an env var anymore. 
If we keep it in the original format, then we don't have to do the same transformation here  https://github.com/codeready-toolchain/host-operator/pull/255/files#diff-ec24f891497e9e435f8f9956e23cb379R172 correct?

pkg/configuration/load_configuration.go
tinakurian: @MatousJobanek hmm i'm a bit confused, it does set a map based off of what is found in the secret. how would you write this comment in order to be more clear?

pkg/controller/load_configuration.go
tinakurian: @MatousJobanek no worries! currently these values don't have any defaults. they either exist via the secret or they don't exist at all. but in the future, if we needed a secret value to have a default it would need to be handled in the configuration `Get*` funcs

pkg/configuration/load_configuration.go
tinakurian: @MatousJobanek good point and good catch! we shouldn't do the transformation unnecessarily, i'll fix this

pkg/configuration/load_configuration.go
MatousJobanek: I was thinking something along these lines:
```suggestion
// LoadFromSecret retrieves an operator secret, loads all keys and values from the secret
// and stores them in a map. This map is then returned by the function.
// The function doesn't take into account any default values - this has to be 
// handled while getting the values in the configuration object. 
```

pkg/configuration/load_configuration.go
MatousJobanek: well, I was talking only about the  `secretKey := createOperatorEnvVarKey(prefix, key)` part, but your version works too

pkg/configuration/load_configuration.go
```diff
@@ -0,0 +1,117 @@
...
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+// LoadFromSecret retrieves an operator secret and sets a map
+// in order to override default configurations.
+// If no secret is found, then configuration will use defaults.
+// Returns error if WATCH_NAMESPACE is not set, if the resource GET request failed
+// (for other reasons apart from isNotFound) and if setting env vars fails.
+//
+// prefix: represents the operator prefix (HOST_OPERATOR/MEMBER_OPERATOR)
```
MatousJobanek, line 23: the prefix is not used
```suggestion
```

tinakurian: fixed here: bf4fb50b7964b03d59e810a8275ee5af586da8ac

pkg/configuration/load_configuration.go
MatousJobanek: @tinakurian ^

tinakurian: sweet! done as part of: d24d418681e1bcca5e857f042cb5d6f07fa08162

pkg/configuration/load_configuration.go
tinakurian: @MatousJobanek hm, actually this is less efficient. i've changed it here so we only need to convert the value to a string once rather than every time we get the config value: d24d418681e1bcca5e857f042cb5d6f07fa08162

PR #121, opened 2020-08-03
migrate add-cluster to ToolchainCluster

MatousJobanek:
migrates add-cluster.sh script to use ToolchainCluster


MatousJobanek:
>We will also need to update toolchain-infra scripts. Just comments and readme. I don't think we need to update any logic in toolchain-infra.

here is the PR: https://github.com/codeready-toolchain/toolchain-infra/pull/30

> We will also need to migrate our crt-stage clusters to ToolchainClusters.

sure - I'll take care of it before merging the PRs

PR #122, opened 2020-08-07
Add HasExpectedProvisionedTime assertion function

rajivnathan:
This PR adds an assertion for the unit tests related to the MUR.Status.ProvisionedTime

Related PRs:
https://github.com/codeready-toolchain/api/pull/176
https://github.com/codeready-toolchain/host-operator/pull/261

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -40,4 +40,6 @@ replace (
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20200204173128-addea2498afe // avoids case-insensitive import collision: "github.com/googleapis/gnostic/openapiv2" and "github.com/googleapis/gnostic/OpenAPIv2"
 )
 
+replace github.com/codeready-toolchain/api => github.com/rajivnathan/api v0.0.0-20200805154153-997c3ad9c93c
+
 go 1.13
--- a/pkg/test/masteruserrecord/assertion.go
+++ b/pkg/test/masteruserrecord/assertion.go
@@ -3,12 +3,15 @@ package masteruserrecord
 import (
 	"context"
 	"fmt"
+	"time"
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
 	"github.com/codeready-toolchain/toolchain-common/pkg/test"
 	"github.com/davecgh/go-spew/spew"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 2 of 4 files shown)


pkg/test/masteruserrecord/assertion.go
```diff
@@ -102,6 +105,20 @@ func (a *Assertion) HasStatusUserAccounts(targetClusters ...string) *Assertion {
 	return a
 }
 
+func (a *Assertion) HasExpectedProvisionedTime(expectedProvisionedTime metav1.Time) *Assertion {
+	err := a.loadUaAssertion()
+	require.NoError(a.t, err)
+
+	if expectedProvisionedTime == (metav1.Time{}) {
```
MatousJobanek, line 112: wouldn't be better to:
* either create another assertion method called `HasNoProvisionedTime()` instead of checking emptiness
* or accepting a pointer as a parameter `HasExpectedProvisionedTime(expectedProvisionedTime *metav1.Time)` when the MUR.status keeps the pointer instead of the value

go.mod
```diff
@@ -40,4 +40,6 @@ replace (
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20200204173128-addea2498afe // avoids case-insensitive import collision: "github.com/googleapis/gnostic/openapiv2" and "github.com/googleapis/gnostic/OpenAPIv2"
 )
 
+replace github.com/codeready-toolchain/api => github.com/rajivnathan/api v0.0.0-20200805154153-997c3ad9c93c
```
MatousJobanek, line 43: Reminder :-)

pkg/test/masteruserrecord/assertion.go
```diff
@@ -102,6 +105,20 @@ func (a *Assertion) HasStatusUserAccounts(targetClusters ...string) *Assertion {
...
+		require.Nil(a.t, a.masterUserRecord.Status.ProvisionedTime)
+	} else {
+		// when the expected ProvisionedTime is not nil, check that the time was recently updated
+		require.WithinDuration(a.t, expectedProvisionedTime.Time, a.masterUserRecord.Status.ProvisionedTime.Time, 1*time.Second)
```
MatousJobanek, line 116: btw, instead of this, you could expect a time after which the user was provisioned. 
In other words, in the `// given` part you would store the time `metav1.Now()`, then trigger the reconcile, and here you would verify that the actual provisioned time was set after the given time. With such a flow, you could easily debug the code, stop the execution for some time and the tests will still pass.

pkg/test/masteruserrecord/assertion.go
rajivnathan: Good ideas! I went with creating two new functions, `verifyMurStatusExpectedProvisionedTime` and `verifyMurStatusNoProvisionedTime`

pkg/test/masteruserrecord/assertion.go
rajivnathan: That's a very good point about debugging the code and stopping the execution. It would be very confusing if the test fails if you wait for some time. Thanks for the suggestion!

alexeykazakov: I agree with both Matous' comments. I think the name of the function is a bit misleading.
Can we rename it to `ProvisionedAfter(expectedProvisionedTime metav1.Time)` which will check that the provisioned time is set to something between the `expectedTIme` and `time.Now` (just to make sure it's not set to something in the future).

pkg/test/masteruserrecord/assertion.go
alexeykazakov: Hm.. I don't see any updates in the PR but also would prefer a separate function here:
`HasNoProvisionedTime()` in addition to `ProvisionedAfter(expectedProvisionedTime metav1.Time)`

rajivnathan:
@MatousJobanek @alexeykazakov Thank you both for the reviews! Because of Matous' suggestion in https://github.com/codeready-toolchain/host-operator/pull/261 I think this PR is no longer needed so I will close it for now. I can reopen it if we believe we still need changes here

PR #133, opened 2020-10-09
fix: create new SA when joining a cluster

MatousJobanek:
The OLM creates new SA, (Cluster)Role and (Cluster)RoleBinding for every release/deployment. This means that the SA tokens we copied when joining host/member clusters is not valid anymore. Thus, I had to adjust the script a bit, so we don't rely on the SA created by OLM nor the names.
Changes in the `add-cluster.sh` script:
* dropped Minishift support to make the code cleaner
* add a new param `--type-name` to specify the name of the cluster type - this will be used in e2e tests by passing different name than `host` or `member`
* when the script is used for normal joinining (not during e2e tests) then it creates in each of the clusters a new ServiceAccount  (`toolchaincluster-${JOINING_CLUSTER_TYPE}-operator`) with a Role alowing modification of Toolchain resources and ClusterRole allowing retreival of routes). The token of this new SA is then copied to the other cluster.
* when the script is used during e2e tests, then it creates a new SA called `e2e-service-account`, retrieves the names of the Role & ClusterRole created by OLM and uses the names in role bindings. Then, analogically, it copies the token and creates a new `ToolchainCluster` resource. This resource will be then used in e2e tests.

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/176


alexeykazakov (approved):
Nice! Even a smaller change than I expected.

xcoulon:
sorry, I'm late on this, but what problem is it going to solve?

MatousJobanek:
> sorry, I'm late on this, but what problem is it going to solve? 

sorry for not tu writing the whole context of this effort. The original code was wrong and didn't solve anything. :grimacing: 
I adjusted the script again and also rewrote the description of the PR - please, take a look at it ;-)

scripts/add-cluster.sh
```diff
@@ -13,25 +14,126 @@ user_help () {
...
+metadata:
+  name: ${SA_NAME}
+rules:
+- apiGroups:
+  - route.openshift.io
+  resources:
+  - routes
+  verbs:
+  - "get"
+  - "list"
```
MatousJobanek, line 65: @alexeykazakov yes, I know that this won't probably work in OSD clusters, but I have a solution for that. 
We need this permission in host operator for getting the console & Che URL of the member cluster:  https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/masteruserrecord/sync.go#L144-L193
My plan is to drop this code from the host operator, move it to member to `MemberStatus` controller and store the URLs in the `MemberStatus` CR.
I believe that this will be also a good isolation - the `ToolchainCluster` tokens will be able to modify only the `toolchain.*` resources and the information about the other resources should be passed as part of the `MemberStatus` CR.

alexeykazakov: Yeah... I don't like the current way of obtaining/setting the console and Che URLs anyway... it currently looks more like a hack.
I'm OK to move it to the member status. It sounds reasonable. And I think we should then add it to the toolchain status too, so our reg service can obtain it from there instead of relying on individual URLs per user. Those URLs are global anyway. And we don't want to deal with updating all the users if those URLs are changed.
But for now let's do the minimum. Just enough to unblock the toolchaincluster thing.

PR #138, opened 2020-11-06
Support sandbox config

MatousJobanek:
It adds support for sandbox config and download of "let's encrypt" certificate when requested

### DIFF
--- a/scripts/add-cluster.sh
+++ b/scripts/add-cluster.sh
@@ -51,7 +51,22 @@ rules:
 - apiGroups:
   - toolchain.dev.openshift.com
   resources:
-  - "*"
+  - "bannedusers"
+  - "changetierrequests"
+  - "hostoperatorconfigs"
+  - "idlers"
+  - "masteruserrecords"
+  - "memberstatuses"
+  - "notifications"
+  - "nstemplatesets"
+  - "nstemplatetiers"
+  - "registrationservices"
+  - "templateupdaterequests"
+  - "tiertemplates"
+  - "toolchainclusters"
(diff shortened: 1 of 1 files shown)


scripts/add-cluster.sh
```diff
@@ -43,29 +51,24 @@ rules:
 - apiGroups:
   - toolchain.dev.openshift.com
   resources:
-  - "*"
```
alexeykazakov, line 46: Why not using *?
We will have to update this file every time we add a new CRD if we don't use *. Which I'm sure we will forget to do.

MatousJobanek: Because dedicated-admin doesn't have the `*` permission, so it cannot grant permission that doesn't hold. 

I know that it's not really flexible, but I'll try to change it in the future. I think that it would be good to rewrite it to go command where we can set the resources dynamically

alexeykazakov (approved):

scripts/add-cluster.sh
alexeykazakov: Ok. Not super critical right now but would be great to load them dynamical. For now let's add a comment to the make file in the api repo? The place where we list the kinds for make generate.

alexeykazakov: This is the make file we can add a comment to for now: https://github.com/codeready-toolchain/api/blob/master/make/generate.mk#L6-L8

MatousJobanek: https://github.com/codeready-toolchain/api/pull/203

PR #145, opened 2020-12-16
fix: retain Service's 'spec.ClusterIP' field during update

xcoulon:
This field is part of the spec but it is immutable, so we
have to retain the existing value when performing an update.
Otherwise, we get this kind of error:
```
Service "<name>" is invalid: spec.clusterIP: Invalid value: "": field is immutable
```

Since the object arg can be explicitely a `corev1.Service` or
an `unstructured.Untructured` (if coming from a template), we need
to support both cases.

Also: reorganized the tests (and added a few more!)

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -10,9 +10,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
@@ -91,9 +93,22 @@ func (p ApplyClient) createOrUpdateObj(newResource runtime.Object, forceUpdate b
 
 	// retrieve the current 'resourceVersion' to set it in the resource passed to the `client.Update()`
 	// otherwise we would get an error with the following message:
-	// "nstemplatetiers.toolchain.dev.openshift.com \"basic\" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update"
+	// `nstemplatetiers.toolchain.dev.openshift.com "basic" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update`
 	originalGeneration := metaExisting.GetGeneration()
 	metaNew.SetResourceVersion(metaExisting.GetResourceVersion())
+
+	// also, if the resource to create is a Service and there's a previous version, we should retain its `spec.ClusterIP`, otherwise
+	// the update will fail with the following error:
+	// `Service "<name>" is invalid: spec.clusterIP: Invalid value: "": field is immutable`
+	switch newResource := newResource.(type) {
+	case *corev1.Service:
+		newResource.Spec.ClusterIP = existing.(*corev1.Service).Spec.ClusterIP
+	case *unstructured.Unstructured:
+		if clusterIP, found, err := unstructured.NestedString(existing.(*unstructured.Unstructured).Object, "spec", "clusterIP"); err == nil && found {
+			unstructured.SetNestedField(newResource.Object, clusterIP, "spec", "clusterIP")
+		}
+	}
+
 	if err := p.cl.Update(context.TODO(), newResource); err != nil {
 		return false, errors.Wrapf(err, "unable to update the resource '%v'", newResource)
 	}
@@ -109,20 +124,19 @@ func (p ApplyClient) createOrUpdateObj(newResource runtime.Object, forceUpdate b
 }
 
 func getNewConfiguration(newResource runtime.Object) string {
-	newJson, err := marshalObjectContent(newResource)
+	newJSON, err := marshalObjectContent(newResource)
 	if err != nil {
 		log.Error(err, "unable to marshal the object", "object", newResource)
 		return fmt.Sprintf("%v", newResource)
 	}
-	return string(newJson)
+	return string(newJSON)
 }
 
 func marshalObjectContent(newResource runtime.Object) ([]byte, error) {
 	if newRes, ok := newResource.(runtime.Unstructured); ok {
 		return json.Marshal(newRes.UnstructuredContent())
-	} else {
-		return json.Marshal(newResource)
 	}
+	return json.Marshal(newResource)
 }
 
 func (p ApplyClient) createObj(newResource runtime.Object, metaNew v1.Object, owner v1.Object) error {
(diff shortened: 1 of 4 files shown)


tinakurian (approved):

alexeykazakov (approved):

pkg/client/client_test.go
```diff
@@ -59,133 +66,248 @@ func TestApplySingle(t *testing.T) {
...
+					cl, _ := newClient(t, s)
+					obj := defaultService.DeepCopy()
+					_, err := cl.CreateOrUpdateObject(obj, true, nil)
+					require.NoError(t, err)
+					originalGeneration := obj.GetGeneration()
+
+					// when updating with the modified obj
+					modifiedObj := modifiedService.DeepCopy()
+					modifiedObj.Spec.ClusterIP = ""
+					modifiedObj.ObjectMeta.Generation = obj.GetGeneration()
```
MatousJobanek, line 126: why do you copy the Generation? this shouldn't be needed
```suggestion
```

pkg/client/client_test.go
```diff
@@ -59,133 +66,248 @@ func TestApplySingle(t *testing.T) {
...
+					createdOrChanged, err := cl.CreateOrUpdateObject(modifiedObj, true, nil)
+
+					// then
+					require.NoError(t, err)
+					assert.True(t, createdOrChanged) // resource was updated on the server, so returned value if `true`
+					updateGeneration := modifiedObj.GetGeneration()
+					assert.Equal(t, originalGeneration+1, updateGeneration)
+				})
+
+				t.Run("it should update when specs are different including ClusterIP", func(t *testing.T) {
```
MatousJobanek, line 136: I maybe don't understand the message - the ClusterIP is not changed

pkg/client/client_test.go
```diff
@@ -59,133 +66,248 @@ func TestApplySingle(t *testing.T) {
...
+					// given
+					cl, _ := newClient(t, s)
+					obj := defaultService.DeepCopy()
+					_, err := cl.CreateOrUpdateObject(obj, true, nil)
+					require.NoError(t, err)
+					originalGeneration := obj.GetGeneration()
+
+					// when updating with the modified obj
+					modifiedObj := modifiedService.DeepCopy()
+					modifiedObj.ObjectMeta.Generation = obj.GetGeneration()
```
MatousJobanek, line 146: the same here - we don't need to copy that:
```suggestion
```

pkg/client/client_test.go
```diff
@@ -59,133 +66,248 @@ func TestApplySingle(t *testing.T) {
...
+			// only testing the specific cases of Services where an existing version exists, with a `spec.clusterIP` set
+			// and the updated version has no value for this field
+
+			t.Run("when using forceUpdate=true", func(t *testing.T) {
+
+				t.Run("it should not update when specs are same except ClusterIP", func(t *testing.T) {
+					// given
+					cl, _ := newClient(t, s)
+					// convert to unstructured
+					obj, err := toUnstructured(defaultService.DeepCopy())
```
MatousJobanek, line 255: you could use 
```
runtime.DefaultUnstructuredConverter.ToUnstructured(defaultService.DeepCopy())
```

pkg/client/client_test.go
```diff
@@ -59,133 +66,248 @@ func TestApplySingle(t *testing.T) {
...
+					require.NoError(t, err)
+					_, err = cl.CreateOrUpdateObject(obj, true, nil)
+					require.NoError(t, err)
+					originalGeneration := obj.GetGeneration()
+					err = unstructured.SetNestedField(obj.Object, "", "spec", "clusterIP") // modify for version to update
```
MatousJobanek, line 260: I would rather use a new object instead of reusing the existing `obj` - this will also simulate the real usage

MatousJobanek (commented):
Nice improvement :+1: :1st_place_medal: 
I have a few suggestions for the tests

pkg/client/client_test.go
xcoulon: Well, that was part of the existing test ;) There was simply a change on these lines because I moved the test into a subtest. 
But you're right, this line is not needed and I'll remove it ;)

pkg/client/client_test.go
xcoulon: doh! I forgot to add something like `modifiedObj.Spec.ClusterIP = ""` 🤦‍♂️

pkg/client/client_test.go
xcoulon: sadly, `runtime.DefaultUnstructuredConverter.ToUnstructured(defaultService.DeepCopy())` returns `map[string]interface{}` and I need an `*unstructured.Unstructured` (which implements `runtime.Object`)

pkg/client/client_test.go
xcoulon: done in 83e733b

pkg/client/client_test.go
xcoulon: done in 83e733b

pkg/client/client_test.go
xcoulon: done in 83e733b

pkg/client/client_test.go
xcoulon: done in 83e733b

xcoulon:
> Nice improvement 👍 🥇
> I have a few suggestions for the tests

thanks for the review @MatousJobanek ! I applied the requested changes in 83e733b. Please let me know if it's good for you now ;)

pkg/client/client_test.go
MatousJobanek: ah, right... you would need to do:
```
converted, err := runtime.DefaultUnstructuredConverter.ToUnstructured(defaultService.DeepCopy())
require.NoError(t, err)
obj := &unstructured.Unstructured{
	Object: converted,
}
```

pkg/client/client.go
```diff
@@ -91,9 +93,22 @@ func (p ApplyClient) createOrUpdateObj(newResource runtime.Object, forceUpdate b
...
+
+	// also, if the resource to create is a Service and there's a previous version, we should retain its `spec.ClusterIP`, otherwise
+	// the update will fail with the following error:
+	// `Service "<name>" is invalid: spec.clusterIP: Invalid value: "": field is immutable`
+	switch newResource := newResource.(type) {
+	case *corev1.Service:
+		newResource.Spec.ClusterIP = existing.(*corev1.Service).Spec.ClusterIP
+	case *unstructured.Unstructured:
+		if clusterIP, found, err := unstructured.NestedString(existing.(*unstructured.Unstructured).Object, "spec", "clusterIP"); err == nil && found {
+			unstructured.SetNestedField(newResource.Object, clusterIP, "spec", "clusterIP")
```
MatousJobanek, line 108: btw, this function returns an error:
```
Returns an error if value cannot be set because one of the nesting levels is not a map[string]interface{}.
```

pkg/client/client_test.go
xcoulon: are you asking me to apply this change in the `toUnstructured` func?

pkg/client/client.go
xcoulon: yes, but I checked the `err` and here we only have 1 level: `spec` which should not be anything else than `map[string]interface{}`. Or am I missing something?

MatousJobanek: yeah, you are right - I didn't realize that. 
Anyway, wouldn't it be better to add either comment there explaining it (plus say linter to ignore it) or just handle it since it's very simple change?
```suggestion
			if err := unstructured.SetNestedField(newResource.Object, clusterIP, "spec", "clusterIP"); err != nil {
				return false, err
			}
```

xcoulon:
@MatousJobanek I believe I addressed all your comments. Could you give this PR another look?

pkg/client/client_test.go
MatousJobanek: no, I'm not asking you for applying the change, I just offered you the solution :wink:  your version is fine too :v:

MatousJobanek (approved):
Thanks a lot for addressing my comments :pray:

xcoulon:
> Thanks a lot for addressing my comments 🙏

thanks for you review ;)

PR #147, opened 2021-01-28
Multi member

rajivnathan:
Prerequisite PR for https://github.com/codeready-toolchain/toolchain-e2e/pull/226

Adds a new flag `-ms` or `--multi-ns` that appends the namespace to the names of resources that may conflict when a member operator is deployed more than once to the same cluster. This is primarily for e2e tests to be able to test the multi-member logic using a single cluster.


MatousJobanek (approved):
Nice :+1: 
I didn't realize that the names would cause also a problem with the conflicts, good catch!

scripts/add-cluster.sh
```diff
@@ -256,21 +267,31 @@ else
...
 fi
 oc create secret generic ${SECRET_NAME} --from-literal=token="${SA_TOKEN}" --from-literal=ca.crt="${SA_CA_CRT}" -n ${CLUSTER_JOIN_TO_OPERATOR_NS} ${OC_ADDITIONAL_PARAMS}
 
+if [[ ${MULTI_NS} != "true" ]]; then
+    TOOLCHAINCLUSTER_NAME=${JOINING_CLUSTER_TYPE_NAME}-${JOINING_CLUSTER_NAME}
+    OWNER_CLUSTER_NAME=${CLUSTER_JOIN_TO}-${CLUSTER_JOIN_TO_NAME}
+else
+    echo "Multi-namespace member mode"
+    TOOLCHAINCLUSTER_NAME=${JOINING_CLUSTER_TYPE_NAME}-${JOINING_CLUSTER_NAME}-2
+    OWNER_CLUSTER_NAME=${CLUSTER_JOIN_TO}-${CLUSTER_JOIN_TO_NAME}-2
```
MatousJobanek, line 283: btw, just a small comment:
instead of introducing a boolean param `--multi-ns` you could add for example `--suffix-name` parameter that would add the suffix after both the `ToolchainCluster` CR as well as the ServiceAccount name.
We could call:
```
add-cluster.sh --suffix-name "-2"
```
so it would create SA `toolchaincluster-member-2` or `e2e-service-account-2`
analogically also the ToolchainCluster CRs.

It would also simplify the logic since you could only append the variable:
```
 ${JOINING_CLUSTER_TYPE_NAME}-${JOINING_CLUSTER_NAME}${SUFFIX_NAME}
```
without introducing the if-else statements

This is a proposal for improvement, it's nothing I would insist on nor require to be implemented in this PR ;-)

rajivnathan: That's a great idea! I'll do that.

scripts/add-cluster.sh
```diff
@@ -224,10 +230,16 @@ echo ${CLUSTER_JOIN_TO_OPERATOR_NS}
 login_to_cluster ${JOINING_CLUSTER_TYPE}
 
 if [[ ${JOINING_CLUSTER_TYPE_NAME} != "e2e" ]]; then
-    SA_NAME="toolchaincluster-${JOINING_CLUSTER_TYPE}-operator"
+    SA_NAME="toolchaincluster-${JOINING_CLUSTER_TYPE}"
+    if [[ ! -z ${MULTI_MEMBER} ]]; then
+      SA_NAME="${OPERATOR_NS}"
+    fi
```
MatousJobanek, line 236: couldn't you do the same as for `ToolchainCluster` CRs?
```suggestion
    SA_NAME="toolchaincluster-${JOINING_CLUSTER_TYPE}${MULTI_MEMBER}"
```

scripts/add-cluster.sh
```diff
@@ -256,21 +268,31 @@ else
...
 oc create secret generic ${SECRET_NAME} --from-literal=token="${SA_TOKEN}" --from-literal=ca.crt="${SA_CA_CRT}" -n ${CLUSTER_JOIN_TO_OPERATOR_NS} ${OC_ADDITIONAL_PARAMS}
 
+TOOLCHAINCLUSTER_NAME=${JOINING_CLUSTER_TYPE_NAME}-${JOINING_CLUSTER_NAME}
+OWNER_CLUSTER_NAME=${CLUSTER_JOIN_TO}-${CLUSTER_JOIN_TO_NAME}
+
+if [[ ! -z ${MULTI_MEMBER} ]]; then
+    echo "Multi-namespace member mode"
+    TOOLCHAINCLUSTER_NAME=${JOINING_CLUSTER_TYPE_NAME}-${JOINING_CLUSTER_NAME}-${MULTI_MEMBER}
+    OWNER_CLUSTER_NAME=${CLUSTER_JOIN_TO}-${CLUSTER_JOIN_TO_NAME}-${MULTI_MEMBER}
+fi
```
MatousJobanek, line 285: Why not just:
```suggestion
TOOLCHAINCLUSTER_NAME=${JOINING_CLUSTER_TYPE_NAME}-${JOINING_CLUSTER_NAME}${MULTI_MEMBER}
OWNER_CLUSTER_NAME=${CLUSTER_JOIN_TO}-${CLUSTER_JOIN_TO_NAME}${MULTI_MEMBER}
```

scripts/add-cluster.sh
rajivnathan: That's true, if -mm is not used then ${MULTI_MEMBER} should be empty and will result in the same naming pattern before this PR so your suggestion should work.

PR #148, opened 2021-02-08
refactor: use options to create or update object

xcoulon:
move the `owner` and `forceUpdate` parameters as options,
add add an option to avoid saving the configuration
(which causes troubles when the resource is too big, for example,
a ConfigMap with large JSON files)

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -19,10 +19,12 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+// LastAppliedConfigurationAnnotationKey the key to save the last applied configuration in the resource annotations
 const LastAppliedConfigurationAnnotationKey = "toolchain.dev.openshift.com/last-applied-configuration"
 
 var log = logf.Log.WithName("apply_client")
 
+// ApplyClient the client to use when creating or updating objects
 type ApplyClient struct {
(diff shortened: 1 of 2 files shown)


pkg/client/client.go
```diff
@@ -33,44 +33,81 @@ func NewApplyClient(cl client.Client, scheme *runtime.Scheme) *ApplyClient {
 	return &ApplyClient{cl: cl, scheme: scheme}
 }
 
+type createOrUpdateObjectConfiguration struct {
+	owner             v1.Object
+	forceUpdate       bool
+	saveConfiguration bool
+}
+
+func newCreateOrUpdateObjectConfiguration(options ...CreateOrUpdateObjectOption) createOrUpdateObjectConfiguration {
```
MatousJobanek, line 42: I guess that we could make the names shorter - something like `applyObjectConfig`, but it's not critical

MatousJobanek (approved):
Nice improvement :+1:

pkg/client/client.go
xcoulon: good idea! Let me rename `createOrUpdateObjectConfiguration` to `applyObjectConfiguration` (I tend to prefer full words)

MatousJobanek (approved):

PR #151, opened 2021-02-11
Overwrite Preferred UserNameClaim For e2e-tests

tinakurian:
Adding the ability to overwrite the username in order to add an e2e-test to test that the registration service does not create signup resources for users who signup with the crtadmin suffix in their username. 

Jira: Jira: https://issues.redhat.com/browse/CRT-913
Related PR: https://github.com/codeready-toolchain/registration-service/pull/170, https://github.com/codeready-toolchain/host-operator/pull/373, https://github.com/codeready-toolchain/toolchain-e2e/pull/237

### DIFF
--- a/pkg/usersignup/usersignup.go
+++ b/pkg/usersignup/usersignup.go
@@ -1,0 +1,84 @@
+package usersignup
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+	"strings"
+
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/validation"
+)
+
+var (
+	specialCharRegexp = regexp.MustCompile("[^A-Za-z0-9]")
+	onlyNumbers       = regexp.MustCompile("^[0-9]*$")
+)
+
+func GenerateCompliantUsername(instance *toolchainv1alpha1.UserSignup, cl client.Client, forbiddenUsernamePrefixes []string ) (string, error) {
+	replaced := TransformUsername(instance.Spec.Username)
+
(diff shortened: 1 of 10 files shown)


alexeykazakov (approved):

MatousJobanek (approved):

alexeykazakov (approved):

pkg/usersignup/usersignup.go
```diff
@@ -0,0 +1,84 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/validation"
+)
+
+var (
+	specialCharRegexp = regexp.MustCompile("[^A-Za-z0-9]")
+	onlyNumbers       = regexp.MustCompile("^[0-9]*$")
+)
+
+func GenerateCompliantUsername(instance *toolchainv1alpha1.UserSignup, cl client.Client, forbiddenUsernamePrefixes []string ) (string, error) {
```
MatousJobanek, line 22: Correct me if I'm wrong, but this should not be here. We wanted to move only the tranformation of the name - just `TransformUsername` function, not check of the duplications. This gets already existing MURs and appends only the number suffix (if there is a name conflict), so it should not have any significant impact on the resulting name format.
In addition, at the time of the UserSignup creation, it's still not clear if any suffix will be added or not - this is evaluated when the UserSignup is approved, which can take some time.

MatousJobanek (changes requested):
Please, move the `GenerateCompliantUsername` function back to host-operator

pkg/usersignup/usersignup.go
tinakurian: Ohhh okay, sorry, i moved both, I'll move GenerateCompliantUsername back in that case..

tinakurian: @MatousJobanek fixed here: c34f4b515b6cfe887adf2a55f09ed4b8a26e9b8c & 8325ca1002576ab4a71f891d0fdedc548580fb8f

MatousJobanek (approved):
:+1:

xcoulon (approved):

PR #154, opened 2021-03-01
Add new labels and generation predicate

rajivnathan:
Related issue: https://issues.redhat.com/browse/CRT-941

Adds a new predicate to only reconcile update events if Generation or Labels have changed.

### DIFF
--- a/pkg/predicate/predicate.go
+++ b/pkg/predicate/predicate.go
@@ -1,8 +1,11 @@
 package predicate
 
 import (
+	"reflect"
+
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
 
 var log = logf.Log.WithName("generation_not_changed_predicate").WithName("eventFilters")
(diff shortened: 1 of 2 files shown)


pkg/predicate/predicate.go
```diff
@@ -49,3 +52,40 @@ func (OnlyUpdateWhenGenerationNotChanged) Delete(e event.DeleteEvent) bool {
...
+
+	// reconcile if the labels have changed
+	if !reflect.DeepEqual(e.MetaOld.GetLabels(), e.MetaNew.GetLabels()) {
+		return true
+	}
+
+	if e.MetaNew.GetGeneration() == e.MetaOld.GetGeneration() && e.MetaNew.GetGeneration() != 0 {
+		return false
+	}
+	return true
```
MatousJobanek, line 90: What is the purpose of `e.MetaNew.GetGeneration() != 0`? I think that generation starts from `1` so it shouldn't be equal to zero ever.
It could be simplified to:
```suggestion
	return e.MetaNew.GetGeneration() != e.MetaOld.GetGeneration() || e.MetaNew.GetGeneration() == 0
```

MatousJobanek, line 90: or, in case you dropped the `e.MetaNew.GetGeneration() != 0` then it could be simplified to:
```suggestion
	return !reflect.DeepEqual(e.MetaOld.GetLabels(), e.MetaNew.GetLabels()) ||
		e.MetaNew.GetGeneration() != e.MetaOld.GetGeneration()
```

MatousJobanek (commented):
Thanks for moving it to common and adding the test. 
Could you please take a look at the if statements?

pkg/predicate/predicate.go
rajivnathan: It was based on the one from operator sdk: https://github.com/operator-framework/operator-sdk/blob/ef0bc885e40e6c86c0c079bea612b4eb908f6ec5/pkg/predicate/predicate.go#L56 but looks like that is outdated now and kubernetes has it the way you're suggesting: https://github.com/kubernetes-sigs/controller-runtime/blob/bdb3f845d6e2510d441f593fb420781e8a05f894/pkg/predicate/predicate.go#L170

So I'll go ahead and use this suggestion.

MatousJobanek (approved):
Thanks :+1:

PR #156, opened 2021-03-23
Add state management helper functions

sbryzak:
Add helper functions to manage UserSignup states.


pkg/states/state_manager.go
```diff
@@ -0,0 +1,35 @@
+package states
+
+import "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
+
+func Deactivating(userSignup *v1alpha1.UserSignup) bool {
+	return contains(userSignup.Spec.States, v1alpha1.UserSignupStateDeactivating)
+}
+
+func SetUserSignupState(userSignup *v1alpha1.UserSignup, state v1alpha1.UserSignupState) {
```
alexeykazakov, line 9: Maybe AddUserSignupSrate() instead? Just to avoid confusion since this function do not replace the current states and only appends a new one.

Or maybe we should avoid generic exported functions here at all because some states should replace specific states while others can be just added or removed.

So maybe something like SetDeactivatingState(), UnsetDeactivatingState(), SetDeactivated(), SetApproved(), etc.

WDYT?

sbryzak: I like the idea of specific functions like `SetDeactivating()`, `SetVerificationRequired()` etc, but perhaps these functions could take a `bool` param to indicate whether the state should be set or unset?

alexeykazakov: Yep. +1 to using a bool param.

pkg/states/state_manager.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func SetDeactivating(userSignup *v1alpha1.UserSignup, val bool) {
+	if val && !contains(userSignup.Spec.States, v1alpha1.UserSignupStateDeactivating) {
+		userSignup.Spec.States = append(userSignup.Spec.States, v1alpha1.UserSignupStateDeactivating)
+	}
+
+	if !val && contains(userSignup.Spec.States, v1alpha1.UserSignupStateDeactivating) {
+		userSignup.Spec.States = remove(userSignup.Spec.States, v1alpha1.UserSignupStateDeactivating)
+	}
```
alexeykazakov, line 50: I guess we can add a generic function like:
```
func setState(usersignup, val, stateType)
```
And reuse it.

pkg/states/state_manager.go
```diff
@@ -0,0 +1,83 @@
...
+		!VerificationRequired(userSignup) &&
+		!Deactivated(userSignup)
+}
+
+func Approved(userSignup *v1alpha1.UserSignup) bool {
+	return contains(userSignup.Spec.States, v1alpha1.UserSignupStateApproved)
+}
+
+func SetApproved(userSignup *v1alpha1.UserSignup, val bool) {
+	if val && !contains(userSignup.Spec.States, v1alpha1.UserSignupStateApproved) {
```
alexeykazakov, line 16: Shouldn't we replace some states by new ones instead of always appending the new one?

For example, when setting to Approved we should remove VerificationRequired. When setting Deactivated we should remove Approved.

pkg/states/state_manager.go
sbryzak: That makes a lot of sense. :)

pkg/states/state_manager.go
sbryzak: Sure thing, we just need to define the business rules here.  I've implemented the approved and deactivated cases already as you describe.

alexeykazakov (approved):

pkg/states/state_manager.go
```diff
@@ -0,0 +1,77 @@
...
+	return Approved(userSignup) &&
+		!VerificationRequired(userSignup) &&
+		!Deactivated(userSignup)
+}
+
+func Approved(userSignup *v1alpha1.UserSignup) bool {
+	return contains(userSignup.Spec.States, v1alpha1.UserSignupStateApproved)
+}
+
+func SetApproved(userSignup *v1alpha1.UserSignup, val bool) {
```
MatousJobanek, line 15: I'm slightly missing the purpose of the boolean parameter? Why should I want to call `SetApproved(userSignup, false)` that doesn't make sense to me - if I want to approve a user, then I just call `SetApproved(usersignup)` or just `Approve(userSignup)`

or is there any use-case when we would need to call `SetApproved(userSignup, false)`?

pkg/states/state_manager.go
```diff
@@ -0,0 +1,83 @@
...
+
+func Active(userSignup *v1alpha1.UserSignup) bool {
+	return Approved(userSignup) &&
+		!VerificationRequired(userSignup) &&
+		!Deactivated(userSignup)
+}
+
+func Approved(userSignup *v1alpha1.UserSignup) bool {
+	return contains(userSignup.Spec.States, v1alpha1.UserSignupStateApproved)
+}
```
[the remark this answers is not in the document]
MatousJobanek: btw, shouldn't be the function of the name rather `IsApproved` ?

pkg/states/state_manager.go
```diff
@@ -0,0 +1,77 @@
...
+
+func SetApproved(userSignup *v1alpha1.UserSignup, val bool) {
+	setState(userSignup, v1alpha1.UserSignupStateApproved, val)
+
+	if val {
+		setState(userSignup, v1alpha1.UserSignupStateVerificationRequired, false)
+	}
+}
+
+func VerificationRequired(userSignup *v1alpha1.UserSignup) bool {
```
MatousJobanek, line 23: the same here with the name of the function - it makes more sense to call `IsVerificationRequired` when the function returns a boolean

pkg/states/state_manager.go
```diff
@@ -0,0 +1,77 @@
...
+
+func Approved(userSignup *v1alpha1.UserSignup) bool {
+	return contains(userSignup.Spec.States, v1alpha1.UserSignupStateApproved)
+}
+
+func SetApproved(userSignup *v1alpha1.UserSignup, val bool) {
+	setState(userSignup, v1alpha1.UserSignupStateApproved, val)
+
+	if val {
+		setState(userSignup, v1alpha1.UserSignupStateVerificationRequired, false)
```
MatousJobanek, line 19: how about also setting:
```
setState(userSignup, v1alpha1.UserSignupStateDeactivated, false)
```
I guess that we should make sure that the deactivated state is not present when approving a user.

MatousJobanek (commented):
Sorry for my late review @sbryzak 
Could you please take a look at my comments?

pkg/states/state_manager.go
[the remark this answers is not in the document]
sbryzak: I'm not so sure that `Active` is the same as `Approved`.  We have a well established concept of an "active" user, for a couple of examples see [here](https://github.com/codeready-toolchain/registration-service/blob/master/pkg/kubeclient/signup.go#L90) or [here](https://github.com/codeready-toolchain/api/blob/master/pkg/apis/toolchain/v1alpha1/usersignup_types.go#L80), and this function simply enforces this convention.

For the function name, it is inline with the Go language guidelines - see [here](https://golang.org/doc/effective_go#Getters) for a description of getter/accessor naming recommendations.

pkg/states/state_manager.go
sbryzak: We could probably omit the parameter in the case of `SetApproved`, however in other cases it's necessary - for example calling `SetVerificationRequired(userSignup, false)` once the user successfully verifies.  On one hand I'm open to removing the parameter in this case, however on the other hand it seems that consistency is important also.  I guess I'm sitting on the fence here ;)

pkg/states/state_manager.go
sbryzak: See above comment about Golang naming recommendations.

pkg/states/state_manager.go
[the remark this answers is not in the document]
MatousJobanek: > I'm not so sure that Active is the same as Approved. We have a well established concept of an "active" user, for a couple of examples see here or here, and this function simply enforces this convention.

ok, I wasn't aware of these usages. TBH, I was always trying to avoid using the "active" word in the code, since it doesn't really represent the "active user", but rather an "approved and provisioned user". I would understand (with my poor English :wink:) "active user" as a user who is logged in to Dev Sandbox.

My point was that we use the `approved` word in:
* usersignup spec
* as a reason in `Completed` condition type
* in UserSignup controller to check if user is provisioned
* in sandbox-cli commands
* and mainly also in our metrics

So I thought that it would be fine to use the same word also for the name of this function. WDYT @alexeykazakov ?

Btw - I just shared my point of view and my feeling, but I don't insist on changing it. If you agree that using `Active` is fine, then I'm fine too and I'll get used to it :wink: 

> For the function name, it is inline with the Go language guidelines - see here for a description of getter/accessor naming recommendations.

thanks for the link :+1:

pkg/states/state_manager.go
MatousJobanek: fair enough - let's keep it as is it

pkg/states/state_manager.go
MatousJobanek: :+1:

pkg/states/state_manager.go
MatousJobanek: @sbryzak WDYT about this suggestion ^^^?

pkg/states/state_manager.go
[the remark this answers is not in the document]
alexeykazakov: @MatousJobanek you brought a legit concern and we have discussed it with Shane. We currently have a two very similar but sometimes confusing notions/states: approved and active.
What we want to have is some definition of a desired state which will tell the controller that everything is good (from the signup point of view) and we want to provision the account.
Do we call it "approved", "active", or something else is an open question.
But since we do not plan to use this particular state(s) right away and the only state which will be used right now is "deprovisioning" we agreed on leaving it as is. As a separate task we will review our approved/active state notions and since we will need to refactor our controllers (and cli) to migrate to the new Spec.State for provisioning new users anyway we do not have to solve this problem right now.
Bottom line. Let's keep it as is and have a detailed discussion when we ready to start working on the Spec.Approved/Deactivated controller refactoring.

pkg/states/state_manager.go
alexeykazakov: And here too. We might change this when we decide what we want to have as a definition of the desired "active/approved" state.

pkg/states/state_manager.go
sbryzak: Good idea, I'll be adding a subtask for this in the following story: https://issues.redhat.com/browse/CRT-1003

PR #158, opened 2021-03-30
fix: delete ClusterRoleBinding before applying it

MatousJobanek:
We need to delete ClusterRoleBinding before applying because we cannot update `roleRef` and this is needed when we need to rerun the setup in one cluster multiple times.
We didn't face it before because OLM created ClusterRoles with the fixed name


rajivnathan (approved):
Nice!

PR #162, opened 2021-05-03
Set deactivating state to false when deactivated set to true

sbryzak:
This PR set the UserSignup deactivating state to false when deactivated is set to true.

Part of https://issues.redhat.com/browse/CRT-1006


pkg/states/state_manager.go
```diff
@@ -45,6 +45,7 @@ func SetDeactivated(userSignup *v1alpha1.UserSignup, val bool) {
 
 	if val {
 		setState(userSignup, v1alpha1.UserSignupStateApproved, false)
+		setState(userSignup, v1alpha1.UserSignupStateDeactivating, false)
```
MatousJobanek, line 48: don't forget to modify also the unit test

sbryzak: Done.

alexeykazakov (approved):

PR #176, opened 2021-05-18
adding BlockOwnerDeletion at MUR

ranakan19:
setting BlockOwnerDeletion to true in OwnerReference for MUR, for foreground Garbage Collection

### DIFF
--- a/pkg/test/masteruserrecord/master_user_record.go
+++ b/pkg/test/masteruserrecord/master_user_record.go
@@ -91,6 +91,11 @@ func NewMasterUserRecord(t *testing.T, userName string, modifiers ...MurModifier
 			Labels: map[string]string{
 				templateTierHashLabelKey(DefaultNSTemplateTier.Name): hash,
 			},
+			OwnerReferences: []metav1.OwnerReference{
+				{
+					BlockOwnerDeletion: func(b bool) *bool { return &b }(true),
+				},
+			},
 		},
 		Spec: toolchainv1alpha1.MasterUserRecordSpec{
 			UserID:       userID,


alexeykazakov (approved):

MatousJobanek (commented):
May I ask what the purpose of this change is? This code is used only for test purposes.
In the production code, we call `SetControllerReference` which should set that field to true:
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/controller/controllerutil/controllerutil.go#L74-L81

ranakan19:
Don't know if we still need this, since block owner deletion is set true [here](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/controller/controllerutil/controllerutil.go#L79) which is called from [here](https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/usersignup/usersignup_controller.go#L511)
On second thought, it wouldn't harm and only ensures foreground deletion; WDYT?

ranakan19:
> May I ask what the purpose of this change is? This code is used only for test purposes.
> In the production code, we call `SetControllerReference` which should set that field to true:
> https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/controller/controllerutil/controllerutil.go#L74-L81

sorry @MatousJobanek i saw your comment later, you are right, i probably don't need it.

pkg/test/masteruserrecord/master_user_record.go
```diff
@@ -91,6 +91,11 @@ func NewMasterUserRecord(t *testing.T, userName string, modifiers ...MurModifier
 			Labels: map[string]string{
 				templateTierHashLabelKey(DefaultNSTemplateTier.Name): hash,
 			},
+			OwnerReferences: []metav1.OwnerReference{
+				{
+					BlockOwnerDeletion: func(b bool) *bool { return &b }(true),
```
xcoulon, line 96: is it something you could test, to ensure it cannot be removed by accident in the future?

xcoulon (changes requested):

alexeykazakov:
@ranakan19 any updates on this PR?

ranakan19:
closing as this change is not needed.

PR #177, opened 2021-05-20
chore: update api with new format & kubebuilder

MatousJobanek:
result of codeready-toolchain/api#237
 * small changes in go.mod so we can use the new controller-tools 0.4.1 together with client-go 0.18.3
 * update all imports to api repo/types

### DIFF
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -8,7 +8,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/codeready-toolchain/api/api/v1alpha1"
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
 	"github.com/codeready-toolchain/toolchain-common/pkg/client"
 	"github.com/codeready-toolchain/toolchain-common/pkg/template"
 	"github.com/codeready-toolchain/toolchain-common/pkg/test"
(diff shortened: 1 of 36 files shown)


pkg/client/client_test.go
```diff
@@ -8,7 +8,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/codeready-toolchain/api/pkg/apis"
+	"github.com/codeready-toolchain/api/api/v1alpha1"
```
xcoulon, line 11: Could we all agree to use `toolchainv1alpha1` as an alias for `github.com/codeready-toolchain/api/api/v1alpha1` in all our codebase? 
IMHO, `v1alpha1` alone is a bit meaningless. Also, we've been doing something similar for other packages we import. Eg:
```
metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```

xcoulon (approved):
I left a comment, but nothing blocking. We can discuss about it and/or address it later

pkg/client/client_test.go
MatousJobanek: sure - done in f9d3333

alexeykazakov (approved):

pkg/client/client_test.go
xcoulon: awesome, thanks!

PR #182, opened 2021-06-04
build: use Codecov GitHub Action

xcoulon:


xcoulon:
@MatousJobanek @alexeykazakov I wonder if we could switch from OpenShift CI to GitHub Action for our "build" and "test with coverage" jobs, and use the [codecov-action](https://github.com/codecov/codecov-action) to upload the coverage reports on codecov.io. 
WDYT?

alexeykazakov (approved):
Nice! But shouldn't we also remove report upload from our make target?

alexeykazakov:
Ah.. this is only about uploading the report from master... we still need to get the report for PRs, right?

xcoulon:
> Nice! But shouldn't we also remove report upload from our make target?

yes, that's right ;)

xcoulon:
> Ah.. this is only about uploading the report from master... we still need to get the report for PRs, right?

hmm, no, it should work in both cases, since the jobs are triggered for 
```
on:
  push:
    branches:
      - master
    tags-ignore:
      - '*.*'
  pull_request:
    branches:
      - master
 ```

xcoulon:
> > Nice! But shouldn't we also remove report upload from our make target?
> 
> yes, that's right ;)

I suggest I remove this makefile goal once the OpenShift CI job has be deleted: https://github.com/openshift/release/pull/19022

PR #184, opened 2021-06-09
Replace satori uuid lib with gofrs

sbryzak:
The satori uuid library is no longer maintained and contains some serious flaws ([details](https://github.com/gofrs/uuid#project-history)), we shouldn't be using it any more.


MatousJobanek (approved):

alexeykazakov (approved):
@sbryzak can you please update it in the other repos too?

PR #186, opened 2021-06-10
Use gofrs uuid in other places as well

rajivnathan:
This is a follow up to https://github.com/codeready-toolchain/toolchain-common/pull/184. There were a few other places that needed to be updated to use gofrs uuid


alexeykazakov (approved):
No changes in go.sum?
Even after `go mod tidy`?

PR #188, opened 2021-06-14
Add toolchainconfig assertions

rajivnathan:
Add common test assertions for toolchainconfig resources.

### DIFF
--- a/pkg/test/config/toolchainconfig_assertion.go
+++ b/pkg/test/config/toolchainconfig_assertion.go
@@ -1,0 +1,71 @@
+package config
+
+import (
+	"context"
+
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
+	"github.com/codeready-toolchain/toolchain-common/pkg/test"
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type ToolchainConfigAssertion struct {
+	config         *toolchainv1alpha1.ToolchainConfig
+	client         client.Client
+	namespacedName types.NamespacedName
+	t              test.T
+}
+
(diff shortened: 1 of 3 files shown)


ranakan19 (approved):
LGTM

pkg/test/config/toolchainconfig_assertion.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func (a *ToolchainConfigAssertion) loadToolchainConfig() error {
+	if a.client != nil {
+		toolchainConfig := &toolchainv1alpha1.ToolchainConfig{}
+		err := a.client.Get(context.TODO(), a.namespacedName, toolchainConfig)
+		a.config = toolchainConfig
+		return err
+	}
+	return nil
```
ranakan19, line 28: although an unlikely scenario, when there is no client - would returning a nil be perceived as positive case? I guess what my question is whether we want to throw an error when there is no client - like no client so can't load anything, hence error or no client so no action.

alexeykazakov (approved):

sbryzak (approved):

pkg/test/config/toolchainconfig_assertion.go
rajivnathan: The other assertion files we have also return nil if no client is provided: https://github.com/codeready-toolchain/host-operator/blob/master/test/configmap_assertions.go#L23 and https://github.com/codeready-toolchain/host-operator/blob/ad2a9c6c1dc6ab57f1ed4fde554a23eb31d850f8/test/serviceaccount_assertions.go#L26 for example.

I guess the intent here is to be able to use the assertions without the need for a client. eg. If I wanted to have a quick unit test for toolchainconfig without a client I could do something like
```
customAssertion := &ToolchainConfigAssertion{
  config: customToolchainConfig,
  t: a.T()
}
customAssertion.HasConditions(ToSyncFailure()).HasSyncErrors(map[string]string{"missing-member": "specific member configuration exists but no matching toolchaincluster was found"})
```

PR #189, opened 2021-06-15
feat: migrate common to operator-sdk 1.8.0

MatousJobanek:
## Migrates toolchain-common repo to operator-sdk 1.8.0

* updates dependencies to use k8s v0.20.2 - as a result:
    *  updates the mappers used in controllers to match the latest API - it doesn't expect a type, but a function
    * migrates most of the code from `runtime.Object` to `client.Object`, but here is still for an improvement - will be done in next PR
    * use the new API for defining the watched resources in controllers: `ctrl.NewControllerManagedBy`
    * changes `Reconcile(request ctrl.Request)` to `Reconcile(ctx context.Context, request ctrl.Request)`
* don't store log inside the `Reconciler` type but create it using `log.FromContext(ctx)`
* add functions `GetWatchNamespace` & `GetOperatorName` as a replacement of those that were provided by operator-sdk. The latest version doesn't have it - mainly because the default recommended install mode is `allNamespaces`

### TODO in the next PRs
* additional cleanup & refactoring as a result of introducing  `client.Object` 
* move `add-cluster.sh` script to toolchain-cicd repo


pkg/client/toolchain_object.go
```diff
@@ -97,3 +98,21 @@ func NewComparableToolchainObject(ob runtime.Object, compare CompareToolchainObj
 func (o *comparableToolchainObjectImpl) IsSame(otherObject ToolchainObject) (bool, error) {
 	return o.compare(o.ToolchainObject, otherObject)
 }
+
+func SortedComparableToolchainObjects(objects []ComparableToolchainObject) []ComparableToolchainObject {
```
xcoulon, line 102: could you add a comment to explain that this function sorts the objects by their name, and maybe even rename it to something like `SortToolchainObjectsByName`?

pkg/configuration/load_test.go
```diff
@@ -60,7 +59,7 @@ func TestLoadFromConfigMap(t *testing.T) {
 		}
 		cl := test.NewFakeClient(t, createConfigMap("test-config", "toolchain-host-operator", data))
 
-		cl.MockGet = func(ctx context.Context, key client.ObjectKey, obj runtime.Object) error {
+		cl.MockGet = func(ctx context.Context, key client.ObjectKey, obj client.Object) error {
```
xcoulon, line 62: same here: `runtimeclient` alias for the `sigs.k8s.io/controller-runtime/pkg/client` pkg

pkg/predicate/predicate.go
```diff
@@ -31,11 +27,7 @@ func (OnlyUpdateWhenGenerationNotChanged) Update(e event.UpdateEvent) bool {
 		log.Error(nil, "Update event has no new runtime object for update", "event", e)
 		return false
 	}
-	if e.MetaNew == nil {
-		log.Error(nil, "Update event has no new metadata", "event", e)
-		return false
-	}
-	return e.MetaNew.GetGeneration() == e.MetaOld.GetGeneration()
+	return e.ObjectNew.GetGeneration() == e.ObjectOld.GetGeneration()
```
xcoulon, line 30: just curious: what was wrong with the 2 cases that you deleted? didn't they apply to creation and deletion?

pkg/template/processor.go
```diff
@@ -1,15 +1,17 @@
 package template
 
 import (
+	"fmt"
 	"math/rand"
 	"time"
 
-	"github.com/codeready-toolchain/toolchain-common/pkg/client"
+	commonclient "github.com/codeready-toolchain/toolchain-common/pkg/client"
```
xcoulon, line 8: same remark here about aliases, for the sake of consistency with other files ;)

pkg/client/client.go
```diff
@@ -78,12 +78,21 @@ func SaveConfiguration(saveConfiguration bool) ApplyObjectOption {
 	}
 }
 
+// ApplyRuntimeObject casts the provided object to client.Object and calls ApplyClient.ApplyObject method
+func (p ApplyClient) ApplyRuntimeObject(obj runtime.Object, options ...ApplyObjectOption) (bool, error) {
+	clientObj, ok := obj.(client.Object)
```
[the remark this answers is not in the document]
MatousJobanek: I would rather use `client` alias for this package. We can change it in the tests so it is consistent with the production code, but let's do all this craziness around import aliases in separate PR. We can create jira story for this and track all planed/needed renaming there. Is it OK for you?

pkg/client/toolchain_object.go
```diff
@@ -2,12 +2,13 @@ package client
...
 import (
 	"fmt"
+	"sort"
 
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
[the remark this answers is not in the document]
MatousJobanek: and the same answer - let's use the default `client` for this package

pkg/template/processor.go
MatousJobanek: yeah, this should have been `applyclient`

pkg/predicate/predicate.go
MatousJobanek: that's the result of the API changes in the latest k8s - the objects that are provided by the event is not a `runtime.Client` but `client.Client`

pkg/client/toolchain_object.go
MatousJobanek: sure - done in 69cce42

pkg/template/processor.go
MatousJobanek: hmmm... not sure what should be the right alias for that - in other cases we use:
* `applycl`
* `commonclient`
* `commoncl`
* `applyclientlib`

:man_shrugging:

MatousJobanek: let's keep it as it is and let's add it to the effort of aligning the import aliases.

pkg/client/client.go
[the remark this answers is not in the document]
alexeykazakov: Please no more big changes in package allies! :)
It's cosmetic but creates tons of conflicts.

alexeykazakov (approved):

pkg/client/client_test.go
```diff
@@ -813,5 +814,7 @@ func addToScheme(t *testing.T) *runtime.Scheme {
 	require.NoError(t, err)
 	err = toolchainv1alpha1.AddToScheme(s)
 	require.NoError(t, err)
+	err = templatev1.Install(s)
```
rajivnathan, line 817: Out of curiosity, why does Install need to be called now, whereas before it was not?

pkg/configuration/load_test.go
```diff
@@ -236,3 +235,68 @@ func createConfigMap(name, namespace string, data map[string]string) *v1.ConfigM
 		Data: data,
 	}
 }
+
+func TestGetWatchNamespaceWhenSet(t *testing.T) {
+	// given
+	restore := test.SetEnvVarAndRestore(t, "WATCH_NAMESPACE", "member-operator")
```
rajivnathan, line 241: Should we add a test for when it's set to empty `""`? Similar to OPERATOR_NAME below?

rajivnathan (approved):
Nice!

pkg/client/client_test.go
MatousJobanek: Why do we need to call the `Install` here? The answer is simple - because one of the tests processes a template content into a `Template` object. More interesting is the second part of your question - why didn't we have to call it in the previous code. 

TBH with you - this scheme is a kind of magic black box for me, so I'm not sure. There were some changes in the default set of the API schemes that are automatically added to that instance `scheme.Scheme`, but the template is not one of them. There were also some changes in the conversion logic that converts the parsed `Unstructured` objects into the typed object, but I didn't dig so much deep into it. 
I also printed out (in the original code) the content of the scheme before it is passed to the `Convert` function and the template API scheme is not there, so I have no clue how it could work. My guess is that the original conversion code from the apimachinery library did some default assumption or added the template scheme there - but as I said - I'm not sure :man_shrugging:

pkg/configuration/load_test.go
MatousJobanek: This is a very good point and it also brings an important question - shouldn't it fail when it is set to an empty string...? Since we want to use namespace-scoped operators, then my guess is that yes - it should fail.
Fixed and improved here: 7979364

pkg/template/processor.go
xcoulon: ok, sounds good. This is a cosmetic remark and not a blocker for this PR for sure, but if we could all agree on a policy for aliases, that would be great IMO

pkg/predicate/predicate.go
xcoulon: I'm not sure I understand your response. My question was about the removal of
```
if e.MetaOld == nil {
	log.Error(nil, "Update event has no old metadata", "event", e)
	return false
}
...
if e.MetaNew == nil {
	log.Error(nil, "Update event has no new metadata", "event", e)
	return false
}
```
in this func.

MatousJobanek: The ` event.UpdateEvent` doesn't contain the `e.MetaOld` nor `e.MetaNew` now. Instead of `runtime.Client` it provides `client.Client` which has already access to all meta information. So, there is no reason for providing that meta thingy as a separate parameter and thus we just check if the `client.Client` objects are nil or not - that's sufficient. If there is no meta, then there is no `client.Client` object.
This same checks are also used in the built-in predicates

Sorry, for not being clear before.

xcoulon: ok, thanks for the clarification

pkg/client/client_test.go
rajivnathan: Thanks for taking the time to explain it so thoroughly

pkg/configuration/load_test.go
rajivnathan: Agreed, I think it should fail when empty as well as well.

pkg/client/client.go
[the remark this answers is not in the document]
xcoulon: ok, let's talk about aliases another time. I'm still convinced it's worth it, but I understand it can cause annoying merge conflicts :/

PR #190, opened 2021-06-22
build: run GitHub Actions on Ubuntu 20.04

xcoulon:
using an LTS version instead of `lastest`

Fixes https://issues.redhat.com/browse/CRT-1033

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -12,7 +12,7 @@ on:
 jobs:
   test-with-coverage:
     name: Test with Coverage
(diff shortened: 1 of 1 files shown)


MatousJobanek (commented):
I think that we don't need it - `20.04` and `latest` should be the same: https://github.com/actions/virtual-environments#available-environments

alexeykazakov (approved):

alexeykazakov:
Matous, latest is 20.04 until there is a newer version.
I'm fine with both options but though that using some specific version could be more stable.

PR #196, opened 2021-07-05
Common config functions

rajivnathan:
Some common functions used in both host and member operators for retrieving config from ToolchainConfig and MemberOperatorConfig.


pkg/configuration/get_configuration.go
```diff
@@ -0,0 +1,45 @@
+package configuration
```
alexeykazakov, line 1: I guess it could be just `pkg/configuration/get.go`

pkg/configuration/get_configuration.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+func GetString(value *string, defaultValue string) string {
+	if value != nil {
+		return *value
+	}
+	return defaultValue
+}
+
+func GetDuration(value *string, defaultValue time.Duration) time.Duration {
```
alexeykazakov, line 26: So, we are ignoring the case when `value` is not nil but invalid? Let's add a comment stating that the default value is returned when `value` is nil or invalid and can't be parsed as a duration.

alexeykazakov (approved):

pkg/configuration/get_configuration.go
rajivnathan: Done in 0c83f63f5b77f7e1b93559b5cffc8ff2cdc21725

I also renamed load_configuration.go to load.go for consistency.

pkg/configuration/get_configuration.go
rajivnathan: Done in 689af28e98b281b7dc11a4f88dfd23249163e572

PR #199, opened 2021-07-19
script: hide CABundle when printing the ToolchainCluster CR

xcoulon:
it's verbose and probably useless if we need to debug the tests.

### DIFF
--- a/scripts/add-cluster.sh
+++ b/scripts/add-cluster.sh
@@ -309,8 +309,6 @@ spec:
 "
 
 echo "Creating ToolchainCluster representation of ${JOINING_CLUSTER_TYPE} in ${CLUSTER_JOIN_TO}:"
-echo ${TOOLCHAINCLUSTER_CRD} | yq '.spec.caBundle="***"'
-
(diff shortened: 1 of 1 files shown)


scripts/add-cluster.sh
```diff
@@ -309,7 +309,7 @@ spec:
 "
 
 echo "Creating ToolchainCluster representation of ${JOINING_CLUSTER_TYPE} in ${CLUSTER_JOIN_TO}:"
-echo ${TOOLCHAINCLUSTER_CRD}
+echo ${TOOLCHAINCLUSTER_CRD} | yq '.spec.caBundle="***"'
```
MatousJobanek, line 313: I think that we don't need to print the content of the CR at all - I don't remember that I would need this information in the last years
```suggestion
```

MatousJobanek (approved):

scripts/add-cluster.sh
xcoulon: ok, so that's even simpler, then

PR #200, opened 2021-07-22
Create common toolchainconfig lib for host and reg service

rajivnathan:
Both the host operator and registration service will use the ToolchainConfig resource for configuration. The cache and configuration functions will all be part of toolchain-common and will be imported for use by both.

Summary of changes:
- Added cache for ToolchainConfig, similar to MemberOperatorConfig in member-operator repo
- Added the remaining configuration functions for all ToolchainConfig fields
- Added test helper functions for creating ToolchainConfig resources in unit and e2e tests

### DIFF
--- a/pkg/configuration/toolchainconfig/cache.go
+++ b/pkg/configuration/toolchainconfig/cache.go
@@ -1,0 +1,91 @@
+package toolchainconfig
+
+import (
+	"context"
+	"sync"
+
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
+	commonconfig "github.com/codeready-toolchain/toolchain-common/pkg/configuration"
+	errs "github.com/pkg/errors"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var configCache = &cache{}
+
+var cacheLog = logf.Log.WithName("cache_toolchainconfig")
+
+type cache struct {
+	sync.RWMutex
+	config  *toolchainv1alpha1.ToolchainConfig
+	secrets map[string]map[string]string // map of secret key-value pairs indexed by secret name
+}
+
+func (c *cache) set(config *toolchainv1alpha1.ToolchainConfig, secrets map[string]map[string]string) {
+	c.Lock()
+	defer c.Unlock()
+	c.config = config.DeepCopy()
+	c.secrets = commonconfig.CopyOf(secrets)
+}
+
+func (c *cache) get() (*toolchainv1alpha1.ToolchainConfig, map[string]map[string]string) {
+	c.RLock()
+	defer c.RUnlock()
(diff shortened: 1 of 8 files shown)


pkg/configuration/toolchainconfig/configuration.go
```diff
@@ -0,0 +1,373 @@
...
+	logger.Info("Toolchain configuration variables", "ToolchainConfigSpec", c.cfg)
+}
+
+func (c *ToolchainConfig) Environment() string {
+	return getString(c.cfg.Host.Environment, "prod")
+}
+
+func (c *ToolchainConfig) AutomaticApproval() AutoApprovalConfig {
+	return AutoApprovalConfig{c.cfg.Host.AutomaticApproval}
+}
```
MatousJobanek, line 43: just curious - do we really need/want to move all the methods to common? wouldn't it be easier to keep here some high-level abstraction and reuse it in host & registration service? I guess that none of the methods of the `ToolchainConfig` type is used in both host & reg-service. I that's true, then we can split it and create a new types in the repos with related functions. Cache and other stuff can stay here.

pkg/configuration/toolchainconfig/configuration.go
```diff
@@ -0,0 +1,373 @@
...
+	}
+	return defaultValue
+}
+
+func getString(value *string, defaultValue string) string {
+	if value != nil {
+		return *value
+	}
+	return defaultValue
+}
```
MatousJobanek, line 373: this is already in common if I'm not mistaken

pkg/configuration/toolchainconfig/configuration.go
rajivnathan: Yeah, you're right. I thought about this as well while making the registration service changes. The registration service only uses the `RegistrationService` section of the ToolchainConfig so let me do some refactoring to try to abstract out the cache and stuff.

pkg/configuration/toolchainconfig/configuration.go
rajivnathan: Yep, I had moved it to common when I did the member operator refactoring. I'll remove this too.

pkg/configuration/toolchainconfig/cache.go
```diff
@@ -0,0 +1,91 @@
...
+	defer c.RUnlock()
+	return c.config.DeepCopy(), commonconfig.CopyOf(c.secrets)
+}
+
+func updateConfig(config *toolchainv1alpha1.ToolchainConfig, secrets map[string]map[string]string) {
+	configCache.set(config, secrets)
+}
+
+func LoadLatest(cl client.Client) error {
+	namespace, err := commonconfig.GetWatchNamespace()
```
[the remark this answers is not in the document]
rajivnathan: I guess it should: https://github.com/codeready-toolchain/host-operator/blob/master/deploy/registration-service/registration-service.yaml#L118

But I'm almost at the point where I can try it to be sure. If it doesn't then I'll change it to accept namespace as a function parameter.

rajivnathan:
I'm going to close this PR for now. I need to make some substantial changes in order to abstract out the cache so there's no point in anyone else reviewing it for now. I'll reopen it after the refactoring.

PR #210, opened 2021-09-14
refactor: use `client.Object` instead of `meta.Accessor()`

xcoulon:
Fixes https://issues.redhat.com/browse/CRT-1182

see also: https://github.com/codeready-toolchain/member-operator/pull/297, https://github.com/codeready-toolchain/toolchain-e2e/pull/384 and https://github.com/codeready-toolchain/registration-service/pull/215

Signed-off-by: Xavier Coulon <xcoulon@redhat.com>

PR #211, opened 2021-09-21
refactor: remove 'client.ToolchainObject'

xcoulon:
this PR goes one step further than https://github.com/codeready-toolchain/toolchain-common/pull/210: here, the `client.ToolchainObject` wrapper is totally removed since with the latest SDK upgrade, the `client.Object` (from `sigs.k8s.io/controller-runtime/pkg/client`) has all the methods that we need.

see also: 
- https://github.com/codeready-toolchain/member-operator/pull/298 
- https://github.com/codeready-toolchain/toolchain-e2e/pull/385
- https://github.com/codeready-toolchain/registration-service/pull/216


xcoulon:
moving to WIP until the changes can be applied on `sandbox-sre` repo as well

pkg/client/object.go
```diff
@@ -0,0 +1,32 @@
+package client
+
+import (
+	"fmt"
+	"sort"
+
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type CompareToolchainObjects func(firstObject, secondObject runtimeclient.Object) (bool, error)
```
MatousJobanek, line 10: is this type still used anywhere? It doesn't seem so.

MatousJobanek (commented):
Nice cleanup :+1: 
Shouldn't be also renamed all the related functions/types so they don't contain the removed `ToolchainObject` type name? The same applies also to the comments.

pkg/client/object.go
xcoulon: > Nice cleanup 👍
> Shouldn't be also renamed all the related functions/types so they don't contain the removed `ToolchainObject` type name? The same applies also to the comments.

Actually, I did such renaming in sandbox-sre (PR coming soon), but I may have forgotten to do it in a few other places. Let me check that!

xcoulon: (looks like my response to your comment landed at the wrong location 😮 )

xcoulon: done in 573a24b

pkg/client/object.go
```diff
@@ -0,0 +1,38 @@
+package client
+
+import (
+	"fmt"
+	"sort"
+
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type CompareObjects func(firstObject, secondObject runtimeclient.Object) (bool, error)
```
MatousJobanek, line 10: So, is this really used somewhere? if so, then where? 
https://github.com/codeready-toolchain/toolchain-common/pull/211#discussion_r716450286

pkg/client/object.go
```diff
@@ -0,0 +1,38 @@
...
+	toolchainObjects := make([]runtimeclient.Object, len(objects))
+	for i, name := range names {
+		for _, object := range objects {
+			if fmt.Sprintf("%s,%s", object.GetNamespace(), object.GetName()) == name {
+				toolchainObjects[i] = object
+				break
+			}
+		}
+	}
+	return toolchainObjects
```
MatousJobanek, line 31:
```suggestion
	sortedObjects := make([]runtimeclient.Object, len(objects))
	for i, name := range names {
		for _, object := range objects {
			if fmt.Sprintf("%s,%s", object.GetNamespace(), object.GetName()) == name {
				sortedObjects[i] = object
				break
			}
		}
	}
	return sortedObjects
```

MatousJobanek (approved):
Thanks for renaming it :+1:

pkg/client/object.go
xcoulon: indeed, it's not (I forgot to check and delete it)

xcoulon: removed in 0676b94

PR #215, opened 2021-11-01
Add kube core api group to member client

alexeykazakov:
So the member cluster client (used by host) can get Secrets and Service Accounts. It's required for the proxy.

### DIFF
--- a/pkg/cluster/service.go
+++ b/pkg/cluster/service.go
@@ -84,6 +84,9 @@ func (s *ToolchainClusterService) addToolchainCluster(log logr.Logger, toolchain
 		if err := toolchainv1alpha1.AddToScheme(scheme); err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


rajivnathan (approved):

MatousJobanek (approved):
:+1: 
BTW, it looks like a dedicated-admin can list all SAs at the cluster scope, but sadly not the secrets. So as we discussed in the call, we would either need to use cluster-admin or create a new RoleBinding in the users' namespaces for that. 
Just FYI - the RoleBinding would let all SAs from the member-operator ns read the secrets - it could be added only to the AppStudio tier.

alexeykazakov:
> the RoleBinding would let all SAs from the member-operator ns read the secrets

Right! I didn't know we could bind all the SA from a specific namespace. It would be the best option IMO. I will add it to the AppStudio tier.

PR #217, opened 2021-11-19
feat: introduce cluster Config and a way to list them

MatousJobanek:
This PR introduces a `Config` type in cluster package - it more or less represents the configuration of ToolchainCluster. There is nothing new in there - just moving the params around.
The reason for that is that in host-operator we need to list all cluster configs for member clusters to be able to add it to Manager.

related PRs:
https://github.com/codeready-toolchain/host-operator/pull/542
https://github.com/codeready-toolchain/member-operator/pull/312
https://github.com/codeready-toolchain/toolchain-e2e/pull/412
https://github.com/codeready-toolchain/registration-service/pull/229

### DIFF
--- a/controllers/toolchaincluster/healthchecker.go
+++ b/controllers/toolchaincluster/healthchecker.go
@@ -68,7 +68,7 @@ func updateClusterStatuses(namespace string, cl client.Client) {
 			continue
 		}
 
-		clientSet, err := kubeclientset.NewForConfig(cachedCluster.Config.Config)
+		clientSet, err := kubeclientset.NewForConfig(cachedCluster.RestConfig)
 		if err != nil {
 			clusterLogger.Error(err, "cannot create ClientSet for a ToolchainCluster")
 			continue
--- a/controllers/toolchaincluster/healthchecker.go
+++ b/controllers/toolchaincluster/healthchecker.go
@@ -68,7 +68,7 @@ func updateClusterStatuses(namespace string, cl client.Client) {
 			continue
 		}
 
-		clientSet, err := kubeclientset.NewForConfig(cachedCluster.Config)
+		clientSet, err := kubeclientset.NewForConfig(cachedCluster.Config.Config)
 		if err != nil {
 			clusterLogger.Error(err, "cannot create ClientSet for a ToolchainCluster")
 			continue
--- a/pkg/cluster/service_test.go
+++ b/pkg/cluster/service_test.go
@@ -1,12 +1,19 @@
 package cluster_test
 
 import (
+	"context"
+	"fmt"
 	"testing"
+	"time"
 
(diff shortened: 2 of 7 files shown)


alexeykazakov (approved):

alexeykazakov (commented):
@MatousJobanek please update the reg-service too. It's using the toolchain cluster service from common too (the proxy thingy).

MatousJobanek:
> please update the reg-service too. It's using the toolchain cluster service from common too (the proxy thingy).

good point - I forgot about that:
https://github.com/codeready-toolchain/registration-service/pull/229

controllers/toolchaincluster/healthchecker.go
```diff
@@ -68,7 +68,7 @@ func updateClusterStatuses(namespace string, cl client.Client) {
 			continue
 		}
 
-		clientSet, err := kubeclientset.NewForConfig(cachedCluster.Config)
+		clientSet, err := kubeclientset.NewForConfig(cachedCluster.Config.Config)
```
rajivnathan, line 71: Feels a bit odd to have Config within Config, maybe rename the inner one to RestConfig so that when you see `Config` you're not wondering which Config?

rajivnathan (approved):
Looks good, just left a minor suggestion

controllers/toolchaincluster/healthchecker.go
MatousJobanek: this is a really good point :sweat_smile:  thanks ae1da04

pkg/cluster/service_test.go
```diff
@@ -58,3 +65,95 @@ func TestDeleteToolchainClusterWhenDoesNotExist(t *testing.T) {
...
+			HasName("west").
+			HasOperatorNamespace("toolchain-member-operator").
+			HasOwnerClusterName("m2ClusterName").
+			HasAPIEndpoint("http://m2.com").
+			RestConfigHasHost("http://m2.com")
+	})
+
+	t.Run("list host", func(t *testing.T) {
+		// when
+		clusterConfigs, err := cluster.ListToolchainClusterConfigs(cl, m1.Namespace, cluster.Host, time.Second)
```
xcoulon, line 104:
```suggestion
		clusterConfigs, err := cluster.ListToolchainClusterConfigs(cl, host.Namespace, cluster.Host, time.Second)
```

MatousJobanek: this code lists `ToolchainClusters` with the type `host` in the member namespace - in other words, this returns the list of host clusters registered for the particular member cluster. The assertion `HasOperatorNamespace` then verifies that the operator namespace of the found `ToolchianCluster` CR is really the expected host operator namespace.

xcoulon: ok

xcoulon (approved):

PR #218, opened 2021-11-24
test: MasterUserRecord with default TierName

xcoulon:

### DIFF
--- a/pkg/test/masteruserrecord/assertion.go
+++ b/pkg/test/masteruserrecord/assertion.go
@@ -138,13 +138,19 @@ func (a *Assertion) AllUserAccountsHaveCondition(expected toolchainv1alpha1.Cond
 	return a
 }
 
+func (a *Assertion) HasTier(tier toolchainv1alpha1.NSTemplateTier) *Assertion {
+	err := a.loadUaAssertion()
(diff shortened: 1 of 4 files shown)


pkg/test/masteruserrecord/assertion.go
```diff
@@ -144,6 +144,7 @@ func (a *Assertion) AllUserAccountsHaveTier(tier toolchainv1alpha1.NSTemplateTie
 	for _, ua := range a.masterUserRecord.Spec.UserAccounts {
 		a.userAccountHasTier(ua, tier)
 	}
+	assert.Equal(a.t, tier.Name, a.masterUserRecord.Spec.TierName)
```
MatousJobanek, line 147: wouldn't it be better to create a new function for this particular field? There are two reasons:
* it's not related to UserAccounts/NSTemplateSet CRDs
* we don't miss this verification code as soon as we drop NSTemplateSet from MUR

xcoulon: right, I can do

xcoulon: done in 1f17dba

MatousJobanek (approved):

PR #219, opened 2021-11-26
test: add NSTemplateSet Assertions

xcoulon:
Moved from github.com/codeready-toolchain/member-operator/test

also renaming other `assertions.go` files with a prefix based on pkg,
for the sake of consistency


MatousJobanek (approved):
> also renaming other assertions.go files with a prefix based on pkg,

I don't think that renaming the files is really needed. The context is contained in the package name - there is no need to replicate it in the file names if we don't have to deal with ambiguity. 

Apart from that - you renamed it to `masteruserrecord/master_user_record_assertion.go` but we don't use the `master_user_record` name format anywhere - shouldn't that have been `masteruserrecord_assertion.go`?
The same for `user_account_assertion.go` => `useraccount_assertion.go`
Otherwise, you would need to name the new file `ns_template_set_assertion.go` to keep the consistency, but I believe that we don't wan to use this format :-)

xcoulon:
> > also renaming other assertions.go files with a prefix based on pkg,
> 
> I don't think that renaming the files is really needed. The context is contained in the package name - there is no need to replicate it in the file names if we don't have to deal with ambiguity.
> 
> Apart from that - you renamed it to `masteruserrecord/master_user_record_assertion.go` but we don't use the `master_user_record` name format anywhere - shouldn't that have been `masteruserrecord_assertion.go`? The same for `user_account_assertion.go` => `useraccount_assertion.go` Otherwise, you would need to name the new file `ns_template_set_assertion.go` to keep the consistency, but I believe that we don't wan to use this format :-)

ah right, I probably got too excited with underscores 😂 
Note that I'm also fine with keeping the files named `assertions`, since there's one per package. It's just that I value consistency, despite the mistakes you pointed above 😜

xcoulon:
> > > also renaming other assertions.go files with a prefix based on pkg,
> > 
> > 
> > I don't think that renaming the files is really needed. The context is contained in the package name - there is no need to replicate it in the file names if we don't have to deal with ambiguity.
> > Apart from that - you renamed it to `masteruserrecord/master_user_record_assertion.go` but we don't use the `master_user_record` name format anywhere - shouldn't that have been `masteruserrecord_assertion.go`? The same for `user_account_assertion.go` => `useraccount_assertion.go` Otherwise, you would need to name the new file `ns_template_set_assertion.go` to keep the consistency, but I believe that we don't wan to use this format :-)
> 
> ah right, I probably got too excited with underscores 😂 Note that I'm also fine with keeping the files named `assertions`, since there's one per package. It's just that I value consistency, despite the mistakes you pointed above 😜

actually I stand corrected: we already do use this format in https://github.com/codeready-toolchain/toolchain-common/tree/master/pkg/test/masteruserrecord and https://github.com/codeready-toolchain/toolchain-common/tree/master/pkg/test/useraccount but I'm fine with also renaming the existing files to `pkg/test/masteruserrecord/masteruserrecord.go`, etc. to improve consistency. 
WDYT?

MatousJobanek:
Yeah please use the "less underscore format". Thanks

xcoulon:
> Yeah please use the "less underscore format". Thanks

done in ef0e108

PR #223, opened 2021-12-16
fix: don't watch changes of the ToolchainCluster.Status

MatousJobanek:
don't watch changes of the ToolchainCluster.Status so the cache doesn't recreate the client every time when health checker updates the status.


alexeykazakov (approved):

MatousJobanek:
That was a wrong assumption - we need to watch also the changes of the status, because we need to store the updated status in the cache :man_facepalming:

PR #224, opened 2021-12-17
fix: compare rest.Config and not cluster.Config

MatousJobanek:
This is the real fix for the weird behavior that I saw in our logs. We should compare the `rest.Config` when checking if the new client should be created.

related PRs: 
https://github.com/codeready-toolchain/member-operator/pull/320
https://github.com/codeready-toolchain/host-operator/pull/561


alexeykazakov (approved):
Good catch.

alexeykazakov:
It's also used in the reg-service. Could you please update it there too?

PR #229, opened 2022-01-10
test: refactor MasterUserRecord assertion

xcoulon:
- rename `Assertion` to `MasterUserRecordAssertion` for assertions on
a single MasterUserRecord
- rename `Assertion.masterUserRecord` to `MasterUserRecordAssertion.mur`
- introduce `MasterUserRecordsAssertion` for assertions on
multiple MasterUserRecords (counting, etc.)
- add multiple assertions methods to be used in `UserSignupController` tests.

changes to be used in https://github.com/codeready-toolchain/host-operator/pull/569

### DIFF
--- a/pkg/test/masteruserrecord/masteruserrecord_assertion.go
+++ b/pkg/test/masteruserrecord/masteruserrecord_assertion.go
@@ -336,10 +336,10 @@ type MasterUserRecordsAssertion struct {
 	t         test.T
 }
 
-func AssertThatMasterUserRecords(t test.T, namespace string, client client.Client) *MasterUserRecordsAssertion {
+func AssertThatMasterUserRecords(t test.T, client client.Client) *MasterUserRecordsAssertion {
 	return &MasterUserRecordsAssertion{
 		client:    client,
-		namespace: namespace,
+		namespace: test.HostOperatorNs,
 		t:         t,
 	}
 }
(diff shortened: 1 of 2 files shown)


pkg/test/masteruserrecord/masteruserrecord_assertion.go
```diff
@@ -201,50 +219,144 @@ func (a *Assertion) userAccountHasTier(ua toolchainv1alpha1.UserAccountEmbedded,
...
+func AssertThatMasterUserRecords(t test.T, namespace string, client client.Client) *MasterUserRecordsAssertion {
+	return &MasterUserRecordsAssertion{
+		client:    client,
+		namespace: namespace,
+		t:         t,
+	}
+}
+
+func (a *MasterUserRecordsAssertion) loadMasterUserRecords() error {
+	if a.client != nil {
```
MatousJobanek, line 348: why do we need this check? The client should be always set, right?

xcoulon: right, I'm not sure why I added that. Probably a quick copy/paste from https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/test/config/toolchainconfig_assertion.go#L22, but I will remove it

xcoulon: removed (both) in d93f97b

MatousJobanek (approved):

PR #233, opened 2022-01-27
chore: revert linter changes

xcoulon:
revert unneeded changes after gosec linter complains
use `// nolint:gosec` comment instead.

### DIFF
--- a/pkg/cluster/service.go
+++ b/pkg/cluster/service.go
@@ -128,9 +128,10 @@ func (s *ToolchainClusterService) refreshCache() {
 	if err := s.client.List(context.TODO(), toolchainClusters, &client.ListOptions{Namespace: s.namespace}); err != nil {
 		s.log.Error(err, "the cluster cache was not refreshed")
 	}
-	for _, cluster := range toolchainClusters.Items {
-		log := s.enrichLogger(&cluster)             // nolint:gosec
-		err := s.addToolchainCluster(log, &cluster) // nolint:gosec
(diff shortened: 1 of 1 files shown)


pkg/cluster/service.go
```diff
@@ -128,10 +128,9 @@ func (s *ToolchainClusterService) refreshCache() {
 	if err := s.client.List(context.TODO(), toolchainClusters, &client.ListOptions{Namespace: s.namespace}); err != nil {
 		s.log.Error(err, "the cluster cache was not refreshed")
 	}
-	for i := range toolchainClusters.Items {
-		cluster := toolchainClusters.Items[i] // avoids the "G601: Implicit memory aliasing in for loop. (gosec)" problem
-		log := s.enrichLogger(&cluster)
-		err := s.addToolchainCluster(log, &cluster)
+	for _, cluster := range toolchainClusters.Items {
+		log := s.enrichLogger(&cluster)             // nolint:gosec
+		err := s.addToolchainCluster(log, &cluster) // nolint:gosec
```
MatousJobanek, line 133: This is maybe the case when it would make sense to keep the format that was proposed by the linter. It's not really clear what the `s.addToolchainCluster` does.

xcoulon: yeah, that case may cause troubles indeed. I'll re-revert the change here

xcoulon: see 1d077a1

MatousJobanek (approved):

PR #242, opened 2022-02-24
Modified to comply with openshift naming

sbryzak:
This PR modifies the identity name compliance check to ensure the name is compliant with Openshift's naming standard, not the DNS-1123 standard as previous.

Fixes https://issues.redhat.com/browse/CRT-1449

### DIFF
--- a/pkg/identity/identity_naming_standard.go
+++ b/pkg/identity/identity_naming_standard.go
@@ -52,7 +52,7 @@ func (s *identityNamingStandard) IdentityName() string {
 //
 // The code at time of writing can be found at:
 //
-// https://github.com/openshift/oauth-server/blob/master/pkg/api/types.go#L108
+// https://github.com/openshift/oauth-server/blob/ef385cc3c9d90ee52f6db211ceb751e04ae967f5/pkg/api/types.go#L108
 //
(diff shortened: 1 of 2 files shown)


pkg/identity/identity_naming_standard.go
```diff
@@ -53,10 +47,15 @@ func (s *identityNamingStandard) IdentityName() string {
 	return fmt.Sprintf("%s:%s", s.provider, s.username())
 }
 
-// isIdentityNameCompliant returns true if the specified name is RFC-1123 compliant, otherwise it returns false
+// isIdentityNameCompliant returns true if the specified name is compliant with the Openshift identity naming standard,
+// encapsulated in the code found in the urlEncodeIfNecessary() function, otherwise it returns false.
+//
+// The code at time of writing can be found at:
+//
+// https://github.com/openshift/oauth-server/blob/master/pkg/api/types.go#L108
```
alexeykazakov, line 55: I would use a specific commit rather than master in case master is updated in the future:
```suggestion
// https://github.com/openshift/oauth-server/blob/ef385cc3c9d90ee52f6db211ceb751e04ae967f5/pkg/api/types.go#L108
```

alexeykazakov (approved):

PR #243, opened 2022-02-28
feat: update k8s to v0.22.7 and OpenShift to 4.9

MatousJobanek:
Update k8s and OpenShift dependencies so they match the version of OpenShift clusters we run. It also corresponds to the version that is generated by operator-sdk 1.16.0.

It slightly modifies the logic of the test client so it sets some default values that are automatically set by FakeClient

related pr:
https://github.com/codeready-toolchain/api/pull/295


pkg/test/client.go
```diff
@@ -129,6 +132,14 @@ func Update(ctx context.Context, cl *FakeClient, obj client.Object, opts ...clie
 	currentMap["status"] = nil
 	currentMap["kind"] = nil
 	currentMap["apiVersion"] = nil
+	if currentMap["spec"] == nil {
+		currentMap["spec"] = map[string]interface{}{}
+	}
+	for key, value := range currentMap {
+		if _, exist := updatingMap[key]; !exist && value == nil {
+			updatingMap[key] = nil
```
xcoulon, line 140: why not doing `delete(updateMap, key)`?

MatousJobanek: I would need to delete it from the `currentMap` (and not `updateMap` as you suggested) because that is the one that contains the nil value. However, since I would need to check the content of the `updatingMap` anyway, then there wouldn't be any room for making the code simpler.

xcoulon: ok

xcoulon (approved):

go.mod
```diff
@@ -4,34 +4,26 @@ require (
...
-	k8s.io/klog v1.0.0 // indirect
-	k8s.io/kubectl v0.20.2 // indirect
-	sigs.k8s.io/controller-runtime v0.8.3
+	k8s.io/api v0.22.7
+	k8s.io/apimachinery v0.22.7
+	k8s.io/client-go v0.22.7
+	sigs.k8s.io/controller-runtime v0.10.3
 )
 
+replace github.com/codeready-toolchain/api => github.com/matousjobanek/api v0.0.0-20220228101619-30eaab5220b7
```
alexeykazakov, line 27: Reminder.

alexeykazakov (approved):

go.mod
MatousJobanek: done [f0f2ff4](https://github.com/codeready-toolchain/toolchain-common/pull/243/commits/f0f2ff461eaaf6db16febaa5c86ea7d923d8c866)

PR #251, opened 2022-04-18
move notification builder from host to common

ranakan19:
Move notification Builder from host to toolchain-common

changes wrt - https://issues.redhat.com/browse/CRT-1418

### DIFF
--- a/pkg/notification/notification_builder_test.go
+++ b/pkg/notification/notification_builder_test.go
@@ -6,7 +6,7 @@ import (
 	"github.com/gofrs/uuid"
 
 	"github.com/codeready-toolchain/api/api/v1alpha1"
-	test2 "github.com/codeready-toolchain/toolchain-common/pkg/test/usersignup"
+	testusersignup "github.com/codeready-toolchain/toolchain-common/pkg/test/usersignup"
 
 	"github.com/codeready-toolchain/toolchain-common/pkg/test"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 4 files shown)


pkg/usersignup/usersignup.go
```diff
@@ -1,6 +1,12 @@
 package usersignup
 
 import (
+	"crypto/md5" // nolint:gosec
```
alexeykazakov, line 4: This change in `usersignup.go` doesn't seem to be related to the builder.

ranakan19: the notification_builder_test uses the changes in usersignup.go

alexeykazakov: OK. Then I guess you don't need to keep this code in the host operator anymore and can switch to common here: https://github.com/codeready-toolchain/host-operator/pull/629/files

ranakan19: done

alexeykazakov (approved):

MatousJobanek (approved):

pkg/notification/notification_builder_test.go
```diff
@@ -0,0 +1,142 @@
+package notification
+
+import (
+	"testing"
+
+	"github.com/gofrs/uuid"
+
+	"github.com/codeready-toolchain/api/api/v1alpha1"
+	test2 "github.com/codeready-toolchain/toolchain-common/pkg/test/usersignup"
```
rajivnathan, line 9: Probably better to use an alias that's related to usersignup for this. eg. `testusersignup`

rajivnathan (approved):
Looks good, just left 1 minor comment

PR #255, opened 2022-04-28
test: SpaceRoles in NSTemplateSets

xcoulon:

### DIFF
--- a/pkg/test/nstemplateset/nstemplateset.go
+++ b/pkg/test/nstemplateset/nstemplateset.go
@@ -16,9 +16,37 @@ const (
 	codeTemplateRef             = "basic-code-abcde21"
 )
 
+func NewNSTemplateSet(name string, options ...Option) *toolchainv1alpha1.NSTemplateSet {
+	nstmplSet := &toolchainv1alpha1.NSTemplateSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: test.MemberOperatorNs,
+			Name:      name,
+		},
+		Spec: toolchainv1alpha1.NSTemplateSetSpec{
+			TierName: "basic",
+			ClusterResources: &toolchainv1alpha1.NSTemplateSetClusterResources{
+				TemplateRef: clusterResourcesTemplateRef,
+			},
+			Namespaces: []toolchainv1alpha1.NSTemplateSetNamespace{
+				{
+					TemplateRef: devTemplateRef,
+				},
+				{
+					TemplateRef: codeTemplateRef,
+				},
+			},
+		},
+		Status: toolchainv1alpha1.NSTemplateSetStatus{},
+	}
+	for _, apply := range options {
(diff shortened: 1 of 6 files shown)


pkg/test/nstemplateset/nstemplateset_assertion_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+		t.Run("fails when actual user is not expected", func(t *testing.T) {
+			// given
+			mockT := test.NewMockT()
+			// when
+			nstemplateset.AssertThatNSTemplateSet(mockT, test.MemberOperatorNs, "foo", cl).HasSpaceRoles(
+				nstemplateset.SpaceRole("base-admin-123456b", "admin1"), // admin2 is not expected
+				nstemplateset.SpaceRole("base-viewer-123456b", "viewer1"),
+			)
+			// then: all good
```
rajivnathan, line 65: I guess this comment can be removed for all the error scenarios?

xcoulon: may I ask why? the "all good" comment is to explain that no error or failure should have been reported by the assertion

pkg/test/nstemplateset/nstemplateset.go
```diff
@@ -36,6 +64,31 @@ func WithReferencesFor(nstemplateTier *toolchainv1alpha1.NSTemplateTier) Option
...
+			apply(nstemplateTier, nstmplSet)
+		}
+	}
+}
+
+type TierOption func(*toolchainv1alpha1.NSTemplateTier, *toolchainv1alpha1.NSTemplateSet)
+
+func WithSpaceRole(role, username string) TierOption {
+	return func(nstemplateTier *toolchainv1alpha1.NSTemplateTier, nstmplSet *toolchainv1alpha1.NSTemplateSet) {
+		if tierSpaceRole, found := nstemplateTier.Spec.SpaceRoles[role]; found {
```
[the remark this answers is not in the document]
rajivnathan: If the role does not exist then this function is a no op so I don't think it makes sense to provide a role that would not be found?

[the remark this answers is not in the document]
xcoulon: yes, that's right. Which is why I said that the caller should be sure that the role exists. But maybe we can improve this func to make it less error-prone, indeed

pkg/test/nstemplateset/nstemplateset_assertion_test.go
xcoulon: ahh, I see. I did not realize you were pointing to a case where an error is expected! 🤦‍♂️ 
let me fix that in a subsequent PR :)

xcoulon: see https://github.com/codeready-toolchain/toolchain-common/pull/259

PR #261, opened 2022-06-14
test: SocialEvent helper func

xcoulon:
see https://github.com/codeready-toolchain/registration-service/pull/267

### DIFF
--- a/pkg/test/usersignup/usersignup.go
+++ b/pkg/test/usersignup/usersignup.go
@@ -3,6 +3,9 @@ package usersignup
 import (
 	"crypto/md5" // nolint:gosec
 	"encoding/hex"
+	"strconv"
+	"time"
+
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
 	"github.com/codeready-toolchain/toolchain-common/pkg/condition"
 	"github.com/codeready-toolchain/toolchain-common/pkg/states"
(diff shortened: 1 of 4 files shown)


pkg/test/usersignup/usersignup.go
```diff
@@ -137,6 +139,10 @@ func WithActivations(value string) Modifier {
 	return WithAnnotation(toolchainv1alpha1.UserSignupActivationCounterAnnotationKey, value)
 }
 
+func WithVerificationAttempts(value int) Modifier {
+	return WithAnnotation(toolchainv1alpha1.UserVerificationAttemptsAnnotationKey, strconv.Itoa(value))
```
rajivnathan, line 143: Is `UserVerificationAttemptsAnnotationKey` clear enough or could it be confused between activation code and phone code verification?

rajivnathan (approved):
Looks good, just the one minor question

pkg/test/usersignup/usersignup.go
xcoulon: I believe we could use the same key, with a max attempts based on any combination of attempts with a phone code or an activation code. I mean, I believe it's sufficient to give users a max number of attempts, whatever mechanism they choose to use. And besides, it does not impact the logic on the back-end ;)

I hope this makes sense.

rajivnathan: I think that's fine since I guess it should be obvious that this UserSignup used an activation code instead of a phone verification code because it would have the activation label

PR #262, opened 2022-06-21
feat: new SocialEvent name generator

xcoulon:
generates unique random names in the `XXXX-XXXX-XXXX-XXXX` format,
using the `abcdefghjklmnpqrstuvwxyz` letters and `23456789` figures,
making it a valid resource name which can also be used as an
activation code.

### DIFF
--- a/pkg/socialevent/socialevent_name_generator.go
+++ b/pkg/socialevent/socialevent_name_generator.go
@@ -1,0 +1,30 @@
+package socialevent
+
+import (
+	"crypto/rand"
+	"math/big"
+	"strings"
+)
+
+// NewName SocialEvent names (a.k.a, activation codes) are generated in a standard format,
+// 4 blocks of 4 case insensitive, alphanumeric characters separated by a minus sign delimiter, like so:
+// XXXX-XXXX-XXXX-XXXX
+// In order to minimise entry errors, a limited character set will be used with visually ambiguous characters excluded:
+// Letters:		abcdefghjklmnpqrstuvwxyz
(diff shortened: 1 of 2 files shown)


pkg/socialevent/socialevent_name_generator.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"crypto/rand"
+	"math/big"
+	"strings"
+)
+
+// NewName SocialEvent names (a.k.a, activation codes) are generated in a standard format,
+// 4 blocks of 4 case insensitive, alphanumeric characters separated by a minus sign delimiter, like so:
+// XXXX-XXXX-XXXX-XXXX
```
alexeykazakov, line 11: This looks like an overkill :)
Let's make it a bit more user friendly.
For example use digest only and limit the number of digests to 6-8. Extra security is good but we can address it differently/later.

xcoulon: addressed in f8390ff

MatousJobanek (approved):
Looks good :+1: 
We could also make it a bit simpler - just generate UUID and take the first part of it. But it doesn't matter how it is generated if it meets the requirements :+1:

xcoulon:
> Looks good 👍 We could also make it a bit simpler - just generate UUID and take the first part of it. But it doesn't matter how it is generated if it meets the requirements 👍

the design doc specified that we should use a subset of characters, which is why I implemented the generator this way 🤷‍♂️ 
but it's something we can definitely refactor later if needed.

PR #268, opened 2022-09-23
utility func to hash

xcoulon:
make it generic since the logic to hash with md5 is used in multiple places

### DIFF
--- a/pkg/test/usersignup/usersignup.go
+++ b/pkg/test/usersignup/usersignup.go
@@ -1,13 +1,14 @@
 package usersignup
 
 import (
-	"crypto/md5" // nolint:gosec
-	"encoding/hex"
+	// nolint:gosec
+
 	"strconv"
 	"time"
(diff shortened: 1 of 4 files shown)


xcoulon:
pfff, opened from `master` 🤦‍♂️

pkg/test/usersignup/usersignup.go
```diff
@@ -1,13 +1,14 @@
 package usersignup
 
 import (
-	"crypto/md5" // nolint:gosec
-	"encoding/hex"
+	// nolint:gosec
```
mfrancisc, line 4: the `nolint` comment may not be needed anymore, since the import was moved.

xcoulon: yes, I noticed it too, and fixed it in https://github.com/codeready-toolchain/toolchain-common/pull/269

PR #277, opened 2022-11-07
refactor: usersignup approved funcs

xcoulon:
distinguish between automatic approval and manual approval (by admin)

see also:
- https://github.com/codeready-toolchain/member-operator/pull/389
- https://github.com/codeready-toolchain/host-operator/pull/702
- https://github.com/codeready-toolchain/registration-service/pull/286
- https://github.com/codeready-toolchain/sandbox-sre/pull/810
- https://github.com/codeready-toolchain/toolchain-e2e/pull/620

### DIFF
--- a/pkg/states/state_manager.go
+++ b/pkg/states/state_manager.go
@@ -2,30 +2,18 @@ package states
 
 import toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
 
-func ManuallyApproved(userSignup *toolchainv1alpha1.UserSignup) bool {
+func ApprovedManually(userSignup *toolchainv1alpha1.UserSignup) bool {
 	return contains(userSignup.Spec.States, toolchainv1alpha1.UserSignupStateApproved)
 }
 
-// SetManuallyApproved when a user was manually approved by an admin
+// SetApprovedManually when a user was manually approved by an admin
 // Note: when a user is manually approved, the `UserSignup.Spec.States` contains a single entry: `approved`
-func SetManuallyApproved(userSignup *toolchainv1alpha1.UserSignup, approved bool) {
+func SetApprovedManually(userSignup *toolchainv1alpha1.UserSignup, approved bool) {
 	setState(userSignup, toolchainv1alpha1.UserSignupStateApproved, approved)
 	if approved {
 		setState(userSignup, toolchainv1alpha1.UserSignupStateVerificationRequired, false)
 		setState(userSignup, toolchainv1alpha1.UserSignupStateDeactivating, false)
 		setState(userSignup, toolchainv1alpha1.UserSignupStateDeactivated, false)
-	}
-}
-
-func AutomaticallyApproved(userSignup *toolchainv1alpha1.UserSignup) bool {
-	return userSignup.Spec.States == nil
-}
-
-// SetAutomaticallyApproved when a user was automatically approved (verification passed, etc.)
-// Note: when a user is automatically approved, the `UserSignup.Spec.States` is reset
-func SetAutomaticallyApproved(userSignup *toolchainv1alpha1.UserSignup, approved bool) {
-	if approved {
-		userSignup.Spec.States = nil
 	}
 }
 
--- a/pkg/states/state_manager.go
+++ b/pkg/states/state_manager.go
@@ -2,17 +2,30 @@ package states
 
 import toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
 
-func Approved(userSignup *toolchainv1alpha1.UserSignup) bool {
+func ManuallyApproved(userSignup *toolchainv1alpha1.UserSignup) bool {
 	return contains(userSignup.Spec.States, toolchainv1alpha1.UserSignupStateApproved)
 }
 
-func SetApproved(userSignup *toolchainv1alpha1.UserSignup, val bool) {
-	setState(userSignup, toolchainv1alpha1.UserSignupStateApproved, val)
-
-	if val {
+// SetManuallyApproved when a user was manually approved by an admin
+// Note: when a user is manually approved, the `UserSignup.Spec.States` contains a single entry: `approved`
+func SetManuallyApproved(userSignup *toolchainv1alpha1.UserSignup, approved bool) {
+	setState(userSignup, toolchainv1alpha1.UserSignupStateApproved, approved)
+	if approved {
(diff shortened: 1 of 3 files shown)


pkg/states/state_manager.go
```diff
@@ -2,20 +2,33 @@ package states
...
 	}
 }
 
+func AutomaticallyApproved(userSignup *toolchainv1alpha1.UserSignup) bool {
+	return userSignup.Spec.States == nil
+}
+
+// SetAutomaticallyApproved when a user was automatically approved (verification passed, etc.)
+// Note: when a user is automatically approved, the `UserSignup.Spec.States` is reset
+func SetAutomaticallyApproved(userSignup *toolchainv1alpha1.UserSignup, approved bool) {
```
MatousJobanek, line 26: the `approved` boolean doesn't make much sense here - if it's false, then it doesn't do anything. Wouldn't make sense to just drop the parameter?

MatousJobanek, line 26: is there any place in production where any of these two functions would be called? I'm not aware of any.

pkg/test/usersignup/usersignup.go
```diff
@@ -26,20 +26,50 @@ func WithOriginalSub(originalSub string) Modifier {
...
+func ManuallyApproved() Modifier {
 	return func(userSignup *toolchainv1alpha1.UserSignup) {
-		states.SetApproved(userSignup, true)
+		states.SetManuallyApproved(userSignup, true)
 	}
 }
 
-func ApprovedAutomatically(before time.Duration) Modifier {
+// ManuallyApproved sets the UserSignup state to `approved` and adds a status condition
+func ManuallyApprovedAgo(before time.Duration) Modifier {
```
MatousJobanek, line 37: why not `ManuallyApprovedBefore`

pkg/test/usersignup/usersignup.go
```diff
@@ -26,20 +26,50 @@ func WithOriginalSub(originalSub string) Modifier {
...
+func AutomaticallyApproved() Modifier {
+	return func(userSignup *toolchainv1alpha1.UserSignup) {
+		states.SetAutomaticallyApproved(userSignup, true)
+		userSignup.Status.Conditions = condition.AddStatusConditions(userSignup.Status.Conditions,
+			toolchainv1alpha1.Condition{
+				Type:               toolchainv1alpha1.UserSignupApproved,
+				Status:             corev1.ConditionTrue,
+				Reason:             toolchainv1alpha1.UserSignupApprovedAutomaticallyReason,
+				LastTransitionTime: metav1.Time{Time: time.Now()},
+			})
```
MatousJobanek, line 60: my understanding is that in this function we should not set the condition.

MatousJobanek (commented):
I agree with @sbryzak that `ApprovedAutomatically` and `ApprovedManually` sounds better

pkg/states/state_manager.go
sbryzak: I'm in agreement with Matous here.. plus automatic approval is a configuration concern, and i feel it has no place here.

sbryzak (changes requested):
I think this set of changes has introduced a discrepancy... as I mentioned in my comment, automatic approval is purely a configuration-based feature.  To demonstrate the discrepancy, let's say for example we call `SetAutomaticallyApproved` while the automatic approval configuration option is disabled - does this then approve the user?  In this case it would have no effect which means the function is not doing what is described.

I think we should remove anything relating to automatic approval from the state manager altogether, in my opinion it doesn't belong here.

xcoulon:
> I think this set of changes has introduced a discrepancy... as I mentioned in my comment, automatic approval is purely a configuration-based feature. To demonstrate the discrepancy, let's say for example we call `SetAutomaticallyApproved` while the automatic approval configuration option is disabled - does this then approve the user? In this case it would have no effect which means the function is not doing what is described.
> 
> I think we should remove anything relating to automatic approval from the state manager altogether, in my opinion it doesn't belong here.

Fair enough, but frankly, having `spec.states=[approved]` when a user was manually approved vs `spec.states=nil` when the user was automatically approved is very confusing at first!

pkg/test/usersignup/usersignup.go
xcoulon: because the arg is a duration, not a time

pkg/states/state_manager.go
xcoulon: I removed this func in 5af9225

xcoulon: I remove this func in 5af9225

xcoulon:
> I agree with @sbryzak that `ApprovedAutomatically` and `ApprovedManually` sounds better

I removed the `AutomaticallyApproved` funcs and renamed the other ones in 5af9225

MatousJobanek (approved):

sbryzak (approved):
Looks much better, thanks!

pkg/test/usersignup/usersignup.go
```diff
@@ -26,20 +26,22 @@ func WithOriginalSub(originalSub string) Modifier {
...
+func ApprovedManually() Modifier {
 	return func(userSignup *toolchainv1alpha1.UserSignup) {
-		states.SetApproved(userSignup, true)
+		states.SetApprovedManually(userSignup, true)
 	}
 }
 
-func ApprovedAutomatically(before time.Duration) Modifier {
+// ApprovedManuallyAgo sets the UserSignup state to `approved` and adds a status condition
+func ApprovedManuallyAgo(before time.Duration) Modifier {
```
ranakan19, line 37: Maybe I'm missing something, but what does "ago" in the function name imply? It gave me an impression that it is adding a condition to already approved usersignups, but that isn't confirmed anywhere.

ranakan19 (commented):
Looks good, left a question

pkg/test/usersignup/usersignup.go
rajivnathan: I guess it's because it accepts the `before` parameter and that will be reflected in the condition `LastTransitionTime`. Another option for the func name would be something like `ApprovedManuallyEarlier` but not sure it's much better.

rajivnathan (approved):

pkg/test/usersignup/usersignup.go
xcoulon: yes, the `Ago` suffix in this func's name refers to the `duration` arg, which is used to set the `LastTransitionTime` at `now - duration`, ie, somewhere in the past.

PR #279, opened 2022-11-22
Add CapacityThresholds test helpers

mfrancisc:
Add CapacityThresholds helper functions.

Jira Ticket: [ASC-114](https://issues.redhat.com/browse/ASC-114)

changes to be used in:

- [sandbox-sre#828](https://github.com/codeready-toolchain/sandbox-sre/pull/828)
- [host-operator#709](https://github.com/codeready-toolchain/host-operator/pull/709)


pkg/test/config/toolchainconfig.go
```diff
@@ -107,6 +107,38 @@ func (o AutomaticApprovalOption) MaxNumberOfUsers(overall int, perMember ...PerM
...
+}
+
+func CapacityThresholds() *CapacityThresholdsOption {
+	c := &CapacityThresholdsOption{
+		ToolchainConfigOptionImpl: &ToolchainConfigOptionImpl{},
+	}
+	return c
+}
+
+func (c CapacityThresholdsOption) ResourceCapacityThreshold(defaultThreshold int, perMember ...PerMemberClusterOptionInt) CapacityThresholdsOption {
```
xcoulon, line 121: shouldn't it be `(c *CapacityThresholdsOption)`, since the method calls `c.addFunction(..)` which is defined as:
```
func (option *ToolchainConfigOptionImpl) addFunction(funcToAdd ToolchainConfigOptionFunc)
```

mfrancisc: unit tests are passing, so I think it will work either way.

I see that `ToolchainConfigOptionImpl` has pointer receiver for `Apply` and `addFunction` which makes sense since:
- `addFunction` it's appending functions to `option.toApply` function.
- `Apply` is applying all the functions from `option.toApply`

about `ResourceCapacityThreshold` and `MaxNumberOfSpaces` I'm not sure what will be the reasoning behind it, since they are just wrappers and not modifying the _receiver_ .
could you elaborate more? 
Is because of smaller memory footprint or readability (but we will potentially loose _immutability_ ) ?

xcoulon: I'm sure your code is valid, but it's a matter of readability indeed. It seems more logical to me that we manipulate a `*CapacityThresholdsOption` since we're modifying its content when calling its methods.

mfrancisc: ok thanks for the feedback! 
Could you also explain more about why would be better to have  _no pointer to ToolchainConfigOptionImpl_ in this case. I'm referring to this part of code from your suggestion:
```
// no pointer to ToolchainConfigOptionImpl
type CapacityThresholdsOption struct {
  ToolchainConfigOptionImpl 
}
```

MatousJobanek (approved):

pkg/test/config/toolchainconfig.go
mfrancisc: Ok got it!

Maybe we can also consider a different naming ( same separate refactor PR maybe):

from `CapacityThresholds.ResourceCapacityThreshold()` to something like `CapacityThresholds.Resources()` or something more tidy. This also means renaming `ResourceCapacityThreshold` to just `Resources`. 
If makes sense for everyone.

mfrancisc: @xcoulon are you ok with addressing this refactor in a separate PR ?

xcoulon: sure, let's change that in a separate PR so this doesn't block this one.

xcoulon (approved):

PR #284, opened 2023-01-04
Roles as labels at ToolchainCluster CRD

mfrancisc:
This PR adds the `home` label to toolchaincluster of type member.

See Jira: https://issues.redhat.com/browse/ASC-245

Related PRs:

- host-operator: https://github.com/codeready-toolchain/host-operator/pull/722
- member-operator: https://github.com/codeready-toolchain/member-operator/pull/403
- toolchain-e2e: https://github.com/codeready-toolchain/toolchain-e2e/pull/645
- toolchain-cicd: https://github.com/codeready-toolchain/toolchain-cicd/pull/79

PR #285, opened 2023-01-10
comment out unneeded logs

xcoulon:


mfrancisc:
I think the verbosity of the logger cannot be configured right ?
So that for example, in dev and staging we keep them more verbose while production by default they log less stuff.

I found those lines and others that I've added here in this [PR](https://github.com/codeready-toolchain/toolchain-common/pull/284/files) useful while developing locally.

MatousJobanek (approved):

mfrancisc (commented):
Looks good to me! 

Just small comment - I would remove commented lines.

mfrancisc (approved):

PR #309, opened 2023-04-18
Shorten compliant username

ranakan19:
fix wrt https://issues.redhat.com/browse/CRT-1670

shorten the length of the username to maxLength of 20 characters to make the namespace a total of 30 characters such the resulting route with the namespace is DNS compliant.


pkg/usersignup/usersignup.go
```diff
@@ -30,5 +34,12 @@ func TransformUsername(username string) string {
 	if strings.HasSuffix(newUsername, "-") {
 		newUsername = newUsername + "crt"
 	}
+	if len(newUsername) > maxLength {
+		newUsername = newUsername[0:maxLength]
+		if !dnsRegexp.MatchString(newUsername) {
+			// trim starting or ending hyphen
+			newUsername = strings.Trim(newUsername, "-")
+		}
+	}
```
mfrancisc, line 43: maybe you already saw this, if it can help there is a validation function in apimachinery for subdomain names: https://github.com/kubernetes/apimachinery/blob/8df78a92da82be9a15f1a3beaa1d7f22acae3f3d/pkg/api/validation/generic.go#L37-L42

alexeykazakov (approved):
Looks good but please see @mfrancisc's comment.

pkg/usersignup/usersignup_test.go
```diff
@@ -31,6 +31,12 @@ func TestTransformUsername(t *testing.T) {
...
 	assertName(t, "crt-12345", "12345")
+	assertName(t, "thisisabout20charact", "thisisabout20characters@email.com")
+	assertName(t, "isexactly20character", "isexactly20character@email.com")
+	assertName(t, "isexactly20character", "isexactly20character")
+	assertName(t, "isexactly21characte", "isexactly21characte-r") // shortened 20 character will end in hyphen
+	assertName(t, "isexactly20charactr", "isexactly20charactr-")  // but ending in hyphen
+	assertName(t, "thisis19characters-c", "thisis19characters-")  // suffix -crt is added before truncating string
 }
 
 var dnsRegExp = "^[a-z0-9]([-a-z0-9]*[a-z0-9])?$"
```
rajivnathan, line 42: I guess we can change this according to @mfrancisc comment too?

pkg/usersignup/usersignup.go
```diff
@@ -30,5 +35,12 @@ func TransformUsername(username string) string {
 	if strings.HasSuffix(newUsername, "-") {
 		newUsername = newUsername + "crt"
 	}
+	if len(newUsername) > MaxLength {
+		newUsername = newUsername[0:MaxLength]
+		if !dnsRegexp.MatchString(newUsername) {
+			// trim starting or ending hyphen
+			newUsername = strings.Trim(newUsername, "-")
```
rajivnathan, line 42: Do we know why we append `crt` as a prefix on line 33 and on a suffix on line 36? Maybe it should just be trimmed instead of appending `crt`?

Btw, I'm not suggesting we remove the trim here if we do that. I recognize that we'd still need it here since we shorten it to 20 chars and the last character could end up being `-`. So it's more a question of whether that `crt` stuff still has value or whether the reason for doing that would have an impact on what we should do here with regard to trimming the `-`.

rajivnathan (approved):
Nice, looks good. Just a couple questions

pkg/usersignup/usersignup.go
alexeykazakov: crt doesn't have any value on its own as far as I remember. So, I guess trimming would be fine.

sbryzak (approved):

pkg/usersignup/usersignup_test.go
```diff
@@ -31,6 +31,12 @@ func TestTransformUsername(t *testing.T) {
 	assertName(t, "crt-me-crt", "_me_")
 	assertName(t, "crt-me-crt", "-me-")
 	assertName(t, "crt-12345", "12345")
+	assertName(t, "thisisabout20charact", "thisisabout20characters@email.com")
+	assertName(t, "isexactly20character", "isexactly20character@email.com")
+	assertName(t, "isexactly20character", "isexactly20character")
+	assertName(t, "isexactly21characte", "isexactly21characte-r") // shortened 20 character will end in hyphen
```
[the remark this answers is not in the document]
ranakan19: I meant to imply that the truncating on char length limit would have resulted in ending in an hyphen, but logic trims for "-", so the result would not have the trailing hyphen. I was describing the test case more than the result case. Let me add the entire scenario in the comment.

pkg/usersignup/usersignup.go
ranakan19: Thank you for sharing this, looked into it and i think we can directly use the [`IsDNS1123Label`](https://github.com/kubernetes/apimachinery/blob/8df78a92da82be9a15f1a3beaa1d7f22acae3f3d/pkg/util/validation/validation.go#L188) instead for regex.
will still have to use custom length since the function uses maxLength as 63 - which is the maxLimit of the DNSLabel.

ranakan19: @mfrancisc done in [9a9de94](https://github.com/codeready-toolchain/toolchain-common/pull/309/commits/9a9de94456dbb022f7aa75a6dec18f0689fe2171)

pkg/usersignup/usersignup_test.go
ranakan19: done in [9a9de94](https://github.com/codeready-toolchain/toolchain-common/pull/309/commits/9a9de94456dbb022f7aa75a6dec18f0689fe2171)

mfrancisc (approved):
Nice Job! 👍 

Thanks for checking my comments.

pkg/usersignup/usersignup_test.go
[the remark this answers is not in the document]
ranakan19: updated

alexeykazakov (approved):

pkg/usersignup/usersignup.go
```diff
@@ -30,5 +37,40 @@ func TransformUsername(username string) string {
...
+			break
+		}
+	}
+
+	// Check for any forbidden suffixes
+	for _, suffix := range ForbiddenUsernameSuffixes {
+		if strings.HasSuffix(newUsername, suffix) {
+			if len(newUsername) > maxlengthWithPrefix {
+				// replace prefix instead of append
+				newUsername = newUsername[:len(newUsername)-4] + "-crt"
```
MatousJobanek, line 68: just out of curiosity - could it be also:
```suggestion
				newUsername = newUsername[:maxlengthWithPrefix] + "-crt"
```
so we would reuse the variable with the expected length that is already set and used before?

MatousJobanek (approved):
Cool :+1:

PR #310, opened 2023-04-27
feat: Implement "copy secret" functionality in SpaceRequest controller

mfrancisc:
Add `CreateTokenRequest` and testing utilities.

Jira: https://issues.redhat.com/browse/ASC-258

### DIFF
--- a/pkg/client/rest_client.go
+++ b/pkg/client/rest_client.go
@@ -26,9 +26,7 @@ func CreateTokenRequest(restClient *rest.RESTClient, namespacedName types.Namesp
 		Into(result); err != nil {
 		return "", err
 	}
-	if result == nil {
-		return "", fmt.Errorf("unable to create token, got nil response")
-	}
+
 	if len(result.Status.Token) == 0 {
 		return "", fmt.Errorf("unable to create token, got empty string")
 	}
--- a/pkg/client/rest_client.go
+++ b/pkg/client/rest_client.go
@@ -25,5 +25,9 @@ func CreateTokenRequest(restClient *rest.RESTClient, namespacedName types.Namesp
 		Into(result); err != nil {
 		return "", err
 	}
+	if result == nil {
+		return "", fmt.Errorf("unable to create token, got nil response")
+	}
+
 	return result.Status.Token, nil
 }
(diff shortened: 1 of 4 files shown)


pkg/client/rest_client.go
```diff
@@ -0,0 +1,29 @@
...
+	"context"
+	"fmt"
+
+	authv1 "k8s.io/api/authentication/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/rest"
+	"k8s.io/utils/pointer"
+)
+
+// CreateTokenRequest creates a TokenRequest for a service account using given expiration in seconds.
```
alexeykazakov, line 13: Please also indicate that the function returns the generated token.

pkg/client/rest_client_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+		// when
+		require.NoError(t, err)
+		token, err := restclient.CreateTokenRequest(cl, types.NamespacedName{
+			Namespace: "jane-env",
+			Name:      "jane",
+		}, 1)
+
+		// then
+		require.Error(t, err)      // an error should be returned
```
alexeykazakov, line 52: Let's also check the error message, to make sure it's the actual error we expect here and not something else.

alexeykazakov (approved):
Looks good. Just a couple of minor suggestions.

pkg/client/rest_client.go
mfrancisc: fixed the comment in https://github.com/codeready-toolchain/toolchain-common/pull/310/commits/140156e24887adc4521460abff285b48166fd1da

pkg/client/rest_client_test.go
mfrancisc: Actually seems that this can never happen , see also linter issue:
```
pkg/client/rest_client.go:29:5: SA4031: this nil check is never true (staticcheck)
	if result == nil {
	   ^
```
removed that check https://github.com/codeready-toolchain/toolchain-common/pull/310/commits/55fc7e763c742def67bb7bc1f0ae10de667db077

PR #311, opened 2023-05-03
Add signing method param for tokens

rajivnathan:


alexeykazakov (approved):

pkg/test/auth/tokenmanager.go
```diff
@@ -244,8 +244,8 @@ func (c *MyClaims) Valid() error {
 }
 
 // GenerateToken generates a default token.
-func (tg *TokenManager) GenerateToken(identity Identity, kid string, extraClaims ...ExtraClaim) *jwt.Token {
-	token := jwt.New(jwt.SigningMethodRS256)
+func (tg *TokenManager) GenerateToken(method jwt.SigningMethod, identity Identity, kid string, extraClaims ...ExtraClaim) *jwt.Token {
```
alexeykazakov, line 247: Do we really need to use anything else but RS256 here? I don't see you pass anything else in https://github.com/codeready-toolchain/registration-service/pull/321

So I wonder if we can hardcode `jwt.SigningMethodRS256` here. Alternatively we could introduce one more function, something like `GenerateTokenWithMethod()` next to the current `GenerateToken()` which would use the default RS256 if you really need this.

mfrancisc (approved):

pkg/test/auth/tokenmanager.go
rajivnathan: I added this so I can test with a different signing method but then I changed the way I was testing that so this change is no longer needed.

PR #312, opened 2023-05-05
feat: client with event logging

xcoulon:
A facade to the default `client.Client`, which produces logs when
resources are created, updated (or patched) and deleted.

In the future, we could probably merge the current `ApplyClient` into
this new client (not done here as it is out-of-scope)

Also, rafactor the fake client to match the method signature of this
new client


MatousJobanek:
Could we close this PR?

xcoulon:
yes, I guess it's outdated now :(

PR #315, opened 2023-06-02
feat: monitor deployed versions of our sandbox operators

mfrancisc:
Implement functionality that compares deployed commits for our operators/deployments with the latest commit available in the GitHub repositories.

Jira: https://issues.redhat.com/browse/SANDBOX-53

Related PRs:
- https://github.com/codeready-toolchain/host-operator/pull/819
- https://github.com/codeready-toolchain/api/pull/364


pkg/status/versionchecks_test.go
```diff
@@ -0,0 +1,118 @@
...
+	"time"
+
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
+	"github.com/codeready-toolchain/toolchain-common/pkg/test"
+	"github.com/google/go-github/v52/github"
+	"github.com/migueleliasweb/go-github-mock/src/mock"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func TestCheckDeployedVersionIsUpToDate(t *testing.T) {
```
MatousJobanek, line 15: could you please add the `// given`, `// when`, and `// then` comments to each section of the sub-tests? It's easier to read

MatousJobanek (commented):
Looks cool :+1:

pkg/client/github_client.go
```diff
@@ -0,0 +1,26 @@
...
+	"golang.org/x/oauth2"
+)
+
+// NewGitHubClient return a client that interacts with GitHub and has rate limiter configured.
+// With authenticated GitHub api you can make 5,000 requests per hour.
+// But the Search API has a custom rate limit.
+// Unauthenticated clients are limited to 10 requests per minute, while authenticated clients can make up to 30 requests per minute.
+// see: https://github.com/google/go-github#rate-limiting
+//
+// The RoundTripper waits for the secondary rate limit to finish in a blocking mode and then issues/retries requests.
```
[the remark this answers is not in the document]
mfrancisc: From what I can see the `RoundTripper` helps with secondary rate limiting applied by GH: https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#secondary-rate-limits

Especially I was concerned about the "concurrent requests" part of that doc, since the controllers running on different clusters might call GH in parallel. 

About :
```
In the error cases, we would run out of the quota much quickly.
```
I think you are correct, probably I could implement the delay mechanism we discussed some time ago:
- save the last time we called the GH api in memory (using a variable in the controller), and trigger further calls only when there is let's say 5 minutes delay after last call . WDYT?

pkg/status/versionchecks.go
```diff
@@ -0,0 +1,60 @@
...
+	// DeploymentThreshold is the threshold after which we can be almost sure the deployment was not updated on the cluster with the latest version/commit,
+	// in this case some issue is preventing the new deployment to happen.
+	DeploymentThreshold = 30 * time.Minute
+)
+
+// CheckDeployedVersionIsUpToDate verifies if there is a match between the latest commit in Github for a given repo and branch matches the provided commit SHA.
+// There is some preconfigured delay/threshold that we keep in account before returning an `error condition`.
+func CheckDeployedVersionIsUpToDate(githubClient *github.Client, repoName, repoBranch, deployedCommitSHA string) *toolchainv1alpha1.Condition {
+	// get the latest commit from given repository and branch
+	commits, commitsResponse, err := githubClient.Repositories.ListCommits(context.TODO(), toolchainv1alpha1.ProviderLabelValue, repoName, &github.CommitsListOptions{
```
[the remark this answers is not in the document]
mfrancisc: good point, I can update the code to get only last commit using `HEAD`. 

Even if `HEAD` might be attached to some other branch or detached , but I don't think this should happen in our case right ?

pkg/status/versionchecks_test.go
mfrancisc: sure, let me fix this!

pkg/status/versionchecks.go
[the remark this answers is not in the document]
mfrancisc: done in https://github.com/codeready-toolchain/toolchain-common/pull/315/commits/40c65b28c47481b17a33888c6e1fedb1c60c2bad (I'll update host and member to pass `HEAD` as branch name)

pkg/status/versionchecks_test.go
mfrancisc: done, please see: https://github.com/codeready-toolchain/toolchain-common/pull/315/commits/40c65b28c47481b17a33888c6e1fedb1c60c2bad

pkg/status/versionchecks.go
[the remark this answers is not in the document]
mfrancisc: I forgot to update the unit tests 🤦‍♂️ . Please check also https://github.com/codeready-toolchain/toolchain-common/pull/315/commits/dd6a8284e2c66ee824a14421480d97d3a92278dd

pkg/status/versionchecks_test.go
mfrancisc: I missed few places in the previous commit, please check also: https://github.com/codeready-toolchain/toolchain-common/pull/315/commits/dd6a8284e2c66ee824a14421480d97d3a92278dd

pkg/client/github_client.go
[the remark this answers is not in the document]
MatousJobanek: Yes, having the fixed gaps between the calls should be the safest - 1 minute should be sufficient

> Especially I was concerned about the "concurrent requests" part of that doc, since the controllers running on different clusters might call GH in parallel.

would it be still a valid concern? I know that the concurrent calls could still happen, but that would be a very rare corner-case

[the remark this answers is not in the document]
mfrancisc: > Yes, having the fixed gaps between the calls should be the safest - 1 minute should be sufficient
ok , let me work on the fixed delays between calls.

> would it be still a valid concern? I know that the concurrent calls could still happen, but that would be a very rare corner-case
In theory it could happen, it's hard to say since Github doesn't provide clear requirements about concurrent calls. 

Anyway we can:
a ) remove this `RoundTripper` library and eventually see how it behaves without it, and add it if needed.
OR
b) provide a `Single Sleep Limit` option to the `RoundTripper`, so that we know for how long it will block the controller in case it happens (see: https://github.com/gofri/go-github-ratelimit#options)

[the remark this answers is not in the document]
mfrancisc: I've added the delay checks in host operator , please check https://github.com/codeready-toolchain/host-operator/pull/819/commits/d0384c77815b726c068230f89c018fdce41f3c5e

I'll eventually do the same in member-operator, once we clear out all the concerns in host-operator.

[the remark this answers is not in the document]
mfrancisc: @MatousJobanek I've also removed the GH rate limiter , as discussed earlier today, please see https://github.com/codeready-toolchain/toolchain-common/pull/315/commits/96782e36fc95cac6fb253aa8a5b9a7c9c52ddbb6 
and updated host-operator PR as accordingly: https://github.com/codeready-toolchain/host-operator/pull/819/commits/3b346c76d927a66755a277eb590c2acce0056754

pkg/client/github_client.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/google/go-github/v52/github"
+	"golang.org/x/oauth2"
+)
+
+// GitHubAPICallDelay it's used to "slow" down the number of requests we perform to GitHub API , in order to avoid rate limit issues.
+const GitHubAPICallDelay = 1 * time.Minute
```
MatousJobanek, line 12: let's either move this delay to the controller's code, or move the logic here. having the delay and the logic at two places can be complicated.

pkg/test/condition.go
```diff
@@ -27,9 +27,8 @@ func AssertConditionsMatch(t T, actual []toolchainv1alpha1.Condition, expected .
 // LastTransitionTime is ignored.
 func AssertContainsCondition(t T, conditions []toolchainv1alpha1.Condition, contains toolchainv1alpha1.Condition) {
 	for _, c := range conditions {
-		if c.Type == contains.Type {
+		if c.Type == contains.Type && c.Reason == contains.Reason {
 			assert.Equal(t, contains.Status, c.Status)
-			assert.Equal(t, contains.Reason, c.Reason)
```
MatousJobanek, line 32: Why this change?

pkg/status/versionchecks.go
```diff
@@ -0,0 +1,61 @@
...
+	expectedDeploymentTime := githubCommitTimestamp.Add(DeploymentThreshold) // let's consider some threshold for the deployment to happen
+	githubCommitSHA := *latestCommit.SHA
+	if githubCommitSHA != deployedCommitSHA && time.Now().After(expectedDeploymentTime) {
+		// deployed version is not up-to-date after expected threshold
+		err := fmt.Errorf("%s. deployed commit SHA %s ,github latest SHA %s, expected deployment timestamp: %s", ErrMsgDeploymentIsNotUpToDate, deployedCommitSHA, githubCommitSHA, expectedDeploymentTime.Format(time.RFC3339))
+		return NewComponentErrorCondition(toolchainv1alpha1.ToolchainStatusDeploymentNotUpToDateReason, err.Error())
+	}
+
+	// no problems with the deployment version, return a ready condition
+	return NewComponentReadyCondition(toolchainv1alpha1.ToolchainStatusDeploymentUpToDateReason)
```
MatousJobanek, line 60: related to my comment in https://github.com/codeready-toolchain/host-operator/pull/819#discussion_r1221704848 let's consider if it wouldn't make sense to create a new type of condition that would show that the check is running and the deployment is up-to-date.

pkg/status/versionchecks.go
```diff
@@ -0,0 +1,61 @@
...
+	// in this case some issue is preventing the new deployment to happen.
+	DeploymentThreshold = 30 * time.Minute
+)
+
+// CheckDeployedVersionIsUpToDate verifies if there is a match between the latest commit in Github for a given repo and branch matches the provided commit SHA.
+// There is some preconfigured delay/threshold that we keep in account before returning an `error condition`.
+func CheckDeployedVersionIsUpToDate(githubClient *github.Client, repoName, repoBranch, deployedCommitSHA string) *toolchainv1alpha1.Condition {
+	// get the latest commit from given repository and branch
+	latestCommit, commitResponse, err := githubClient.Repositories.GetCommit(context.TODO(), toolchainv1alpha1.ProviderLabelValue, repoName, repoBranch, &github.ListOptions{})
+	fmt.Printf("github commits response statusCode: %d", commitResponse.Response.StatusCode)
```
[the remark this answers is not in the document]
mfrancisc: nope, was leftover from debugging. removed in https://github.com/codeready-toolchain/toolchain-common/pull/315/commits/8dc20d30637d503651bc415fed8c453b4c518a4e

pkg/client/github_client.go
mfrancisc: I've moved the logic here, since we will have to use the same function in member-operator, this way we avoid some duplication (even if it's just few lines of code). Plz see: https://github.com/codeready-toolchain/toolchain-common/pull/315/commits/8dc20d30637d503651bc415fed8c453b4c518a4e

pkg/test/condition.go
mfrancisc: not needed, was leftover from when I was adding twice the `Ready` condition. Reverted in https://github.com/codeready-toolchain/toolchain-common/pull/315/commits/8dc20d30637d503651bc415fed8c453b4c518a4e

pkg/status/versionchecks.go
```diff
@@ -0,0 +1,60 @@
...
+	ErrMsgDeploymentIsNotUpToDate = "deployment version is not up to date with latest github commit SHA"
+
+	// DeploymentThreshold is the threshold after which we can be almost sure the deployment was not updated on the cluster with the latest version/commit,
+	// in this case some issue is preventing the new deployment to happen.
+	DeploymentThreshold = 30 * time.Minute
+)
+
+// CheckDeployedVersionIsUpToDate verifies if there is a match between the latest commit in Github for a given repo and branch matches the provided commit SHA.
+// There is some preconfigured delay/threshold that we keep in account before returning an `error condition`.
+func CheckDeployedVersionIsUpToDate(githubClient *github.Client, repoName, repoBranch, deployedCommitSHA string) *toolchainv1alpha1.Condition {
```
MatousJobanek, line 27: just a side comment - returning the pointer is a bit weird because it never happens that you would return `nil` - the condition is always set, so you could return it as a value. But I guess that you wanted to align with the `NewComponentReadyCondition` which return a pointer as well. Let's keep it as it is for now :+1:

pkg/status/versionchecks.go
```diff
@@ -0,0 +1,60 @@
...
+	// get the latest commit from given repository and branch
```
MatousJobanek, line 28: I have one proposal how we could make the code on the controller's side easier. When I mentioned in one of my comments that we could move entirely the check for the last call to toolchain-common, I mean everything. Looking at the code, we could move even more and simplify the code duplication on the controller side:
```suggestion

type VersionCheckManager struct {
	GetGithubClientFunc client.GetGitHubClientFunc
	lastGHCallsPerRepo  map[string]time.Time
}

// CheckDeployedVersionIsUpToDate verifies if there is a match between the latest commit in Github for a given repo and branch matches the provided commit SHA.
// There is some preconfigured delay/threshold that we keep in account before returning an `error condition`.
func (m *VersionCheckManager) CheckDeployedVersionIsUpToDate(isProd bool, accessTokenKey string, alreadyExistingConditions []toolchainv1alpha1.Condition, repoName, repoBranch, deployedCommitSHA string) *toolchainv1alpha1.Condition {
	// the first two checks are pretty much the same for all components
	if !isProd {
		cond := NewDeploymentVersionUpToDateCondition(toolchainv1alpha1.ToolchainStatusDeploymentVersionCheckDisabledReason)
		// you could change the NewDeploymentVersionUpToDateCondition function (or create a new one) so it would accept also the message
		cond.Message = "is not running in prod environment"
		return cond
	}
	if accessTokenKey == "" {
		cond := NewDeploymentVersionUpToDateCondition(toolchainv1alpha1.ToolchainStatusDeploymentVersionCheckDisabledReason)
		cond.Message = "access token key is not provided"
		return cond
	}
	// we can store the last call per repo name, so it will solve the gaps between calls for host & reg-service which is done form the same controller
	if m.lastGHCallsPerRepo == nil {
		m.lastGHCallsPerRepo = map[string]time.Time{}
	}
	lastCall, present := m.lastGHCallsPerRepo[repoName]
	if present && !client.CanIssueGitHubRequest(lastCall) {
		// we need to make the gap between the calls for this repo, but we have to set the condition to something - let's reuse what was set last time
		previouslySet, found := condition.FindConditionByType(alreadyExistingConditions, toolchainv1alpha1.ToolchainStatusDeploymentVersionCheckDisabledReason)
		if !found {
			cond := NewDeploymentVersionUpToDateCondition("unknown")
			cond.Message = "waiting for the next attempt"
			return cond
		}
		return &previouslySet
	}
	m.lastGHCallsPerRepo[repoName] = time.Now()

	githubClient := m.GetGithubClientFunc(accessTokenKey)
	// get the latest commit from given repository and branch
```

on the controller's side, you would initiate the manager as part of creating the Reconciler:
```go
if err := (&toolchainstatus.Reconciler{
	Client:         mgr.GetClient(),
	Scheme:         mgr.GetScheme(),
	HTTPClientImpl: &http.Client{},
	VersionCheckManager: status.VersionCheckManager{
		GetGithubClientFunc: commonclient.NewGitHubClient,
	},
	GetMembersFunc: commoncluster.GetMemberClusters,
	Namespace:      namespace,
}).SetupWithManager(mgr); err != nil {
	setupLog.Error(err, "unable to create controller", "controller", "ToolchainStatus")
	os.Exit(1)
}
```
so it could be easily overridden in unit tests

MatousJobanek (approved):
Looks nice and clean. :+1: 
Great job :1st_place_medal: :rocket:

PR #319, opened 2023-07-03
test(spaces): move from host-operator repo

xcoulon:
and make it available for other repos including sandbox-sre


MatousJobanek (approved):
I'm wondering if there is already some duplication of the similar code in the e2e tests :thinking:

xcoulon:
> I'm wondering if there is already some duplication of the similar code in the e2e tests 🤔

good point, I'll take a look!

xcoulon:
> Looks good but could you please add unit tests for pkg/hash/hash.go?

sure! done in a9d17f6

PR #326, opened 2023-09-19
support email lists in notifications

MatousJobanek:
so we can provide multiple email addresses for notifications


alexeykazakov (approved):

pkg/notification/notification_builder_test.go
```diff
@@ -70,21 +55,46 @@ func TestNotificationBuilder(t *testing.T) {
 		userSignup.Status = toolchainv1alpha1.UserSignupStatus{
 			CompliantUsername: "jsmith",
 		}
+		emailsToTest := []string{
+			userSignup.Annotations[toolchainv1alpha1.UserSignupUserEmailAnnotationKey],
+			"john.wick@subdomain.domain.com",
+			"john-Wick@domain.com",
+			"john@domain.com,another-john@some.com",
+			"john@domain.com, with-comma@some.com, ",
+			"john@domain.com,another-john@some.com, with-space@john.com",
```
mfrancisc, line 64: I see from the [ParseAddressList function doc](https://pkg.go.dev/net/mail#ParseAddressList) that it supports also:
`Alice <alice@example.com>, Bob <bob@example.com>, Eve <eve@example.com>`

not sure if we want to enumerate also that.

mfrancisc (approved):
Nice 👍

PR #329, opened 2023-10-02
feat: Implement public API that will open SpaceBindings to end users through proxy part 2

mfrancisc:
This PR introduces:
-  spacebinding lister that will be used in both registration-service and host-operator (follow up refactoring PRs to replace [this code here](https://github.com/codeready-toolchain/host-operator/blob/c132bf45b05ad1471fa58c92c8c8592bae44a8d6/controllers/space/space_controller.go#L258-L313)) for listing spacebindings for a given space recursively ( it searches for spacebindings  on the parent spaces as well).
- some cleanup when creating mocked spaces
- utility for populating bindings field in workspace api

Jira: https://issues.redhat.com/browse/ASC-410

Related PRs:
- registration-service: https://github.com/codeready-toolchain/registration-service/pull/350
- toolchain-e2e: https://github.com/codeready-toolchain/toolchain-e2e/pull/805


pkg/spacebinding/spacebindinglister_test.go
```diff
@@ -0,0 +1,104 @@
...
+				spacetest.WithLabel(toolchainv1alpha1.SpaceCreatorLabelKey, "johnny"),
+				spacetest.WithSpecParentSpace(space.GetName()),
+			)
+			mur := NewMasterUserRecord(t, "johnny", TargetCluster(test.MemberClusterName), TierName("deactivate90"))
+			listSpaceBindingsFunc := func(spaceName string) ([]toolchainv1alpha1.SpaceBinding, error) {
+				return []toolchainv1alpha1.SpaceBinding{
+					*spacebinding.NewSpaceBinding(mur, space, "johnny", spacebinding.WithRole("admin")),
+				}, nil
+			}
+			getSpaceFunc := func(spaceName string) (*toolchainv1alpha1.Space, error) {
```
alexeykazakov, line 38: I would make this test a bit smarter to be closer to the real world and cover more usecases.

First at all I would create at least three layers (so you can fully test the recursion): Parent, two child, and maybe one grandchild.

Then create two smart mock function:

```
getSpaceFunc := func(spaceName string) (*toolchainv1alpha1.Space, error) {
        for _, space := range spaces {
                if space.Name == spaceName {
                        return space, mil
                }
        }
        return nil, nil
}
```
```
listSpaceBindingsFunc := func(spaceName string) ([]toolchainv1alpha1.SpaceBinding, error) {
        space, _ := getSpaceFunc(spaceName)
	return []toolchainv1alpha1.SpaceBinding{
		*spacebinding.NewSpaceBinding(mur, space, spaceName + "-binding", spacebinding.WithRole("admin")),
	}, nil
}
```
Then I would test all three cases when you list SBs for all three layers. You could create a matrix of the given data and given space you want to list SB for and the expected spacebings to avoid code duplication.

alexeykazakov (commented):
Looks good. I would just improve the tests.

pkg/spacebinding/spacebindinglister.go
```diff
@@ -0,0 +1,65 @@
...
+		return foundBindings, errs.Wrap(err, "unable to get parent-space")
+	}
+
+	return l.ListForSpace(parentSpace, foundBindings)
+}
+
+func mergeSpaceBindings(foundBindings []toolchainv1alpha1.SpaceBinding, parentBindings []toolchainv1alpha1.SpaceBinding) []toolchainv1alpha1.SpaceBinding {
+	// if both list are not empty, we have to merge them.
+	// roles for the same username on SpaceBindings will override those on parentSpaceBindings,
+	// so let's remove them from parentSpaceBinding before merging the two lists.
```
alexeykazakov, line 51: BTW we also need a test for the case when a SB is overridden when merging.

pkg/spacebinding/spacebindinglister_test.go
mfrancisc: good point! Plz take a look at https://github.com/codeready-toolchain/toolchain-common/pull/329/commits/ef238f6853844b96929ed97b19bfa678b89a9b54

I couldn't implement `listSpaceBindingsFunc` exactly how you suggested it, since I needed more control over the spacebindigns that are returned. For example I don't always want to return a spacebinding object for every space in the tree, but it was good source of inspiration.

Please let me know if some use-case is missing.

pkg/spacebinding/spacebindinglister.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func NewLister(listSpaceBindingsFunc func(spaceName string) ([]toolchainv1alpha1.SpaceBinding, error), getSpaceFunc func(spaceName string) (*toolchainv1alpha1.Space, error)) *Lister {
+	return &Lister{
+		ListSpaceBindingsFunc: listSpaceBindingsFunc,
+		GetSpaceFunc:          getSpaceFunc,
+	}
+}
+
+// ListForSpace recursively searches all the SpaceBindings of a given Space,
```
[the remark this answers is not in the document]
mfrancisc: Makes sense. Thanks for the suggestion!

See https://github.com/codeready-toolchain/toolchain-common/pull/329/commits/01632b1bc715c6c275b534871dc788f99dc7f6c9

pkg/spacebinding/spacebindinglister_test.go
```diff
@@ -0,0 +1,222 @@
...
+		// the defaultGetSpaceFunc returns the space based on the given name
+		defaultGetSpaceFunc := func(spaceName string) (*toolchainv1alpha1.Space, error) {
+			if space, found := spaces[spaceName]; found {
+				return space, nil
+			}
+			return nil, fmt.Errorf("space not found: %s", spaceName)
+		}
+
+		// listParentSpaceBindingFunc returns the spacebinding for the parentSpace
+		listParentSpaceBindingFunc := func(spaceName string) ([]toolchainv1alpha1.SpaceBinding, error) {
```
alexeykazakov, line 58: This function looks wrong TBH. You don't really test your code properly with that. I would just create a singe function which would do that switch with all the cases (it returns SBs for all the spaces depending on the `spaceName`). And I would use that single function in all the tests. Basically you could just merge all the custom functions you use below to a single one. Something like:

```
func(spaceName string) ([]toolchainv1alpha1.SpaceBinding, error) {
	switch spaceName {
	case parentSpace.GetName():
		return []toolchainv1alpha1.SpaceBinding{
			*spacebinding.NewSpaceBinding(mur, parentSpace, parentSpace.Name, spacebinding.WithRole("admin")),
		}, nil
	case grandChildSpace1.GetName():
		return []toolchainv1alpha1.SpaceBinding{
			*spacebinding.NewSpaceBinding(mur, grandChildSpace1, grandChildSpace1.Name, spacebinding.WithRole("viewer")),
		}, nil
	...
	default:
		return nil, nil
	}
},
```
In the runtime code you use the same listSB function. Which returns SBs for the provided space. In the test, however, your mock function returns only the expected SBs.
For example, in the first test the mock function will always return the parent SB no matter what space the runtime code asked the SBs for. It can hide bugs in the code.

But I like the new tests much better overall! Good job!

mfrancisc: If we always return a spacebinding we will not test the case in which the role is inherited from one of the parent spaces right ?

alexeykazakov: In the runtime you provide the same function which returns the actual SBs for the provided space, right?
Shouldn't we do the same in the test? You specify the same single space hierarchy for all the tests. You already know what SBs should be return for a specific space. And this function, in the runtime code, doesn't know anything about the inherited roles. It just returns SBs CRs for the space. It's the job of the lister to look up for SBs in all the parent spaces, so the inherited SBs are included too.
Shouldn't we do the same in the test?
Or I'm missing something?

mfrancisc: > In the runtime you provide the same function which returns the actual SBs for the provided space, right?

Not always. I think this is more close to what happens at runtime because in the case of a subSpace (let's say created with `SpaceRequest`) there is no `SpaceBinding` resource directly linked with it, by default it inherits the parent spacebindings only. So this call here will return empty list https://github.com/codeready-toolchain/toolchain-common/pull/329/files#diff-c83558581efe8e9049f110acd8dc83b158f52beeab1b6da8ee1299ab9ccda983R24
when called with the subSpace.
But the function will go on and search for the spaceBindings associated with the parentSpace and will return them here: https://github.com/codeready-toolchain/toolchain-common/pull/329/files#diff-c83558581efe8e9049f110acd8dc83b158f52beeab1b6da8ee1299ab9ccda983R39-R45

> And this function, in the runtime code, doesn't know anything about the inherited roles. It just returns SBs CRs for the space. It's the job of the lister to look up for SBs in all the parent spaces, so the inherited SBs are included too.

The lister also merges the roles based on the MUR, the child one overrides the parent ones. So if I well understood your proposal - to  always return a spacebinding resource for each space, we are already covering this case here https://github.com/codeready-toolchain/toolchain-common/pull/329/files#diff-8467fbd206b9f86061c7253992813af5478d514bd8061a26920f13fc5042ffb0R142-R168

The other test cases are also covering the cases in which this call https://github.com/codeready-toolchain/toolchain-common/pull/329/files#diff-c83558581efe8e9049f110acd8dc83b158f52beeab1b6da8ee1299ab9ccda983R24 returns an empty list, but subsequent (parent spaces) or previous calls (grandchild) return a spacebinding.

mfrancisc: Following is a list of the hierarchies we cover with this code:

1) parentSpace has it's own spacebinding
```
johnny
|__ parentSpace -> parentSpaceBinding role=admin
```
**result:** list for `parentSpace` expects to return `admin` 

2) spacebinding for childSpace1 is inherited from parentSpace
```
johnny
|__ parentSpace        ---> parentSpaceBinding role=admin
      |__ childSpace1  ---> no specific spacebinding, inherits the one above role=admin
```
**result:** list for `childSpace1` expects to return `admin` 

3) override parentSpace-binding in childSpace2
```
johnny
|__ parentSpace       ---> parentSpaceBinding role=admin
      |__ childSpace1 ---> childSpace2Binding role=viewer
```
**result:** list for `childSpace1` expects to return `viewer`

4) no spacebinding on child1 and grandChildSpace1, spacebinding for grandChildSpace1 is inherited from parentSpace
```
johnny
|__ parentSpace                   ---> parentSpaceBinding role=admin
      |__ childSpace1             ---> no specific spacebinding, inherits the one above role=admin
            |__ grandChildSpace1  ---> no specific spacebinding, inherits the one from parentSpace role=admin because of the recursive searches up for parentSpaces  
```
**result:** list for `grandChildSpace1` expects to return `admin` 

5) no spacebinding on child1 but we do have one on grandChildSpace1 which overrides parentSpace-binding
```
johnny
|__ parentSpace                  ---> parentSpaceBinding role=admin
      |__ childSpace1            ---> no specific spacebinding, inherits the one above role=admin
            |__ grandChildSpace1 ---> grandChildSpace1Binding role=viewer   
```
**result:** list for `grandChildSpace1` expects to return `viewer`

6) spacebinding on grandChildSpace1 overrides child1-binding and parentSpace-binding
```
johnny
|__ parentSpace                  ---> parentSpaceBinding role=admin
      |__ childSpace1            ---> childSpace1Binding role=maintainer
            |__ grandChildSpace1 ---> grandChildSpace1Binding role=viewer   
```
**result:** list for `grandChildSpace1` expects to return `viewer`

NOTE: I can create 2 other use-cases based on this last test case , to list for `parentSpace` and `grandChildSpace1` but not sure how much additional value this will bring since the mock will always return only that specific spacebinding associated with the space and the merge function will override the parent ones. But I can add those.

Now that I look at this, what is still missing IMO, is some test case with more MURs, so that the listing returns more than one spacebinding for different users, maybe with different roles.

For example:
```
johnny
|__ parentSpace                  ---> parentSpaceBinding role=admin
      |__ childSpace1            ---> no specific spacebinding, inherits the one above role=admin
            |__ grandChildSpace1 ---> no specific spacebinding, inherits the one above role=admin
batman
|__ parentSpace                  ---> no access
      |__ childSpace1            ---> childSpace1Binding role=maintainer
            |__ grandChildSpace1 ---> no specific spacebinding, inherits the one above role=maintainer
robin
|__ parentSpace                  ---> no access
      |__ childSpace1            ---> no access
            |__ grandChildSpace1 ---> grandChildSpace1Binding role=viewer   
```
**results:**
list for _parentSpace_ should return _johnny role=admin_
list for _childSpace1_ should return _johnny role=admin_ , _batman role=maintainer_
list for _grandChildSpace1_ should return _johnny role=admin_ , _batman role=maintainer_ , _robin role=viewer_

sorry for the long comment! 

WDYT?

mfrancisc: @alexeykazakov as discussed yesterday, I've updated the test which now builds the hierarchy upfront , I've also added a doc file with the hierarchy diagram (note needs mermaid plugin locally, but github should render that by default)

please see: https://github.com/codeready-toolchain/toolchain-common/pull/329/commits/9e0ee2b48d1177745f66bf43a8feff148336fcb5

alexeykazakov (approved):
Looks good. Thank you!
Just maybe mention `pkg/spacebinding/spacebindinglister_test.md` in the test code.

mfrancisc:
> Just maybe mention pkg/spacebinding/spacebindinglister_test.md in the test code.

I've added a small comment about the doc. Thanks for the review!

pkg/spacebinding/spacebindinglister_test.md
```diff
@@ -0,0 +1,32 @@
+## SpaceBinding lister test 
+
+The spacebinding lister test verifies that `ListForSpace` function returns the expected spacebindings for a given space resource.
+
+At a high level, we want to make sure that given a specific space, the function covers the following requirements:
+
+- we get spcebindings resources for users that have access to that space
+- we want to make sure the spacebindings are being inherited from the parent spaces
+- spacebindings on the sub-space will override those on the parent space for the same MUR
+- users can have different roles (or no access at all) on the various spaces in the hierarchy
```
MatousJobanek, line 10: one user should have one single role in a particular Space

pkg/spacebinding/spacebindinglister_test.md
```diff
@@ -0,0 +1,32 @@
...
+    D --> E(Space-E <hr> <div align='left'> catwomen - admin <i>inherited</i> <br> batman - admin <i>inherited</i></div>)
+    C --> F(Space-F <hr> <div align='left'> catwomen - admin <i>inherited</i> <br> batman - maintainer <i>inherited</i> <br> robin - viewer</div>)
+```
+
+Given the above hierarchy the expected result for each space should be:
+
+|          |  Space-A  |   Space-B   |  Space-C  | Space-D | Space-E |  Space-F  |
+|:--------:|:---------:|:-----------:|:---------:|:-------:|:-------:|:---------:|
+| catwomen |   admin   |    admin    |   admin   | admin | admin |   admin   |
+|  batman  | no access |  maintainer | no access | admin | admin | no access |
```
MatousJobanek, line 31:
```suggestion
|  batman  | no access |  maintainer | maintainer | admin | admin | no access | 
```

pkg/spacebinding/spacebindinglister_test.go
```diff
@@ -0,0 +1,238 @@
...
+					*spacebinding.NewSpaceBinding(catwomenMur, spaceA, spaceA.Name, spacebinding.WithRole("admin")),
+				}, nil
+			case spaceB.GetName():
+				return []toolchainv1alpha1.SpaceBinding{
+					*spacebinding.NewSpaceBinding(batmanMur, spaceB, spaceB.Name, spacebinding.WithRole("maintainer")),
+				}, nil
+			case spaceC.GetName():
+				return []toolchainv1alpha1.SpaceBinding{
+					*spacebinding.NewSpaceBinding(batmanMur, spaceC, spaceC.Name, spacebinding.WithRole("maintainer")),
+				}, nil
```
MatousJobanek, line 93: what is the benefit (for the purpose of the test) of having two spaces with the same SpaceBindings (maintainer for batmat) and inheriting from the same parent space?
Wouldn't make more sense to either drop one of them or let spaceC inherit from spaceB and expect no change there even though there is another SpaceBinding?

MatousJobanek (approved):
Looks good, I like how the test is written :+1: :rocket: 
Just some minor comments

pkg/spacebinding/spacebindinglister_test.md
mfrancisc: thanks! missed that!

pkg/spacebinding/spacebindinglister_test.md
mfrancisc: hmm yeah, let me change that last sentence a bit. thanks!

pkg/spacebinding/spacebindinglister_test.go
mfrancisc: Good point. Let's have spaceC inheriting from spaceB and expect no change, this is not covered as usecase, even if it doesn't make much sense to create a spacebinding with a role that the user already has (`maintainer` in this case) , but someone might create one by mistake and things should still work.

pkg/spacebinding/spacebindinglister_test.md
mfrancisc: changed in https://github.com/codeready-toolchain/toolchain-common/pull/329/commits/6f1d0354d896c222992c4c482fc0b80cfab19427

pkg/spacebinding/spacebindinglister.go
```diff
@@ -0,0 +1,65 @@
...
+			if parentBindings[i].Spec.MasterUserRecord == spaceBinding.Spec.MasterUserRecord {
+				parentBindings = append(parentBindings[:i], parentBindings[i+1:]...)
+				break
+			}
+		}
+	}
+	// merge lists now that there are no duplicates
+	// and just use the TypeMeta and ListMeta objects from the spaceBinding list
+	// since only the Items field is relevant from this object
+	return append(foundBindings, parentBindings...)
```
MatousJobanek, line 64: looking at the PR in registration-service - we could modify this function to make sure that we return always only one SpaceBinding per username, so we don't have to care about the situation when one user would have accidentally more SpaceBindings, right?

mfrancisc: I can create a map with the username as key, but if there is more then one spacebinding per username I guess we want to be spot that out and throw an error. We can return an error from here and let the client (registration-serivce and host-operator) decide what todo with it in theory. Or as we do right now, the client decides if it's ok or not to have more then one spacebinding per username.

MatousJobanek: the question is where to show the error. IMO, this and the registration-service are the wrong places for throwing an error. If two SpaceBindings are created for one MUR, then the generic Space controller logic will apply templates of  both roles to the given username, which means that the user will really have both roles applied in the namespaces.
Let's do the check in SpaceBindingRequest API - if there is already a SpaceBinding for the given MUR, then the SBR should wail with appropriate condition

MatousJobanek: btw, feel free to create a new story for it and do it as a follow-up of the whole effort. We don't have to do it now.

mfrancisc: > Theoretically, the MUR uniqueness should be ensured by using always the same naming format (which is something we do) so the risk is very minimal.

Right, I think this case is already covered in the SBR controller implicitly by using the same names for the `spacebinding` resource. So in theory if the user tries to create a SBR for the same MUR and Space, the SBR controller will throw an error here: https://github.com/codeready-toolchain/host-operator/blob/8e85ee74a26db8c7c8f7aebaac53449e7b688e13/controllers/spacebindingrequest/spacebindingrequest_controller.go#L286-L296

Is this what you were thinking about or is any other think to cover that I'm missing?

mfrancisc: > btw, feel free to create a new story for it and do it as a follow-up of the whole effort. 

I have created ASC-430 for this small improvement.

PR #336, opened 2023-10-16
Remove unused code related to Che

baijum:
Ref. https://issues.redhat.com/browse/SANDBOX-345


pkg/configuration/load_test.go
```diff
@@ -218,59 +218,59 @@ func TestNoWatchNamespaceSetWhenLoadingConfigMap(t *testing.T) {
 
 func TestLoadSecrets(t *testing.T) {
 	secretData := map[string][]byte{
-		"che-admin-username": []byte("cheadmin"),
-		"che-admin-password": []byte("password"),
+		"admin-username": []byte("cheadmin"),
+		"admin-password": []byte("password"),
```
alexeykazakov, line 222: Those test secrets values are still che related. Let's maybe change them to something more generic. Something like:
```suggestion
		"key-1": []byte("value-1"),
		"key-2": []byte("value-2"),
```

rajivnathan: +1

baijum: Done. c5c1e6b3ad0f2fe6731f35f2cebc4a8b9288f5a9

alexeykazakov (approved):
Looks good. Thank you.

PR #337, opened 2023-11-07
Check disable inheritance when finding spacebindings

mmulholla:
Updates for disable inheritance see: https://issues.redhat.com/browse/ASC-428.
This change is for spaces only. It is used by host-operator a registration-service (proxy) without modification and will not cause regression as it requires the new disableInheritance attribute to be set to true (default is false). It also requires the lates to committed change to the api repository.  

A new test will also be added to toolchain-e2e . Will add PR link here when I have it.


pkg/spacebinding/spacebindinglister_test.go
```diff
@@ -235,4 +235,207 @@ func TestNewSpaceBindingLister(t *testing.T) {
...
+				return space, nil
+			}
+			return nil, fmt.Errorf("space not found: %s", spaceName)
+		}
+
+		// listSpaceBindingFunc returns the specific spacebindings for the given space
+		listSpaceBindingFunc := func(spaceName string) ([]toolchainv1alpha1.SpaceBinding, error) {
+			switch spaceName {
+			case spaceA.GetName():
+				fmt.Println("return binding for SpaceA")
```
MatousJobanek, line 303: do we need all these `fmt.Println` lines? if we need to log some messages, then let's use `t.Log` instead, but still, I'm not sure if it's needed in this case.

pkg/spacebinding/spacebindinglister_test.go
```diff
@@ -235,4 +235,207 @@ func TestNewSpaceBindingLister(t *testing.T) {
...
+				},
+			},
+			"for Space-C": {
+				space:                 spaceC,
+				getSpaceFunc:          getSpaceFunc,
+				listSpaceBindingsFunc: listSpaceBindingFunc,
+				expectedSpaceBindings: []toolchainv1alpha1.SpaceBinding{
+					// we expect spacebinding inherited from space-B
+					*spacebinding.NewSpaceBinding(spaceBMur, spaceB, spaceB.Name, spacebinding.WithRole("viewer")),
+					// and specific one from space-C (which overrides the one from space-B even if it has the same role)
```
MatousJobanek, line 367: it doesn't override anything - I guess that it's copy-paste from the previous test case
```suggestion
					// and the one from space-C
```

pkg/spacebinding/spacebindinglister_test.go
```diff
@@ -235,4 +235,207 @@ func TestNewSpaceBindingLister(t *testing.T) {
...
+					// specific one only for this space
+					*spacebinding.NewSpaceBinding(spaceFMur, spaceF, spaceF.Name, spacebinding.WithRole("viewer")),
+				},
+			},
+		}
+		fmt.Println("\nStart DI test")
+		for k, tc := range tests {
+			t.Run(k, func(t *testing.T) {
+
+				fmt.Println(fmt.Sprintf("\nCheck space %s", tc.space.GetName()))
```
MatousJobanek, line 404: I guess that we don't need any of these logs eitheir
```suggestion
		}
		for k, tc := range tests {
			t.Run(k, func(t *testing.T) {
```

MatousJobanek (commented):
looks good :+1: 
Could you please also open a PR in:
* host-operator 
* registration-service

with the `replace` clause pointing to your `CheckDisableInheritance` branch of the forked toolchain-common repo to verify that the e2e tests will pass with these changes.

pkg/spacebinding/spacebindinglister_test.go
mmulholla: removed

pkg/spacebinding/spacebindinglister_test.go
mmulholla: removed.

mmulholla:
> looks good 👍 Could you please also open a PR in:
> 
> * host-operator
> * registration-service
> 

host operator: https://github.com/codeready-toolchain/host-operator/pull/932
registration service: https://github.com/codeready-toolchain/registration-service/pull/370

also toolchain-e2e which include a new test which test the new code in this PR: https://github.com/codeready-toolchain/toolchain-e2e/pull/841

MatousJobanek:
> also toolchain-e2e which includes a new test which test the new code in this PR: https://github.com/codeready-toolchain/toolchain-e2e/pull/841

nice :+1: thanks for doing that, you just need to pair the e2e PR with the host one to make it work

pkg/spacebinding/spacebindinglister_test.go
```diff
@@ -235,4 +235,199 @@ func TestNewSpaceBindingLister(t *testing.T) {
...
+				}, nil
+			case spaceC.GetName():
+				return []toolchainv1alpha1.SpaceBinding{
+					*spacebinding.NewSpaceBinding(spaceCMur, spaceC, spaceC.Name, spacebinding.WithRole("maintainer")),
+				}, nil
+			case spaceD.GetName():
+				return []toolchainv1alpha1.SpaceBinding{
+					*spacebinding.NewSpaceBinding(spaceDMur, spaceD, spaceD.Name, spacebinding.WithRole("admin")),
+				}, nil
+			case spaceE.GetName():fmt.Println("return no binding for SpaceE")
```
mfrancisc, line 318: even if I'm not sure we really need this print line. Maybe we could even remove it? 
```suggestion
			case spaceE.GetName():
			        fmt.Println("return no binding for SpaceE")
```

pkg/spacebinding/spacebindinglister_test.go
```diff
@@ -235,4 +235,199 @@ func TestNewSpaceBindingLister(t *testing.T) {
...
+		for k, tc := range tests {
+			t.Run(k, func(t *testing.T) {
+
+				// when
+				spaceBindingLister := spacebinding.NewLister(tc.listSpaceBindingsFunc, tc.getSpaceFunc)
+
+				// then
+				spaceBindings, err := spaceBindingLister.ListForSpace(tc.space, []toolchainv1alpha1.SpaceBinding{})
+				if tc.expectedErr != "" {
+					assert.EqualError(t, err, tc.expectedErr)
```
mfrancisc, line 403: I don't see an expected error in any of the test cases. Maybe we can simplify by removeing this if/else statement ?

mfrancisc (commented):
Nice Job 🏅 

I have some minor comments related to the unit tests.

Also there is a document explaining what the test is doing and the hierarchy/scenario being tested here https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/spacebinding/spacebindinglister_test.md

should we update it to reflect this new scenario ?

pkg/spacebinding/spacebindinglister_test.go
mmulholla: thx, removed it

mmulholla:
> > also toolchain-e2e which includes a new test which test the new code in this PR: [codeready-toolchain/toolchain-e2e#841](https://github.com/codeready-toolchain/toolchain-e2e/pull/841)
> 
> nice 👍 thanks for doing that, you just need to pair the e2e PR with the host one to make it work

New host operator PR: https://github.com/codeready-toolchain/host-operator/pull/935

pkg/spacebinding/spacebindinglister_test.go
```diff
@@ -235,4 +235,199 @@ func TestNewSpaceBindingLister(t *testing.T) {
...
+		spaceAMur := NewMasterUserRecord(t, "space-a-user", TargetCluster(test.MemberClusterName), TierName("deactivate90"))
+		spaceBMur := NewMasterUserRecord(t, "space-b-user", TargetCluster(test.MemberClusterName), TierName("deactivate90"))
+		spaceCMur := NewMasterUserRecord(t, "space-c-user", TargetCluster(test.MemberClusterName), TierName("deactivate90"))
+		spaceDMur := NewMasterUserRecord(t, "space-d-user", TargetCluster(test.MemberClusterName), TierName("deactivate90"))
+		spaceFMur := NewMasterUserRecord(t, "space-f-user", TargetCluster(test.MemberClusterName), TierName("deactivate90"))
+		// we have a parent space
+		spaceA := spacetest.NewSpace(test.HostOperatorNs, "Space-A",
+			spacetest.WithTierName("advanced"),
+			spacetest.WithSpecTargetCluster(test.MemberClusterName),
+			spacetest.WithDisableInheritance(false),
```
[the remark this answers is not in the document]
mfrancisc: Thanks for looking into it.

Unless I'm missing something, I just tried removing all the `spacetest.WithDisableInheritance(false)` and all tests are passing. It is set to false by default, as you wrote here https://github.com/codeready-toolchain/api/blob/6074997e1bd7b81cc1d21617c69f9a6a897e615a/api/v1alpha1/space_types.go#L76.

Not a big issue, so feel free to leave it as is if you think it's more clear.

mfrancisc:
Could you please check if the test documentation  https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/spacebinding/spacebindinglister_test.md needs to be updated with this new scenario?

pkg/spacebinding/spacebindinglister_test.go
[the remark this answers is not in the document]
MatousJobanek: Francisc is right, leaving it unset means `false`, which is exactly what we need - disabled by default. It would be different if you would define the field as a pointer, in that case leaving it unset would mean `nil`, but this is not this case.
I didn't see this line as confusing - could be used to document what the test is focused on. 
That being said, either way is fine for me :peace_symbol:

MatousJobanek:
Let's not document every test case in this unit test. While the already existing test is pretty complicated and understanding it without a diagram is a challenge, this new test scenario is pretty simple. Let's skip it for this one and let's discuss other options of documenting the original test-scenario offline.

MatousJobanek (approved):
Nice :+1: :rocket: 
Checking the e2e tests now

mmulholla:
@MatousJobanek is it OK to merge this now? The test in the PR's are passing - only failure is due to the replace in go.mod:

https://github.com/codeready-toolchain/toolchain-e2e/pull/841
https://github.com/codeready-toolchain/host-operator/pull/935
https://github.com/codeready-toolchain/registration-service/pull/370

The toolchain-e2e includes the test so once this is merged I can remove the replace from go.mod and rebuild.

alexeykazakov:
Yes, you can merge this PR now. Then update all the dependent PRs and get them merged if they are green and approved.

PR #351, opened 2024-01-09
move SpaceBindingRequest helpers 

mfrancisc:
Move spacebindingrequest helpers from [host-operator](https://github.com/codeready-toolchain/host-operator/blob/master/test/spacebindingrequest/spacebindingrequest.go) so they can be reused.

I'll replace the usages in follow up PRs.

### DIFF
--- a/pkg/test/spacebindingrequest/spacebindingrequest.go
+++ b/pkg/test/spacebindingrequest/spacebindingrequest.go
@@ -1,0 +1,66 @@
+package spacebindingrequest
+
+import (
+	"time"
+
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
+	"github.com/gofrs/uuid"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+type Option func(spaceRequest *toolchainv1alpha1.SpaceBindingRequest)
+
+func NewSpaceBindingRequest(name, namespace string, options ...Option) *toolchainv1alpha1.SpaceBindingRequest {
+	spaceBindingRequest := &toolchainv1alpha1.SpaceBindingRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			UID:       types.UID(uuid.Must(uuid.NewV4()).String()),
+		},
+	}
+	for _, apply := range options {
+		apply(spaceBindingRequest)
+	}
+	return spaceBindingRequest
+}
+
+func WithMUR(mur string) Option {
+	return func(spaceBindingRequest *toolchainv1alpha1.SpaceBindingRequest) {
+		spaceBindingRequest.Spec.MasterUserRecord = mur
+	}
+}
+
+func WithSpaceRole(spaceRole string) Option {
+	return func(spaceBindingRequest *toolchainv1alpha1.SpaceBindingRequest) {
+		spaceBindingRequest.Spec.SpaceRole = spaceRole
+	}
+}
+
+func WithLabel(key, value string) Option {
+	return func(space *toolchainv1alpha1.SpaceBindingRequest) {
+		if space.ObjectMeta.Labels == nil {
+			space.ObjectMeta.Labels = map[string]string{}
+		}
+		space.ObjectMeta.Labels[key] = value
+	}
+}
+
+func WithDeletionTimestamp() Option {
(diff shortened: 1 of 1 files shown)


pkg/test/spacebindingrequest/spacebindingrequest.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+type Option func(spaceRequest *toolchainv1alpha1.SpaceBindingRequest)
+
+func NewSpaceBindingRequest(name, namespace string, options ...Option) *toolchainv1alpha1.SpaceBindingRequest {
+	spaceBindingRequest := &toolchainv1alpha1.SpaceBindingRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			UID:       types.UID(uuid.Must(uuid.NewV4()).String()),
```
filariow, line 19: this can cause a `panic`. Is it ok or we'd prefer to return `(*toolchainv1alpha1.SpaceBindingRequest, error)`?

filariow (commented):
just a nit...

pkg/test/spacebindingrequest/spacebindingrequest.go
mfrancisc: I think it's fine to ignore that, since those are test utilities , not production code. But I'm fine with adding the err if you think it's more correct.

filariow (approved):
lgtm

alexeykazakov (approved):

pkg/test/spacebindingrequest/spacebindingrequest.go
```diff
@@ -0,0 +1,66 @@
...
+		now := metav1.NewTime(time.Now())
+		spaceBindingRequest.DeletionTimestamp = &now
+	}
+}
+
+func WithFinalizer() Option {
+	return func(spaceBindingRequest *toolchainv1alpha1.SpaceBindingRequest) {
+		spaceBindingRequest.Finalizers = append(spaceBindingRequest.Finalizers, toolchainv1alpha1.FinalizerName)
+	}
+}
```
sadlerap, line 60: Should this check if the given finalizer is already set?

sadlerap (commented):
One question, otherwise lgtm

pkg/test/spacebindingrequest/spacebindingrequest.go
filariow: may [controller-runtime's AddFinalizer](https://github.com/kubernetes-sigs/controller-runtime/blob/984a121dbb8f6273eef55ebb4016e3b3deef1919/pkg/controller/controllerutil/controllerutil.go#L440) be helpful here?

mfrancisc: makes sense. Let's use the AddFinalizer util , thanks for the suggestions! 

Please see https://github.com/codeready-toolchain/toolchain-common/pull/351/commits/b0db03ef3fa2c54d02152e3554f831606266756c

sadlerap (approved):

pkg/test/spacebindingrequest/spacebindingrequest.go
```diff
@@ -0,0 +1,67 @@
...
+
+func WithFinalizer() Option {
+	return func(spaceBindingRequest *toolchainv1alpha1.SpaceBindingRequest) {
+		controllerutil.AddFinalizer(spaceBindingRequest, toolchainv1alpha1.FinalizerName)
+	}
+}
+
+func WithCondition(c toolchainv1alpha1.Condition) Option {
+	return func(spaceBindingRequest *toolchainv1alpha1.SpaceBindingRequest) {
+		spaceBindingRequest.Status.Conditions = append(spaceBindingRequest.Status.Conditions, c)
```
MatousJobanek, line 65: you could use `AddOrUpdateStatusConditions` function for that
https://github.com/codeready-toolchain/toolchain-common/blob/c425b7e6bc32a44874fd49d54e4565e724a8368b/pkg/condition/condition.go#L14-L16
but it's just a detail :-)

MatousJobanek (approved):

pkg/test/spacebindingrequest/spacebindingrequest.go
mfrancisc: Let's use that. See https://github.com/codeready-toolchain/toolchain-common/pull/351/commits/43de5f98a768e0ee40feb6b26371b77d4b0c22bf

Thanks for the suggestion.

PR #354, opened 2024-01-12
Update JWT dependency

alexeykazakov:
Needed for https://github.com/codeready-toolchain/registration-service/pull/391


ranakan19 (approved):
Looks good. Go-lint has pointed out some things to be fixed. Maybe [NewNumericDate](https://github.com/golang-jwt/jwt/blob/v5.2.0/types.go#L38) function should be used to  set NumericDate values.

mfrancisc (approved):
Looks good 👍 

Just few more linting issues apparently.

xcoulon:
> Looks good 👍
> 
> Just few more linting issues apparently.

that's interesting, I did not realize that Golint would also check the dependencies. I think we should disable that since we can't control their contents.

PR #355, opened 2024-01-15
chore: fix linter issues

xcoulon:
also, disable rule on dot-imorts (ie, make always permitted)

### DIFF
--- a/pkg/status/toolchaincluster_test.go
+++ b/pkg/status/toolchaincluster_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/codeready-toolchain/toolchain-common/pkg/test"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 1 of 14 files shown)


pkg/status/toolchaincluster_test.go
```diff
@@ -42,7 +43,7 @@ func TestGetToolchainClusterConditions(t *testing.T) {
 			err := ValidateComponentConditionReady(conditions...)
 
 			// then
-			assert.NoError(t, err)
+			require.NoError(t, err)
```
MatousJobanek, line 46: may I know why there are all these changes for replacing `assert` with `require`?

xcoulon: that was a feedback from ["testifylint"](https://golangci-lint.run/usage/linters/#testifylint):
```
require-error: for error assertions use require (testifylint)
```
If it's an issue for us, then we could probably try to disable this check 🤷‍♂️

MatousJobanek: ok, that makes sense. It's fine - let's keep it there :+1:

MatousJobanek (commented):
Thanks :1st_place_medal:

MatousJobanek (approved):
And approve :smile:

xcoulon:
> And approve 😄

thanks!

PR #357, opened 2024-01-24
Assertions and predicates for the SpaceProvisionerConfig CRD

metlos:
A generic way of declaring predicates that can be used both as assertions (through `AssertThat` function) and as criterions in e2e tests (in `wait.For().FirstThat()` introduced in the associated PR https://github.com/codeready-toolchain/toolchain-e2e/pull/883). 

Additionally, also as an example how to use it, this introduces support functions and predicates for `SpaceProvisionerConfig` unit tests in `host-operator` (https://github.com/codeready-toolchain/host-operator/pull/963) and e2e tests in `toolchain-e2e` (https://github.com/codeready-toolchain/toolchain-e2e/pull/883).


MatousJobanek:
Let's add also the options to set the fields in the spec - what you've done in your e2e PR. But you can expand it by adding options to set "enabled" "thresholds" and other fields

MatousJobanek:
my point is to have functions that will help you to create the SPC objects:
```go
spc := testSpc.NewSpaceProvisionerConfig("spc", test.HostOperatorNs, "cluster1", 
            testSpc.Enabled(), testSpc.WithReadyCondition(), testSpc.MaxNumberOfSpaces(1000), testSpc.MaxMemoryUtilizationPercent(85))
```

MatousJobanek:
just a side note: I would never ever ask you to move one simple function to toolchain-common :smile: 
I just saw the same patterns in your e2e and host-operator PRs and we will definitely need hepler functions that are mentioned above, that's why I proposed moving it to toolchain-common.

MatousJobanek:
but, if you feel that you won't use it in the next changes in both repos and you feel that it's not worth moving it here, then just say - it's fine ;-)

alexeykazakov (approved):

metlos:
@MatousJobanek the upload to codecov failed in this PR a couple of times in a row. Would you have an idea what might be causing this?

alexeykazakov:
> the upload to codecov failed in this PR a couple of times in a row. Would you have an idea what might be causing this?

Yes, it happens from time to time. Codecov uploads are not very reliable on the codecov side. We just have to keep re-running the failed job to get it through.

pkg/test/spaceprovisionerconfig/spaceprovisionerconfig_assertions.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+func NotReadyWithReason(reason string) assertions.Predicate[*toolchainv1alpha1.SpaceProvisionerConfig] {
+	return predicate(func(spc *toolchainv1alpha1.SpaceProvisionerConfig) bool {
+		cond, found := condition.FindConditionByType(spc.Status.Conditions, toolchainv1alpha1.ConditionReady)
+		if !found {
+			return false
+		}
+
+		return cond.Status == corev1.ConditionFalse && cond.Reason == reason
```
MatousJobanek, line 42: similarly
```go
condition.IsFalseWithReason(spc.Status.Conditions, toolchainv1alpha1.ConditionReady, reason)
```

MatousJobanek (approved):
Nice and clean :star_struck:

pkg/test/spaceprovisionerconfig/spaceprovisionerconfig_assertions.go
```diff
@@ -0,0 +1,44 @@
...
+	corev1 "k8s.io/api/core/v1"
+)
+
+type predicate func(*toolchainv1alpha1.SpaceProvisionerConfig) bool
+
+var _ assertions.Predicate[*toolchainv1alpha1.SpaceProvisionerConfig] = predicate(nil)
+
+func (p predicate) Matches(obj *toolchainv1alpha1.SpaceProvisionerConfig) bool {
+	return p(obj)
+}
```
MatousJobanek, line 16: btw, there is one disadvantage of using the same type predicate that returns just a boolean at both places - it's hard to understand why the unit test failed - there is no assertion message, that something was expected but the actual value is different

MatousJobanek: Anyway, let's not try to solve all the problems of the universe in one PR - let's merge it as it is and let's see how this will evolve and is gonna be used

metlos: That is true... There are 2 ways of addressing it:
1. Change `AssertThat` to accept `msgAndArgs ...any` and pass it to the inner `assert.True` so that the caller has some control over the output of the asserts.
2. Change the `Predicate.Matches` to return `(bool, string)` where the second parameter could contain a description of why the predicate didn't match. This could then somehow be incorporated into `AssertThat` and `wait.For().FirstThat()`.

I will update this PR with the former, but I would wait with the latter until we find we actually need it. WDYT?

MatousJobanek: Sounds good :+1:

PR #359, opened 2024-01-30
KSPACE-20: Drop the distinction between host & member ToolchainClusters

fbm3307:
This is related to Drop the distinction between host & member ToolchainClusters
Related PRs:
Host-Operator - https://github.com/codeready-toolchain/host-operator/pull/971
Member-Operator - https://github.com/codeready-toolchain/member-operator/pull/531
Registration-Service - https://github.com/codeready-toolchain/registration-service/pull/402
Toolchain-e2e - https://github.com/codeready-toolchain/toolchain-e2e/pull/893
Sandbox-sre - https://github.com/codeready-toolchain/sandbox-sre/pull/1524


pkg/cluster/cache.go
```diff
@@ -81,23 +79,21 @@ var Ready Condition = func(cluster *CachedToolchainCluster) bool {
 	return IsReady(cluster.ClusterStatus)
 }
 
-func (c *toolchainClusterClients) getCachedToolchainClustersByType(clusterType Type, conditions ...Condition) []*CachedToolchainCluster {
+func (c *toolchainClusterClients) getCachedToolchainClustersByType(conditions ...Condition) []*CachedToolchainCluster {
```
alexeykazakov, line 82: The function name doesn't make much sense now. It's still called `getCachedToolchainClusterByType()` but there is no type anymore.

pkg/cluster/cache.go
```diff
@@ -117,12 +113,12 @@ var HostCluster GetHostClusterFunc = GetHostCluster
 // GetHostCluster returns the kube client for the host cluster from the cache of the clusters
 // and info if such a client exists
 func GetHostCluster() (*CachedToolchainCluster, bool) {
```
alexeykazakov, line 115: Same here. Shouldn't we have a single `GetClusters()` function instead of two `GetHostCluster()` and `GetMemberClusters()`?
Also we need to update the description of the function and stay why a single function is enough (there is an assumption that in the host namespace there are member clusters only and in the member namespace there is a single host cluster only).

alexeykazakov (commented):
We need to make sure that these changes do not break anything.
Please open PRs which uses this updated common for host, member, reg-service, sandbox-sre and e2e tests.
Be aware of that you can't pair multiple PRs with the same e2e test PR though.

metlos (commented):
Apart from the comments that @alexeykazakov made, I think it is going to be more important to take a look at how these functions are used in the operators as well as in the unit and e2e tests and only then decide about how to drop the cluster type. I am afraid that the changes outside of `toolchain-common` are going to be much larger.

pkg/cluster/cache.go
fbm3307: We do require single host cluster to be fetched out for few of the scenarios in member operator . If i make a single func instead of two over here, this same gethostcluster() logic will have to be added in member operator. so the only difference here would be that we transfer the logic from toolchain-common to member-operator. 
So do we remove the dependency of this logic in toolchain-common and add it to member-operator.?

metlos: I'm leaning towards @alexeykazakov's proposal here to just have a single `GetClusters` in `toolchain-common`. We can have `GetHostCluster` function in the `member-operator` that would just use `cache.GetClusters()` and check that there is exactly 1 result and return that.

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -61,33 +61,5 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
-	} else if clusterType != string(cluster.Member) {
-		logger.Info("cluster `type` is not member, skipping cluster role label setting")
-		return nil
-	}
-	clusterRoleLabel := cluster.RoleLabel(cluster.Tenant)
-	if _, exists := toolchainCluster.Labels[clusterRoleLabel]; !exists {
-		logger.Info("setting cluster role label for toolchaincluster", clusterRoleLabel, toolchainCluster.Name)
-		// We use only the label key, the value can remain empty.
-		toolchainCluster.Labels[clusterRoleLabel] = ""
-		if err := r.client.Update(ctx, toolchainCluster); err != nil {
```
alexeykazakov, line 88: Where is this label used right now? I would guess that it's safer to stop using this label first and then stop setting this label. So we can migrate everything smoothly.

alexeykazakov:
> Apart from the comments that @alexeykazakov made, I think it is going to be more important to take a look at how these functions are used in the operators as well as in the unit and e2e tests and only then decide about how to drop the cluster type. I am afraid that the changes outside of toolchain-common are going to be much larger.

Right.
Instead of trying to update everything at once I would try to do it in multiple steps. For example:
- Add a new function to common: `GetClusters()` but do not delete the existing `GetHost()` and `GetMembers()` yet. The old ones are deprecated now basically.
- Create PRs for host and member, reg-service to start using the new function. Please remember to create a paired e2e for every host, member, reg-service PR **if needed**. You can't pair one e2e-tests PR with multiple PRs.
Then we can start merging the PRs one by one without breaking other components and their dependencies.
When all done we can now delete the deprecated functions from common since they are not used anymore.

controllers/toolchaincluster/toolchaincluster_controller.go
fbm3307: To answer @alexeykazakov , it is being used to for setting the placement roles i think, and then we have some [e2e tests](https://github.com/codeready-toolchain/toolchain-e2e/blob/34b67fd73734470c935e9c1105290d196aa50dd2/test/e2e/toolchaincluster_test.go#L146) to make sure the placement roles are set based on the cluster type as @metlos mentioned. So in this case, only the member type is supposed to have cluster role which we are setting here.

fbm3307:
> > Apart from the comments that @alexeykazakov made, I think it is going to be more important to take a look at how these functions are used in the operators as well as in the unit and e2e tests and only then decide about how to drop the cluster type. I am afraid that the changes outside of toolchain-common are going to be much larger.
> 
> Right. Instead of trying to update everything at once I would try to do it in multiple steps. For example:
> 
> * Add a new function to common: `GetClusters()` but do not delete the existing `GetHost()` and `GetMembers()` yet. The old ones are deprecated now basically.
> * Create PRs for host and member, reg-service to start using the new function. Please remember to create a paired e2e for every host, member, reg-service PR **if needed**. You can't pair one e2e-tests PR with multiple PRs.
>   Then we can start merging the PRs one by one without breaking other components and their dependencies.
> * When all done we can now delete the deprecated functions from common since they are not used anymore.

i will  raise another new PR to add a common new function for this .. and then take it ahead.

controllers/toolchaincluster/toolchaincluster_controller.go
fbm3307: updated the logic to accommodate this .

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -16,13 +16,14 @@ import (
...
 // NewReconciler returns a new Reconciler
-func NewReconciler(mgr manager.Manager, namespace string, timeout time.Duration) *Reconciler {
+func NewReconciler(mgr manager.Manager, namespace string, timeout time.Duration, useClusterRL bool) *Reconciler {
 	cacheLog := log.Log.WithName("toolchaincluster_cache")
 	clusterCacheService := cluster.NewToolchainClusterService(mgr.GetClient(), cacheLog, namespace, timeout)
 	return &Reconciler{
 		client:              mgr.GetClient(),
 		scheme:              mgr.GetScheme(),
 		clusterCacheService: clusterCacheService,
+		useClusterRL:        useClusterRL,
```
mfrancisc, line 26: I wasn't sure what the variable purpose was at first, maybe we could rename it to `useClusterRole` or `setClusterRole`. 

Maybe we could make this a string and set it to `cluster.Tenant` in host-operator and leave it empty in member operator. Thus check if the variable is not empty and only then set the role in the toolchaincluster cr.

WDYT?

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -66,18 +68,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
 
 func (r *Reconciler) addToolchainClusterRoleLabelFromType(ctx context.Context, toolchainCluster *toolchainv1alpha1.ToolchainCluster) error {
 	logger := log.FromContext(ctx)
 
-	if clusterType, found := toolchainCluster.Labels[cluster.LabelType]; !found {
-		logger.Info("cluster `type` label not found, unable to add toolchain cluster role label from type")
-		return nil
-	} else if clusterType != string(cluster.Member) {
-		logger.Info("cluster `type` is not member, skipping cluster role label setting")
+	if !r.useClusterRL {
```
mfrancisc, line 77: See my comment above , you could pass `cluster.Tenant` instead of using a bool here, and you could even avoid calling this function `addToolchainClusterRoleLabelFromType` when this `r.setClusterRole` variable is empty.

Also I would rename the function from `addToolchainClusterRoleLabelFromType` to something like `addToolchainClusterRole`.

pkg/cluster/service.go
```diff
@@ -126,16 +123,6 @@ func (s *ToolchainClusterService) addToolchainCluster(log logr.Logger, toolchain
...
-	if cluster.Type == "" {
-		cluster.Type = Member
-	}
-	if cluster.OperatorNamespace == "" {
-		if cluster.Type == Host {
-			cluster.OperatorNamespace = defaultHostOperatorNamespace
-		} else {
-			cluster.OperatorNamespace = defaultMemberOperatorNamespace
-		}
-	}
```
mfrancisc, line 138: just curious - are we deprecating also the `cluster.OperatorNamespace` field ?

pkg/test/verify/cluster.go
```diff
@@ -47,21 +47,9 @@ func AddToolchainClusterAsMember(t *testing.T, functionToVerify FunctionToVerify
...
+
 				// check that toolchain cluster role label tenant was set only on member cluster type
 				require.NoError(t, cl.Get(context.TODO(), client.ObjectKeyFromObject(toolchainCluster), toolchainCluster))
-				expectedToolChainClusterRoleLabel := cluster.RoleLabel(cluster.Tenant)
-				_, found := toolchainCluster.Labels[expectedToolChainClusterRoleLabel]
-				if labels["type"] == string(cluster.Member) {
-					require.True(t, found)
-				} else {
-					require.False(t, found)
-				}
```
mfrancisc, line 64: I think we still need to check that the `cluster.Tenant` role is being set on the toolchaincluster CRs that are in toolchain-host-operator ( thus representing member clusters ). Don't we ?

pkg/test/verify/cluster.go
```diff
@@ -210,16 +191,8 @@ func DeleteToolchainCluster(t *testing.T, functionToVerify FunctionToVerify) {
...
-func Labels(clType cluster.Type, ns, ownerClusterName string) map[string]string {
+func Labels(ns, ownerClusterName string) map[string]string {
 	labels := map[string]string{}
-	if clType != "" {
-		labels["type"] = string(clType)
-		// Set cluster role tenant label only for member type clusters.
-		if clType == cluster.Member {
-			// We use only the label key, the value can remain empty.
-			labels[cluster.RoleLabel(cluster.Tenant)] = ""
-		}
```
mfrancisc, line 221: don't we want to keep setting this role for toolchaincluster representing members ?

mfrancisc (commented):
Nice! 

I have few comments - mainly for me to better understand the motivation behind the changes and long term goal :)

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -66,18 +68,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+		logger.Info("Skiping the cluster role label setting as not running in host cluster")
 		return nil
 	}
```
MatousJobanek, line 80: I would drop this function completely. IIRC, this is a leftover from the time when we introduced the labels and we wanted to populate all existing ToolchainCluster CRs with this label in the already running instances. 
The labels is set when the ToolchainCluster is created for all member CRs https://github.com/codeready-toolchain/toolchain-cicd/blob/34ea85fe21ae05c9bb1a0a2bbdc986b94c8d9e92/scripts/add-cluster.sh#L387-L400 plus we are moving this piece of configuration to SpaceProvisionerConfig anyway.

pkg/cluster/service.go
MatousJobanek, line 138: Let's fail if the operatorNamespace is not set (just to be on the safe side :-) :
```
if cluster.OperatorNamespace == "" {
	return fmt.Errorf("the operatorNamespace is not set in the ToolchainCluster CR")
}
```

pkg/cluster/service_test.go
```diff
@@ -69,66 +69,47 @@ func TestDeleteToolchainClusterWhenDoesNotExist(t *testing.T) {
...
 	cl := test.NewFakeClient(t, m1, m2, host, noise, sec1, sec2, secHost, secNoise)
 
 	t.Run("list members", func(t *testing.T) {
 		// when
-		clusterConfigs, err := cluster.ListToolchainClusterConfigs(cl, m1.Namespace, cluster.Member, time.Second)
+		clusterConfigs, err := cluster.ListToolchainClusterConfigs(cl, m1.Namespace, time.Second)
 
 		// then
 		require.NoError(t, err)
-		require.Len(t, clusterConfigs, 2)
```
MatousJobanek, line 85: Let's update it so it's testing real scenarios:
```go
	require.NoError(t, toolchainv1alpha1.AddToScheme(scheme.Scheme))

	t.Run("list members", func(t *testing.T) {
		// when
		cl := test.NewFakeClient(t, m1, m2, sec1, sec2, secNoise)
		clusterConfigs, err := cluster.ListToolchainClusterConfigs(cl, m1.Namespace, time.Second)

		// then
		require.NoError(t, err)
		require.Len(t, clusterConfigs, 2)
```

pkg/cluster/service_test.go
```diff
@@ -69,66 +69,47 @@ func TestDeleteToolchainClusterWhenDoesNotExist(t *testing.T) {
...
 		// when
-		clusterConfigs, err := cluster.ListToolchainClusterConfigs(cl, m1.Namespace, cluster.Host, time.Second)
+		clusterConfigs, err := cluster.ListToolchainClusterConfigs(cl, m1.Namespace, time.Second)
 
 		// then
 		require.NoError(t, err)
-		require.Len(t, clusterConfigs, 1)
-		verify.AssertClusterConfigThat(t, clusterConfigs[0]).
-			IsOfType(cluster.Host).
+		verify.AssertClusterConfigThat(t, clusterConfigs[1]).
```
MatousJobanek, line 105: the same here:
```go
		// when
		cl := test.NewFakeClient(t, host, secHost, secNoise)
		clusterConfigs, err := cluster.ListToolchainClusterConfigs(cl, m1.Namespace, time.Second)

		// then
		require.NoError(t, err)
		require.Len(t, clusterConfigs, 1)
		verify.AssertClusterConfigThat(t, clusterConfigs[0]).
```

pkg/cluster/service_test.go
```diff
@@ -69,66 +69,47 @@ func TestDeleteToolchainClusterWhenDoesNotExist(t *testing.T) {
...
 			HasName("host").
 			HasOperatorNamespace("toolchain-host-operator").
 			HasOwnerClusterName("hostClusterName").
 			HasAPIEndpoint("http://cluster.com").
 			RestConfigHasHost("http://cluster.com")
 	})
 
-	t.Run("list members when there is none present", func(t *testing.T) {
```
MatousJobanek, line 120: why did you drop this test? we want to test situation when there is no ToolchainCluster CR present, right?

pkg/test/verify/cluster.go
```diff
@@ -47,21 +47,9 @@ func AddToolchainClusterAsMember(t *testing.T, functionToVerify FunctionToVerify
 				require.NoError(t, err)
 				cachedToolchainCluster, ok := cluster.GetCachedToolchainCluster("east")
 				require.True(t, ok)
-				assert.Equal(t, cluster.Member, cachedToolchainCluster.Type)
-				if labels["namespace"] == "" {
-					assert.Equal(t, "toolchain-member-operator", cachedToolchainCluster.OperatorNamespace)
-				} else {
-					assert.Equal(t, labels["namespace"], cachedToolchainCluster.OperatorNamespace)
-				}
```
MatousJobanek, line 55: let's have some verification of the namespace value there

pkg/test/verify/cluster.go
```diff
@@ -96,12 +84,7 @@ func AddToolchainClusterAsHost(t *testing.T, functionToVerify FunctionToVerify)
 				require.NoError(t, err)
 				cachedToolchainCluster, ok := cluster.GetCachedToolchainCluster("east")
 				require.True(t, ok)
-				assert.Equal(t, cluster.Host, cachedToolchainCluster.Type)
-				if labels["namespace"] == "" {
-					assert.Equal(t, "toolchain-host-operator", cachedToolchainCluster.OperatorNamespace)
-				} else {
-					assert.Equal(t, labels["namespace"], cachedToolchainCluster.OperatorNamespace)
-				}
```
MatousJobanek, line 104: the same, we still want to verify the namespace value

pkg/test/verify/cluster.go
```diff
@@ -25,9 +25,9 @@ func AddToolchainClusterAsMember(t *testing.T, functionToVerify FunctionToVerify
 	defer gock.Off()
 	status := test.NewClusterStatus(toolchainv1alpha1.ToolchainClusterReady, corev1.ConditionTrue)
 	memberLabels := []map[string]string{
-		Labels("", "", test.NameHost),
-		Labels(cluster.Member, "", test.NameHost),
-		Labels(cluster.Member, "member-ns", test.NameHost)}
+		Labels("", test.NameHost),
+		Labels("", test.NameHost),
```
MatousJobanek, line 29: looks like a duplicated labels
```suggestion
		Labels("", test.NameHost),
```
and don't forget that we want to fail when the namespace is not set in the label.

MatousJobanek (commented):
Looks good, but please keep an eye on the tests - we still want to verify that the namespace and the type labels are properly taken from the labels and set in the CachedToolchainCluster object

controllers/toolchaincluster/toolchaincluster_controller.go
MatousJobanek: let's drop this field, we don't need it - see my other comments.

controllers/toolchaincluster/toolchaincluster_controller.go
MatousJobanek: as I mentioned here https://github.com/codeready-toolchain/toolchain-common/pull/359#discussion_r1504144684
we should drop the whole `addToolchainClusterRoleLabelFromType` function. The roles should be set when the ToolchainCluster CR is created https://github.com/codeready-toolchain/toolchain-cicd/blob/34ea85fe21ae05c9bb1a0a2bbdc986b94c8d9e92/scripts/add-cluster.sh#L387-L400 
IIRC, this was a function to migrate the already existing CRs.

pkg/cluster/service.go
MatousJobanek: nope, we just don't need to have a logic that would set the namespace label if it's not already there since it should be set when the ToolchainCluster CRs are created https://github.com/codeready-toolchain/toolchain-cicd/blob/34ea85fe21ae05c9bb1a0a2bbdc986b94c8d9e92/scripts/add-cluster.sh#L398

pkg/test/verify/cluster.go
MatousJobanek: Yup, I agree with @mfrancisc. But let's expect that the label is properly set in `cachedToolchainCluster.Labels["namespace"]`, we can assume that the add-cluster.sh script does that.

pkg/test/verify/cluster.go
MatousJobanek: nope, as mentioned in the other comments, this is done in add-cluster.sh script.

controllers/toolchaincluster/toolchaincluster_controller.go
MatousJobanek: As mentioned in the other comments, this logic was there only to populate the already existing ToolchainCluster CRs when we introduced the label and we apparently forgot to remove it.
The placement roles should be set via add-cluster.sh script https://github.com/codeready-toolchain/toolchain-cicd/blob/34ea85fe21ae05c9bb1a0a2bbdc986b94c8d9e92/scripts/add-cluster.sh#L387-L400 - no need to do it here as well.

pkg/cluster/service.go
MatousJobanek: Yeah, right, but only if the OperatorNamespace is not already set. Take a look at this line https://github.com/codeready-toolchain/toolchain-common/blob/d1606659ae07cb3ad6d1d998ebdcacdce783fc62/pkg/cluster/service.go#L208 
We set the OperatorNamespace based on the value of the label, and as I mentioned above, we can rely on the fact that the namespace label should be always set on the ToolchainCluster CR as part of its creation.
In other words, the OperatorNamespace should be always set, and if not then it should fail - see https://github.com/codeready-toolchain/toolchain-common/pull/359#discussion_r1504151698

mfrancisc: Ok got it, I missed the part where it was set from the label. Thanks for the explanation!

pkg/cluster/service_test.go
fbm3307: Yes we do need it, i got confused.. added it back again

pkg/cluster/service.go
fbm3307: This will always throw the error making the unit tests fail as we are not setting the operator namespace here the operator namespace is going to be empty ,  https://github.com/codeready-toolchain/toolchain-common/blob/b32711b41532170d5a241bf61fce363b23b5d16c/controllers/toolchaincluster/healthchecker_test.go#L37-L98 ,  https://github.com/codeready-toolchain/toolchain-common/blob/b32711b41532170d5a241bf61fce363b23b5d16c/controllers/toolchaincluster/toolchaincluster_controller_test.go#L15-L88,  https://github.com/codeready-toolchain/toolchain-common/blob/b32711b41532170d5a241bf61fce363b23b5d16c/pkg/cluster/service_whitebox_test.go#L19-L125,   https://github.com/codeready-toolchain/toolchain-common/blob/b32711b41532170d5a241bf61fce363b23b5d16c/pkg/cluster/service_test.go#L19-L67,  https://github.com/codeready-toolchain/toolchain-common/blob/b32711b41532170d5a241bf61fce363b23b5d16c/pkg/socialevent/socialevent_name_generator_test.go#L24-L34

MatousJobanek: Then just update the unit tests so they match the modified production code. Make sure that the label is set when it's expected to be set.

MatousJobanek: yes, we should fail if the operator namespace is empty. The unit tests should be updated so they match the expected behavior.

pkg/test/verify/cluster.go
```diff
@@ -181,9 +190,7 @@ func UpdateToolchainCluster(t *testing.T, functionToVerify FunctionToVerify) {
 	require.True(t, ok)
 	assert.Equal(t, statusFalse, *cachedToolchainCluster.ClusterStatus)
 	AssertClusterConfigThat(t, cachedToolchainCluster.Config).
-		IsOfType(cluster.Host).
 		HasName("east").
-		HasOperatorNamespace("toolchain-host-operator").
```
mfrancisc, line 186: I think we still want to check the namespace ?

mfrancisc (commented):
Looks good! 

I have some comments/questions related to the tests.

controllers/toolchaincluster/healthchecker_test.go
```diff
@@ -81,7 +85,7 @@ func TestClusterHealthChecks(t *testing.T) {
 		updateClusterStatuses(context.TODO(), "test-namespace", cl)
 
 		// then
-		assertClusterStatus(t, cl, "stable", healthy())
+		//assertClusterStatus(t, cl, "stable", healthy())
```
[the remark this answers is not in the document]
fbm3307: yes right, i missed it

controllers/toolchaincluster/healthchecker_test.go
```diff
@@ -65,14 +69,14 @@ func TestClusterHealthChecks(t *testing.T) {
 		updateClusterStatuses(context.TODO(), "test-namespace", cl)
 
 		// then
-		assertClusterStatus(t, cl, "unstable", notOffline(), unhealthy())
+		//assertClusterStatus(t, cl, "unstable", notOffline(), unhealthy())
```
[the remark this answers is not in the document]
fbm3307: yes

controllers/toolchaincluster/healthchecker_test.go
```diff
@@ -38,7 +38,9 @@ func TestClusterHealthChecks(t *testing.T) {
 		unstable, sec := newToolchainCluster("unstable", "http://unstable.com", toolchainv1alpha1.ToolchainClusterStatus{})
 		notFound, _ := newToolchainCluster("not-found", "http://not-found.com", toolchainv1alpha1.ToolchainClusterStatus{})
 		stable, _ := newToolchainCluster("stable", "http://cluster.com", toolchainv1alpha1.ToolchainClusterStatus{})
```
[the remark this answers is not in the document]
fbm3307: yes right, thanks for pointing it out

pkg/test/verify/cluster.go
fbm3307: yes, missed it somehow, adding it back again

pkg/cluster/cache_whitebox_test.go
```diff
@@ -425,12 +389,8 @@ func TestMultipleActionsInParallel(t *testing.T) {
...
 				defer waitForFinished.Done()
 				latch.Wait()
-				clusters := clusterCache.getCachedToolchainClustersByType(clusterToTest.Type)
-				if len(clusters) == 1 {
-					assert.Equal(t, clusterToTest, clusters[0])
-				} else {
-					assert.Empty(t, clusters)
-				}
+				clusterCache.getCachedToolchainClusters()
+
```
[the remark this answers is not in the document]
fbm3307: My initial understanding was different, i thought since we are not giving type, it will bring in all clusters and finding a single cluster out it was somehow confusing for me.. , but i get your point, adding it

pkg/cluster/cache_whitebox_test.go
```diff
@@ -107,13 +107,13 @@ func TestGetClustersByType(t *testing.T) {
 			// empty cache
 
 			//when
-			clusters := clusterCache.getCachedToolchainClustersByType(Member)
+			clusters := clusterCache.getCachedToolchainClusters()
```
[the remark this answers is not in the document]
fbm3307: i think yes, will remove this test

pkg/cluster/cache_whitebox_test.go
```diff
@@ -206,48 +187,31 @@ func TestGetClustersByType(t *testing.T) {
...
 		})
 	})
 
 	t.Run("get host cluster", func(t *testing.T) {
 
-		// noise
-		member1 := newTestCachedToolchainCluster(t, "cluster-member-1", Member, ready)
-		clusterCache.addCachedToolchainCluster(member1)
-
-		t.Run("not found", func(t *testing.T) {
```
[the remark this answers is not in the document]
fbm3307: yes, just that we dont require noise now, added the rest

pkg/cluster/cache_whitebox_test.go
```diff
@@ -123,68 +123,50 @@ func TestGetClustersByType(t *testing.T) {
...
 			assert.Contains(t, clusters, host)
 		})
 	})
 
 	t.Run("get member clusters", func(t *testing.T) {
-		// noise
-		host := newTestCachedToolchainCluster(t, "cluster-host", Host, ready)
-		clusterCache.addCachedToolchainCluster(host)
-
-		t.Run("not found", func(t *testing.T) {
```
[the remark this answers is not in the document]
fbm3307: just that we dont require noise now, added the rest , my initial understanding was different

pkg/cluster/cache_whitebox_test.go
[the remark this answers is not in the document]
fbm3307: but we do require that generic func here as we are testing multiple actions in parallel, so removing this whole func wont work for the whole test, and keeping a check of length also doesnt work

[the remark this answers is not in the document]
MatousJobanek: I fully agree with @mfrancisc, this test doesn't verify anything - it only calls the function, but doesn't do any assertion on the returned value

pkg/cluster/service_test.go
```diff
@@ -69,47 +69,44 @@ func TestDeleteToolchainClusterWhenDoesNotExist(t *testing.T) {
...
 		// then
 		require.NoError(t, err)
 		require.Len(t, clusterConfigs, 2)
 		verify.AssertClusterConfigThat(t, clusterConfigs[0]).
-			IsOfType(cluster.Member).
 			HasName("east").
 			HasOperatorNamespace("toolchain-member-operator").
 			HasOwnerClusterName("m1ClusterName").
 			HasAPIEndpoint("http://m1.com").
-			ContainsLabel(cluster.RoleLabel(cluster.Tenant)). // the value is not used only the key matters
```
MatousJobanek, line 92: why did you remove the check? Let's adjust the setup of the test and make sure that the RoleLabel is present in the clusterConfig.

pkg/cluster/service_test.go
```diff
@@ -69,47 +69,44 @@ func TestDeleteToolchainClusterWhenDoesNotExist(t *testing.T) {
...
 			RestConfigHasHost("http://m1.com")
 		verify.AssertClusterConfigThat(t, clusterConfigs[1]).
-			IsOfType(cluster.Member).
 			HasName("west").
 			HasOperatorNamespace("toolchain-member-operator").
 			HasOwnerClusterName("m2ClusterName").
 			HasAPIEndpoint("http://m2.com").
-			ContainsLabel(cluster.RoleLabel(cluster.Tenant)). // the value is not used only the key matters
```
MatousJobanek, line 100: the same, we want to verify the Tenant label

pkg/cluster/service_test.go
```diff
@@ -69,47 +69,44 @@ func TestDeleteToolchainClusterWhenDoesNotExist(t *testing.T) {
...
+	host, secHost := test.NewToolchainCluster("host", "secretHost", status, verify.Labels(test.HostOperatorNs, "hostClusterName"))
+	noise, secNoise := test.NewToolchainCluster("noise", "secretNoise", status, verify.Labels(test.MemberOperatorNs, "noiseClusterName"))
 	require.NoError(t, toolchainv1alpha1.AddToScheme(scheme.Scheme))
-	cl := test.NewFakeClient(t, m1, m2, host, noise, sec1, sec2, secHost, secNoise)
 
 	t.Run("list members", func(t *testing.T) {
 		// when
-		clusterConfigs, err := cluster.ListToolchainClusterConfigs(cl, m1.Namespace, cluster.Member, time.Second)
+		cl := test.NewFakeClient(t, m1, m2, sec1, sec2, secNoise)
+		clusterConfigs, err := cluster.ListToolchainClusterConfigs(cl, m1.Namespace, time.Second)
```
MatousJobanek, line 81: I see that you don't include the `noise` object in the fakeClient and you limit the scope of the ToolchainClusters that are present there. TBH, we should have some noise present to test that it filters the ToolchainClusters via the provided namespace. In other words, let's update the `NewToolchainClusterWithEndpoint` & `NewToolchainCluster` functions so they takes the target namespace as a parameter:
```go
	m1, sec1 := test.NewToolchainClusterWithEndpoint("east", test.HostOperatorNs, "secret1", "http://m1.com", status, verify.Labels(test.MemberOperatorNs, "m1ClusterName"))
	m2, sec2 := test.NewToolchainClusterWithEndpoint("west", test.HostOperatorNs, "secret2", "http://m2.com", status, verify.Labels(test.MemberOperatorNs, "m2ClusterName"))
	host, secHost := test.NewToolchainCluster("host", test.MemberOperatorNs, "secretHost", status, verify.Labels(test.HostOperatorNs, "hostClusterName"))
	noise, secNoise := test.NewToolchainCluster("noise", "noise-namespace", "secretNoise", status, verify.Labels(test.MemberOperatorNs, "noiseClusterName"))
	require.NoError(t, toolchainv1alpha1.AddToScheme(scheme.Scheme))
	cl := test.NewFakeClient(t, m1, m2, host, noise, sec1, sec2, secHost, secNoise)

	t.Run("list members", func(t *testing.T) {
		// when
		clusterConfigs, err := cluster.ListToolchainClusterConfigs(cl, m1.Namespace, time.Second)
```

pkg/test/verify/cluster.go
```diff
@@ -25,12 +25,12 @@ func AddToolchainClusterAsMember(t *testing.T, functionToVerify FunctionToVerify
...
 	memberLabels := []map[string]string{
-		Labels("", "", test.NameHost),
-		Labels(cluster.Member, "", test.NameHost),
-		Labels(cluster.Member, "member-ns", test.NameHost)}
+		Labels("", test.NameHost),
+		Labels("member-ns", test.NameHost)}
 	for _, labels := range memberLabels {
 
 		t.Run("add member ToolchainCluster", func(t *testing.T) {
+			labels["namespace"] = test.HostOperatorNs
```
MatousJobanek, line 33: this line 
```go
labels["namespace"] = test.HostOperatorNs
```
as well as the other one
```go
labels["namespace"] = ""
```
in a combination with 
```go
	memberLabels := []map[string]string{
		Labels("", test.NameHost),
		Labels("member-ns", test.NameHost)}
	for _, labels := range memberLabels {
```
doesn't make much sense. At first you create two types of labels (either with or without the namespace set) that you iterate over and then you rewrite the configured namespace in the labels which overrides the preconfigured ones. 
Please do the first or the latter, don't combine them - either drop the for preconfigured array and the `for` loop use some if-else based on the value in the label: 
```go
		t.Run("add member ToolchainCluster", func(t *testing.T) {
			for _, withCA := range []bool{true, false} {
				toolchainCluster, sec := test.NewToolchainCluster("east", "secret", status, labels)
				if withCA {
					toolchainCluster.Spec.CABundle = "ZHVtbXk="
				}
				cl := test.NewFakeClient(t, toolchainCluster, sec)
				service := newToolchainClusterService(t, cl, withCA)
				defer service.DeleteToolchainCluster("east")

				// when
				err := functionToVerify(toolchainCluster, cl, service)

				// then
				if labels["namespace"] != "" {
					require.NoError(t, err)
					cachedToolchainCluster, ok := cluster.GetCachedToolchainCluster("east")
					require.True(t, ok)

					assert.Equal(t, labels["namespace"], cachedToolchainCluster.OperatorNamespace)

					// check that toolchain cluster role label tenant was set only on member cluster type
					require.NoError(t, cl.Get(context.TODO(), client.ObjectKeyFromObject(toolchainCluster), toolchainCluster))
					assert.Equal(t, status, *cachedToolchainCluster.ClusterStatus)
					assert.Equal(t, test.NameHost, cachedToolchainCluster.OwnerClusterName)
					assert.Equal(t, "http://cluster.com", cachedToolchainCluster.APIEndpoint)
				} else {
					require.Error(t, err)
					_, ok := cluster.GetCachedToolchainCluster("east")
					require.False(t, ok)
				}
			}
		})
```

pkg/test/verify/cluster.go
```diff
@@ -47,26 +47,36 @@ func AddToolchainClusterAsMember(t *testing.T, functionToVerify FunctionToVerify
...
-				} else {
-					assert.Equal(t, labels["namespace"], cachedToolchainCluster.OperatorNamespace)
-				}
+
+				assert.Equal(t, labels["namespace"], cachedToolchainCluster.OperatorNamespace)
+
 				// check that toolchain cluster role label tenant was set only on member cluster type
 				require.NoError(t, cl.Get(context.TODO(), client.ObjectKeyFromObject(toolchainCluster), toolchainCluster))
-				expectedToolChainClusterRoleLabel := cluster.RoleLabel(cluster.Tenant)
-				_, found := toolchainCluster.Labels[expectedToolChainClusterRoleLabel]
```
MatousJobanek, line 59: I'm still missing the verification of the role label

pkg/cluster/cache_whitebox_test.go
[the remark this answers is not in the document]
metlos: As far as I understand the logic behind the `ToolchainCluster` objects, there never is a place and time where a `ToolchainCluster` for host would co-exist with `ToolchainCluster` for members.

I.e. host-operator only looks at its namespace for member cluster objects and member operator(s) look at their namespaces for a single host cluster object in each.

The cache is actually keyed merely by the cluster name (which makes sense given the usage restrictions above) so there really isn't any assumption that the `ToolchainCluster` objects would be sourced from multiple namespaces IMHO (because then the name is theoretically not enough to distinguish them).

Therefore, IMHO, this test doesn't make sense anymore. In the past, there may have been different assumptions about the `ToolchainCluster` objects but since we're dropping the type, we have to work with the assumptions above.

pkg/test/verify/cluster.go
```diff
@@ -138,7 +145,7 @@ func AddToolchainClusterFailsBecauseOfEmptySecret(t *testing.T, functionToVerify
 	defer gock.Off()
 	status := test.NewClusterStatus(toolchainv1alpha1.ToolchainClusterReady, corev1.ConditionTrue)
 	toolchainCluster, _ := test.NewToolchainCluster("east", "secret", status,
-		Labels("", "", test.NameHost))
+		Labels("", test.NameHost))
```
MatousJobanek, line 148:
```suggestion
		Labels(test.MemberOperatorNs, test.NameHost))
```

pkg/test/verify/cluster.go
```diff
@@ -162,10 +169,12 @@ func UpdateToolchainCluster(t *testing.T, functionToVerify FunctionToVerify) {
 	defer gock.Off()
 	statusTrue := test.NewClusterStatus(toolchainv1alpha1.ToolchainClusterReady, corev1.ConditionTrue)
 	toolchainCluster1, sec1 := test.NewToolchainCluster("east", "secret1", statusTrue,
-		Labels("", "", test.NameMember))
+		Labels("", test.NameMember))
+	toolchainCluster1.Labels["namespace"] = test.HostOperatorNs
```
MatousJobanek, line 173: the `Labels` function already set the `namespace` label, so you can pass it there right away 
```suggestion
		Labels(test.HostOperatorNs, test.NameMember))
```

pkg/test/verify/cluster.go
```diff
@@ -162,10 +169,12 @@ func UpdateToolchainCluster(t *testing.T, functionToVerify FunctionToVerify) {
...
 	statusFalse := test.NewClusterStatus(toolchainv1alpha1.ToolchainClusterReady, corev1.ConditionFalse)
 	toolchainCluster2, sec2 := test.NewToolchainCluster("east", "secret2", statusFalse,
-		Labels(cluster.Host, "", test.NameMember))
+		Labels("", test.NameMember))
+	toolchainCluster2.Labels["namespace"] = test.HostOperatorNs
```
MatousJobanek, line 177: the same
```suggestion
		Labels(test.HostOperatorNs, test.NameMember))
```

pkg/test/verify/cluster.go
```diff
@@ -194,7 +202,8 @@ func DeleteToolchainCluster(t *testing.T, functionToVerify FunctionToVerify) {
 	defer gock.Off()
 	status := test.NewClusterStatus(toolchainv1alpha1.ToolchainClusterReady, corev1.ConditionTrue)
 	toolchainCluster, sec := test.NewToolchainCluster("east", "sec", status,
-		Labels("", "", test.NameHost))
+		Labels("", test.NameHost))
+	toolchainCluster.Labels["namespace"] = test.MemberOperatorNs
```
MatousJobanek, line 206: the same
```suggestion
		Labels(test.MemberOperatorNs, test.NameHost))
```

pkg/cluster/cache_whitebox_test.go
```diff
@@ -195,7 +189,6 @@ func TestGetClustersByType(t *testing.T) {
 			clusters := GetMemberClusters()
 
 			//then
-			assert.Len(t, clusters, 1)
```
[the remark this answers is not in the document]
fbm3307: adding it

pkg/cluster/cache_whitebox_test.go
```diff
@@ -206,32 +199,27 @@ func TestGetClustersByType(t *testing.T) {
...
-		member3 := newTestCachedToolchainCluster(t, "cluster-3", Member, notReady)
+		member3 := newTestCachedToolchainCluster(t, "cluster-3", notReady)
 		clusterCache.addCachedToolchainCluster(member3)
 
 		t.Run("get only ready member clusters", func(t *testing.T) {
 			//when
 			clusters := GetMemberClusters(Ready)
 
 			//then
-			assert.Len(t, clusters, 2)
```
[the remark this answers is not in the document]
fbm3307: here it will be 3 bcoz, the function getmemberclusters and get host cluster is same it gets all the cluster, based on status here it is giving ready status hence the len will be 3 i shall add it

pkg/cluster/cache_whitebox_test.go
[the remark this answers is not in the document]
fbm3307: okay, i think i got the point now, i shall drop the test "`get clusters by type`" and keep `get host cluster` and `get member clusters` test

pkg/cluster/cache_whitebox_test.go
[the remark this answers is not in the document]
fbm3307: okay shall remove it

pkg/test/verify/cluster.go
fbm3307: Okay right

pkg/test/verify/cluster.go
fbm3307: adding it.

pkg/cluster/service_test.go
fbm3307: Added it

pkg/cluster/service_test.go
fbm3307: added it

pkg/cluster/service.go
```diff
@@ -236,9 +227,9 @@ func IsReady(clusterStatus *toolchainv1alpha1.ToolchainClusterStatus) bool {
 	return false
 }
 
-func ListToolchainClusterConfigs(cl client.Client, namespace string, clusterType Type, timeout time.Duration) ([]*Config, error) {
+func ListToolchainClusterConfigs(cl client.Client, namespace string, timeout time.Duration) ([]*Config, error) {
```
filariow, line 230: we should provide context from the caller and use in the `List` below

```suggestion
func ListToolchainClusterConfigs(ctx context.Context, cl client.Client, namespace string, timeout time.Duration) ([]*Config, error) {
```

pkg/cluster/service.go
```diff
@@ -236,9 +227,9 @@ func IsReady(clusterStatus *toolchainv1alpha1.ToolchainClusterStatus) bool {
...
 	toolchainClusters := &toolchainv1alpha1.ToolchainClusterList{}
-	if err := cl.List(context.TODO(), toolchainClusters, client.InNamespace(namespace), client.MatchingLabels{LabelType: string(clusterType)}); err != nil {
+	if err := cl.List(context.TODO(), toolchainClusters, client.InNamespace(namespace)); err != nil {
```
filariow, line 232:
```suggestion
	if err := cl.List(ctx, toolchainClusters, client.InNamespace(namespace)); err != nil {
```

pkg/cluster/cache_whitebox_test.go
```diff
@@ -392,19 +336,19 @@ func TestRefreshCache(t *testing.T) {
...
 	var latch sync.WaitGroup
 	latch.Add(1)
 	var waitForFinished sync.WaitGroup
 
-	memberCluster := newTestCachedToolchainCluster(t, "memberCluster", Member, ready)
-	hostCluster := newTestCachedToolchainCluster(t, "hostCluster", Host, ready)
 	clusterCache.refreshCache = func() {
-		clusterCache.addCachedToolchainCluster(memberCluster)
-		clusterCache.addCachedToolchainCluster(hostCluster)
+		clusterCache.addCachedToolchainCluster(clusterForTest)
```
mfrancisc, line 347: we should use more than one toolchaincluster resource I guess, so we can ensure it can work with multiple ones using concurrency.

pkg/cluster/cache_whitebox_test.go
```diff
@@ -425,7 +369,7 @@ func TestMultipleActionsInParallel(t *testing.T) {
 			go func() {
 				defer waitForFinished.Done()
 				latch.Wait()
-				clusters := clusterCache.getCachedToolchainClustersByType(clusterToTest.Type)
+				clusters := clusterCache.getCachedToolchainClusters()
```
mfrancisc, line 372: here we should expect: that all the returned toolchainclusters are the in the expected list of toolchaincluster or the list can be empty. WDYT?

pkg/cluster/service_test.go
```diff
@@ -69,30 +69,32 @@ func TestDeleteToolchainClusterWhenDoesNotExist(t *testing.T) {
...
-	host, secHost := test.NewToolchainCluster("host", "secretHost", status, verify.Labels(cluster.Host, test.HostOperatorNs, "hostClusterName"))
-	noise, secNoise := test.NewToolchainCluster("noise", "secretNoise", status, verify.Labels(cluster.Type("e2e"), test.MemberOperatorNs, "noiseClusterName"))
 	require.NoError(t, toolchainv1alpha1.AddToScheme(scheme.Scheme))
+
+	m1, sec1 := test.NewToolchainClusterWithEndpoint("east", test.HostOperatorNs, "secret1", "http://m1.com", status, verify.Labels(test.MemberOperatorNs, "m1ClusterName"))
+	m2, sec2 := test.NewToolchainClusterWithEndpoint("west", test.HostOperatorNs, "secret2", "http://m2.com", status, verify.Labels(test.MemberOperatorNs, "m2ClusterName"))
+	host, secHost := test.NewToolchainCluster("host", test.MemberOperatorNs, "secretHost", status, verify.Labels(test.HostOperatorNs, "hostClusterName"))
+	noise, secNoise := test.NewToolchainCluster("noise", "noise-namespace", "secretNoise", status, verify.Labels(test.MemberOperatorNs, "noiseClusterName"))
+	require.NoError(t, toolchainv1alpha1.AddToScheme(scheme.Scheme))
+	m1.Labels[cluster.RoleLabel(cluster.Tenant)] = ""
```
mfrancisc, line 79: minor thing - but coulnd't we pass those labels to NewToolchainClusterWithEndpoint? I see it gets a map of labels as last parameter.

I'm saying this just because this way I got confused , I thought we are resetting the label value, but than I rememberd that this label only has the key and no value.

pkg/test/verify/cluster.go
```diff
@@ -25,46 +25,41 @@ func AddToolchainClusterAsMember(t *testing.T, functionToVerify FunctionToVerify
...
-
 				// then
-				require.NoError(t, err)
-				cachedToolchainCluster, ok := cluster.GetCachedToolchainCluster("east")
-				require.True(t, ok)
-				assert.Equal(t, cluster.Member, cachedToolchainCluster.Type)
-				if labels["namespace"] == "" {
-					assert.Equal(t, "toolchain-member-operator", cachedToolchainCluster.OperatorNamespace)
-				} else {
+				if labels["namespace"] != "" {
```
filariow, line 45: I'd suggest to split this test in two as it is actually checking 2 different behaviors. Something like:

```go
	t.Run("add member ToolchainCluster with namespace label set", func(t *testing.T) {
		labels := Labels("", test.NameHost)

		for _, withCA := range []bool{true, false} {
			toolchainCluster, sec := test.NewToolchainCluster("east", test.HostOperatorNs, "secret", status, labels)
			if withCA {
				toolchainCluster.Spec.CABundle = "ZHVtbXk="
			}
			toolchainCluster.Labels[cluster.RoleLabel(cluster.Tenant)] = ""
			cl := test.NewFakeClient(t, toolchainCluster, sec)
			service := newToolchainClusterService(t, cl, withCA)
			defer service.DeleteToolchainCluster("east")

			// when
			err := functionToVerify(toolchainCluster, cl, service)

			// then
			require.NoError(t, err)
			cachedToolchainCluster, ok := cluster.GetCachedToolchainCluster("east")
			require.True(t, ok)
			assert.Equal(t, labels["namespace"], cachedToolchainCluster.OperatorNamespace)
			// check that toolchain cluster role label tenant was set only on member cluster type
			require.NoError(t, cl.Get(context.TODO(), client.ObjectKeyFromObject(toolchainCluster), toolchainCluster))
			_, found := toolchainCluster.Labels[cluster.RoleLabel(cluster.Tenant)]
			require.True(t, found)

			assert.Equal(t, status, *cachedToolchainCluster.ClusterStatus)
			assert.Equal(t, test.NameHost, cachedToolchainCluster.OwnerClusterName)
			assert.Equal(t, "http://cluster.com", cachedToolchainCluster.APIEndpoint)
		}
	})

	t.Run("add member ToolchainCluster without namespace label should fail", func(t *testing.T) {
		labels := Labels("member-ns", test.NameHost)

		for _, withCA := range []bool{true, false} {
			toolchainCluster, sec := test.NewToolchainCluster("east", test.HostOperatorNs, "secret", status, labels)
			if withCA {
				toolchainCluster.Spec.CABundle = "ZHVtbXk="
			}
			toolchainCluster.Labels[cluster.RoleLabel(cluster.Tenant)] = ""
			cl := test.NewFakeClient(t, toolchainCluster, sec)
			service := newToolchainClusterService(t, cl, withCA)
			defer service.DeleteToolchainCluster("east")
			// when
			err := functionToVerify(toolchainCluster, cl, service)
			// then
			require.Error(t, err)
			_, ok := cluster.GetCachedToolchainCluster("east")
			require.False(t, ok)
		}
	})
```

filariow (commented):
mostly LGTM, just some optional suggestions

pkg/cluster/service.go
fbm3307: I will take this up in a separate PR

pkg/cluster/service.go
fbm3307: I will take this up in a separate PR

pkg/test/verify/cluster.go
```diff
@@ -75,13 +70,12 @@ func AddToolchainClusterAsHost(t *testing.T, functionToVerify FunctionToVerify)
 	defer gock.Off()
 	status := test.NewClusterStatus(toolchainv1alpha1.ToolchainClusterReady, corev1.ConditionFalse)
 	memberLabels := []map[string]string{
-		Labels(cluster.Host, "", test.NameMember),
-		Labels(cluster.Host, "host-ns", test.NameMember)}
+		Labels(test.MemberOperatorNs, test.NameMember),
```
mfrancisc, line 73: shouldn't this case pass an empty namespace
```suggestion
		Labels("", test.NameMember),
```

pkg/test/verify/cluster.go
```diff
@@ -137,8 +134,8 @@ func AddToolchainClusterFailsBecauseOfEmptySecret(t *testing.T, functionToVerify
 	// given
 	defer gock.Off()
 	status := test.NewClusterStatus(toolchainv1alpha1.ToolchainClusterReady, corev1.ConditionTrue)
-	toolchainCluster, _ := test.NewToolchainCluster("east", "secret", status,
-		Labels("", "", test.NameHost))
+	toolchainCluster, _ := test.NewToolchainCluster("east", test.MemberOperatorNs, "secret", status,
+		Labels(test.MemberOperatorNs, test.NameHost))
```
mfrancisc, line 138: minor - but just for consistency with the secret namespace
```suggestion
	toolchainCluster, _ := test.NewToolchainCluster("east", "test-namespace", "secret", status,
		Labels("test-namespace", test.NameHost))
```

mfrancisc (commented):
Thanks for looking at my comments. I have few minor things

Also I was double checking on the usage of `cluster.GetHostCluster` and  `cluster.GetMemberClusters` and seems like the former just returns one toolchaincluster resource. That's the only difference. So I was wondering if we should have just one generic one ( I think it's something that came out in some other comments in this PR ) . We can do that as a follow up PR if we think it makes sense.

pkg/test/verify/cluster.go
fbm3307: yes.. updated it

pkg/test/verify/cluster.go
fbm3307: no the secret namespace  should be changed to "test.MemberOperatorNs" , i have updated it now

pkg/cluster/cache_whitebox_test.go
fbm3307: i think this will be dependent on what we are passing at "clusterCache.addCachedToolchainCluster(clusterForTest)"  , here i removed  the other types bcoz, it becomes difficult to test through member and host, as the function here would bring in all the clusters, we can do if else but that wouldnt be so good, i think to test whether the function getCachedToolchainClusters() works good, passing one is good.. your thoughts..?

pkg/cluster/cache_whitebox_test.go
mfrancisc: We can keep this for follow up PRs, let's maybe think about this later as I've just discussed with Matous, it may be tricky to handle such a scenario.

pkg/cluster/cache_whitebox_test.go
```diff
@@ -392,19 +336,19 @@ func TestRefreshCache(t *testing.T) {
...
-	hostCluster := newTestCachedToolchainCluster(t, "hostCluster", Host, ready)
 	clusterCache.refreshCache = func() {
-		clusterCache.addCachedToolchainCluster(memberCluster)
-		clusterCache.addCachedToolchainCluster(hostCluster)
+		clusterCache.addCachedToolchainCluster(clusterForTest)
+
 	}
 
-	for _, clusterToTest := range []*CachedToolchainCluster{memberCluster, hostCluster} {
+	for _, clusterToTest := range []*CachedToolchainCluster{clusterForTest} {
```
mfrancisc, line 351: I think we can drop this for loop right ? If it contains just one element in list.

pkg/cluster/cache_whitebox_test.go
mfrancisc: ok, makes sense if we use just one toolchaincluster resource. Thanks for explaining that!

pkg/cluster/service_test.go
fbm3307: updated, makes sense

fbm3307:
> Also I was double checking on the usage of `cluster.GetHostCluster` and `cluster.GetMemberClusters` and seems like the former just returns one toolchaincluster resource. That's the only difference. So I was wondering if we should have just one generic one ( I think it's something that came out in some other comments in this PR ) . We can do that as a follow up PR if we think it makes sense.

Yes i intend to do that in separate follow-up PR once this is merged

controllers/toolchaincluster/healthchecker_test.go
```diff
@@ -124,7 +122,7 @@ func withStatus(conditions ...toolchainv1alpha1.ToolchainClusterCondition) toolc
 }
 
 func newToolchainCluster(name, apiEndpoint string, status toolchainv1alpha1.ToolchainClusterStatus) (*toolchainv1alpha1.ToolchainCluster, *corev1.Secret) {
-	toolchainCluster, secret := test.NewToolchainClusterWithEndpoint(name, "secret", apiEndpoint, status, map[string]string{})
+	toolchainCluster, secret := test.NewToolchainClusterWithEndpoint(name, "test-namespace", "secret", apiEndpoint, status, map[string]string{"namespace": "test-namespace"})
```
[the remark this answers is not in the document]
fbm3307: done

pkg/test/verify/cluster.go
fbm3307: updated it

controllers/toolchaincluster/healthchecker_test.go
```diff
@@ -19,6 +19,7 @@ import (
 func TestClusterHealthChecks(t *testing.T) {
 	// given
 	defer gock.Off()
+	ons := "test-namespace"
```
MatousJobanek, line 22: of `ons` stands for "operator namespace', then I'm afraid that you mix two values together. 
operator namespace is the namespace where the operator that ToolchainCluster represents is installed in - this is stored as a label 
but what you are setting here is the namespace where the ToolchainCluster is created.
so could you please just rename the variable to something more meaningful? eg:
* `tcNs`
* `toolchainClusterNs`

It's just a detail, but it may confuse people.

controllers/toolchaincluster/healthchecker_test.go
```diff
@@ -123,8 +122,8 @@ func withStatus(conditions ...toolchainv1alpha1.ToolchainClusterCondition) toolc
 	}
 }
 
-func newToolchainCluster(name, apiEndpoint string, status toolchainv1alpha1.ToolchainClusterStatus) (*toolchainv1alpha1.ToolchainCluster, *corev1.Secret) {
-	toolchainCluster, secret := test.NewToolchainClusterWithEndpoint(name, "secret", apiEndpoint, status, map[string]string{})
+func newToolchainCluster(name, ons string, apiEndpoint string, status toolchainv1alpha1.ToolchainClusterStatus) (*toolchainv1alpha1.ToolchainCluster, *corev1.Secret) {
+	toolchainCluster, secret := test.NewToolchainClusterWithEndpoint(name, ons, "secret", apiEndpoint, status, map[string]string{"namespace": "test-namespace"})
```
MatousJobanek, line 126: this is another place where `ons` should be renamed

pkg/test/cluster.go
```diff
@@ -4,19 +4,19 @@ import (
...
 
 const (
 	NameHost   = "dsaas"
 	NameMember = "east"
 )
 
-func NewToolchainCluster(name, secName string, status toolchainv1alpha1.ToolchainClusterStatus, labels map[string]string) (*toolchainv1alpha1.ToolchainCluster, *corev1.Secret) {
-	return NewToolchainClusterWithEndpoint(name, secName, "http://cluster.com", status, labels)
+func NewToolchainCluster(name, ons, secName string, status toolchainv1alpha1.ToolchainClusterStatus, labels map[string]string) (*toolchainv1alpha1.ToolchainCluster, *corev1.Secret) {
+	return NewToolchainClusterWithEndpoint(name, ons, secName, "http://cluster.com", status, labels)
```
MatousJobanek, line 16: this one too

pkg/test/cluster.go
```diff
@@ -25,7 +25,7 @@ func NewToolchainClusterWithEndpoint(name, secName, apiEndpoint string, status t
 	secret := &corev1.Secret{
 		ObjectMeta: v1.ObjectMeta{
 			Name:      secName,
-			Namespace: "test-namespace",
+			Namespace: ons,
```
MatousJobanek, line 28: this one as well

pkg/test/cluster.go
```diff
@@ -43,7 +43,7 @@ func NewToolchainClusterWithEndpoint(name, secName, apiEndpoint string, status t
 		},
 		ObjectMeta: v1.ObjectMeta{
 			Name:      name,
-			Namespace: "test-namespace",
+			Namespace: ons,
```
MatousJobanek, line 46: this one as well - here you can see the usage - it's not set as a label, but as the location namespace of the ToolchainCluster CR

MatousJobanek (commented):
Looks good, let's just rename one variable to not confuse people

controllers/toolchaincluster/healthchecker_test.go
fbm3307: Okay i agree, changed it

controllers/toolchaincluster/healthchecker_test.go
fbm3307: done

pkg/test/cluster.go
fbm3307: done

pkg/test/cluster.go
fbm3307: done

fbm3307:
> Looks good, let's just rename one variable to not confuse people

@MatousJobanek right, updated it. 👍

MatousJobanek (approved):
cool, thanks for the extra changes. :rocket: 
Let's start merging the PRs :+1: I believe that you have approval for all of them, just don't forget to update the dependency to toolchain-common as soon as you merge this PR

PR #360, opened 2024-01-30
Marking the AssertThat as a test helper method

metlos:
Marking the AssertThat as a helper method so that error reporting skips it and reports the file position in the actual test.


alexeykazakov (approved):
Nice. I didn't know about this.

PR #364, opened 2024-02-29
Include the provisioning configuration in the CachedToolchainCluster

metlos:
Because the provisioning decisions and toolchain cluster configuration are done using the combination of `ToolchainConfig` and `CachedToolchainCluster` objects, this adds the provisioning configuration (loaded from `SpaceProvisionerConfig` objects) to the `CachedToolchainCluster`. 

Doing it this way, we don't have to modify much of the logic elsewhere (mainly in the tests) - we just move the configuration done with `ToolchainConfig` to configuration done with `CachedToolchainCluster`. These two things usually coincide in the codebase so the change is easy.

Note that we keep the default memory usage threshold in the `ToolchainConfig` because there is no place for it in the `SpaceProvisionerConfig` "framework". We can think about dropping it in some future PR.

Note that this is WIP. There still some test failures in the e2e tests and therefore some debugging log statements sprinkled here and there in this code, too.

JIRA issue: https://issues.redhat.com/browse/KSPACE-46

Related PRs:
host-operator: https://github.com/codeready-toolchain/host-operator/pull/977
toolchain-e2e: https://github.com/codeready-toolchain/toolchain-e2e/pull/905

### DIFF
--- a/controllers/toolchaincluster/toolchaincluster_controller.go
+++ b/controllers/toolchaincluster/toolchaincluster_controller.go
@@ -19,9 +19,14 @@ import (
 func NewReconciler(mgr manager.Manager, namespace string, timeout time.Duration) *Reconciler {
 	cacheLog := log.Log.WithName("toolchaincluster_cache")
 	clusterCacheService := cluster.NewToolchainClusterService(mgr.GetClient(), cacheLog, namespace, timeout)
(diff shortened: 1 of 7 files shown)


metlos:
Closing in favor of https://github.com/codeready-toolchain/host-operator/pull/991 and associated PRs.

PR #365, opened 2024-03-04
feat: add LoadObjectsFromEmbedFS functionality

mfrancisc:
Add functionality for rendering k8s templates from `embed.FS` filesystems. 
This will be used, in the first place, by host-operator and member-operator to apply the toolchaincluster resources ( SA, Role, RoleBindings) that are now applied by the `add-cluster.sh` script.

See: https://github.com/codeready-toolchain/toolchain-cicd/blob/34ea85fe21ae05c9bb1a0a2bbdc986b94c8d9e92/scripts/add-cluster.sh#L36

Jira: https://issues.redhat.com/browse/KSPACE-7

Related PRs:
- host-operator: https://github.com/codeready-toolchain/host-operator/pull/974
- member-operator: https://github.com/codeready-toolchain/member-operator/pull/537


pkg/template/testTemplates/member/cluster-role.yaml
```diff
@@ -0,0 +1,11 @@
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: toolchaincluster-member-toolchain-member-operator-toolchaincluster
```
MatousJobanek, line 4: we don't need such a crazy name for the dummy templates

pkg/template/template.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/client-go/kubernetes/scheme"
+)
+
+// TemplateVariables contains all the available variables that are supported by the templates
+type TemplateVariables struct {
```
MatousJobanek, line 18: the package name is `template` so you could rename the type by cutting the prefix to:
```suggestion
type Variables struct {
```

pkg/template/template_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/codeready-toolchain/toolchain-common/pkg/template"
+	"github.com/codeready-toolchain/toolchain-common/pkg/test"
+	"github.com/stretchr/testify/require"
+)
+
+//go:embed testTemplates/*
+var EFS embed.FS
+
+//go:embed testTemplates/host/*
```
MatousJobanek, line 15: we don't use camel-case for the folder/package names, if I'm not mistaken

pkg/template/template_test.go
```diff
@@ -0,0 +1,53 @@
...
+var hostFS embed.FS
+
+func TestLoadObjectsFromEmbedFS(t *testing.T) {
```
MatousJobanek, line 18: could you please add test-cases that validate the exact content of the objects that are in the templates?
These are dummy templates with values that won't change often (compared to what we have in host/member-operator repos), so we can afford hard-coding the expected values for the given kinds in the tests.

pkg/template/template_test.go
```diff
@@ -0,0 +1,53 @@
...
+	t.Run("loads objects recursively from all subdirectories", func(t *testing.T) {
+		// when
+		objects, err := template.LoadObjectsFromEmbedFS(&EFS, &template.TemplateVariables{Namespace: test.HostOperatorNs})
+		// then
+		require.NoError(t, err)
+		require.NotNil(t, objects)
+		// we expect to have 4 objects loaded from the test templates folders
+		require.Equal(t, 4, len(objects), "invalid number of expected objects")
```
MatousJobanek, line 26: could you verify that the number of objects is different for the host & member folders?

MatousJobanek (commented):
Thanks, but let's try to improve the unit tests.

pkg/template/template.go
mfrancisc: Good point, done in https://github.com/codeready-toolchain/toolchain-common/pull/365/commits/f08be69fbfcb8403531f12252c2480366dcffb4e

pkg/template/template_test.go
mfrancisc: Makes sense, please take a look at https://github.com/codeready-toolchain/toolchain-common/pull/365/commits/879f02df98cf7ae4da15c256c8551e1de847c026 and let me know if this could be enough to cover your suggestion.

MatousJobanek (approved):
Nice :+1:

PR #372, opened 2024-03-14
KSPACE-43: Adding common GetClusters func

fbm3307:
This is a Follow-up PR for #359,

- Adding a new common function `GetClusters()` to fetch the clusters from cache as we have dropped the cluster type
- Updated the related function to use the common func instead of `GetHostCluster()` and `GetMemberClusters()`
- Updated the unit test cases as per the given change
- Keeping the `GetHostCluster()` and `GetMemberClusters()` for now and delete them once their dependencies are removed in host, member, registration-service .
- The PRs in other repos which will use this GetCluster() and their paired e2e PRs 
     - Host-Operator - https://github.com/codeready-toolchain/host-operator/pull/990
          - Paired e2e for host Toolchain-E2e - https://github.com/codeready-toolchain/toolchain-e2e/pull/921
     - Registration Service - https://github.com/codeready-toolchain/registration-service/pull/413
          - Paired e2e for host Toolchain-E2e - https://github.com/codeready-toolchain/toolchain-e2e/pull/922
     - Member-Operator - https://github.com/codeready-toolchain/member-operator/pull/543
          - Paired e2e for host Toolchain-E2e - https://github.com/codeready-toolchain/toolchain-e2e/pull/923

### DIFF
--- a/pkg/cluster/cache.go
+++ b/pkg/cluster/cache.go
@@ -103,6 +103,24 @@ func GetCachedToolchainCluster(name string) (*CachedToolchainCluster, bool) {
 	return clusterCache.getCachedToolchainCluster(name, true)
 }
 
+// GetClustersFunc a func that returns all the clusters from the cache
+type GetClustersFunc func(conditions ...Condition) []*CachedToolchainCluster
+
+// Clusters the func to retrieve all the clusters
+var Clusters GetClustersFunc = GetClusters
+
+// GetClusters returns the kube clients for all the clusters from the cache of the clusters
+func GetClusters(conditions ...Condition) []*CachedToolchainCluster {
+	clusters := clusterCache.getCachedToolchainClusters(conditions...)
+	if len(clusters) == 0 {
+		if clusterCache.refreshCache != nil {
+			clusterCache.refreshCache()
+		}
+		clusters = clusterCache.getCachedToolchainClusters(conditions...)
+	}
+	return clusters
+}
+
 // GetHostClusterFunc a func that returns the Host cluster from the cache,
 // along with a bool to indicate if there was a match or not
 type GetHostClusterFunc func() (*CachedToolchainCluster, bool)
--- a/pkg/cluster/cache.go
+++ b/pkg/cluster/cache.go
@@ -103,14 +103,14 @@ func GetCachedToolchainCluster(name string) (*CachedToolchainCluster, bool) {
 	return clusterCache.getCachedToolchainCluster(name, true)
 }
 
-// GetClustersFunc a func that returns all the clusters from the cache
-type GetClustersFunc func(conditions ...Condition) []*CachedToolchainCluster
+// GetToolchainClustersCachedFunc a func that returns all the clusters from the cache
+type GetToolchainClustersCachedFunc func(conditions ...Condition) []*CachedToolchainCluster
 
-// Clusters the func to retrieve all the clusters
-var Clusters GetClustersFunc = GetClusters
+// ToolchainClustersCached the func to retrieve all the clusters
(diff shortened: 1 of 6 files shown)


pkg/status/toolchaincluster.go
```diff
@@ -19,18 +19,20 @@ const (
 
 // ToolchainClusterAttributes required attributes for obtaining ToolchainCluster status
 type ToolchainClusterAttributes struct {
-	GetClusterFunc func() (*cluster.CachedToolchainCluster, bool)
-	Period         time.Duration
-	Timeout        time.Duration
+	GetClustersFunc cluster.GetClustersFunc
```
MatousJobanek, line 22: the purpose and the usage of the `ToolchainClusterAttributes` type and `GetToolchainClusterConditions` function doesn't match with the `cluster.GetClustersFunc` - this is intended to be used for one ToolchainCluster, but for many. I saw that you used 
```go
toolchainCluster := toolchainClusters[0]
```
to get the first one, but you never know if you are supposed to use the first one, second, or third.
I would rather either keep the 
```go
GetClusterFunc func() (*cluster.CachedToolchainCluster, bool)
```
in the same format is it was before and let the caller to pass the ToolchainCluster - something like this:
```go
	attributes := status.ToolchainClusterAttributes{
		GetClusterFunc: func() (*cluster.CachedToolchainCluster, bool) {
			clusters := cluster.GetClustersFunc()
			return clusters[0], len(clusters) != 0
		},
```
or just change the field so it doesn't take a func but rather a `*cluster.CachedToolchainCluster` directly

pkg/cluster/cache.go
```diff
@@ -103,6 +103,24 @@ func GetCachedToolchainCluster(name string) (*CachedToolchainCluster, bool) {
...
+			clusterCache.refreshCache()
+		}
+		clusters = clusterCache.getCachedToolchainClusters(conditions...)
+	}
+	return clusters
+}
+
 // GetHostClusterFunc a func that returns the Host cluster from the cache,
 // along with a bool to indicate if there was a match or not
 type GetHostClusterFunc func() (*CachedToolchainCluster, bool)
```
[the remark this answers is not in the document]
MatousJobanek: ok, I see that you wrote 

> Keeping the GetHostCluster() and GetMemberClusters() for now and delete them once their dependencies are removed in host, member, registration-service .

but let's delete it right away in this PR, so we are sure that we didn't miss any usage anywhere

pkg/cluster/cache.go
```diff
@@ -103,6 +103,24 @@ func GetCachedToolchainCluster(name string) (*CachedToolchainCluster, bool) {
...
 }
 
+// GetClustersFunc a func that returns all the clusters from the cache
+type GetClustersFunc func(conditions ...Condition) []*CachedToolchainCluster
+
+// Clusters the func to retrieve all the clusters
+var Clusters GetClustersFunc = GetClusters
+
+// GetClusters returns the kube clients for all the clusters from the cache of the clusters
+func GetClusters(conditions ...Condition) []*CachedToolchainCluster {
```
MatousJobanek, line 113: one more thing, `GetClusters` is pretty generic and can be confusing. How about:
* GetToolchainClusters
* GetCachedToolchainClusters

so it's clear which clusters we are talking about

fbm3307: There is already a function named as `GetCachedToolchainClusters` , but to avoid confusion and for better understanding updating it to `GetToolchainClustersCachedFunc`

fbm3307:
Closing this PR as Adding the Common Getcluster Func is not required now

PR #373, opened 2024-03-14
feat: create toolchancluster resources controller

mfrancisc:
This PR introduces a new common controller `toolchaincluster_resources_controller`. 
This controller will run in both host and member operator, and it will apply a set of manifests related to the toolchaincluster CRs ( SAs, Role/RoleBindings ... ) .

The controller watches for toolchaincluster CRs , and upon reconcile it applies/creates the given manifests from the embed.FS folder. To each resource it adds a few labels just to keep track of the ownership/relationshifp with the toolchaincluster CR that uses those resources.


controllers/toolchaincluster_resources/toolchaincluster_resources_controller.go
```diff
@@ -0,0 +1,68 @@
...
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&toolchainv1alpha1.ToolchainCluster{}).
+		Complete(r)
```
mfrancisc, line 23: TBH I'm not 100% sure what this controller should watch for. Since it's suppose to be a generic controller that can apply potentially any kind of objects from manifests

controllers/toolchaincluster_resources/toolchaincluster_resources_controller.go
```diff
@@ -0,0 +1,68 @@
...
+
+// Reconciler reconciles a ToolchainCluster object
+type Reconciler struct {
+	client    runtimeclient.Client
+	scheme    *runtime.Scheme
+	templates *embed.FS
+}
+
+//+kubebuilder:rbac:groups=rbac.authorization.k8s.io;authorization.openshift.io,resources=rolebindings;roles;clusterroles;clusterrolebindings,verbs=*
+//+kubebuilder:rbac:groups="",resources=serviceaccounts,verbs=get;list;watch;create;update;patch;delete
```
mfrancisc, line 34: same here concern here about the permissions needed to apply the objects

controllers/toolchaincluster_resources/toolchaincluster_resources_controller.go
```diff
@@ -0,0 +1,68 @@
...
+
+	// read the template fs
+	allObjects, err := template.LoadObjectsFromEmbedFS(r.templates, &template.Variables{Namespace: request.Namespace})
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	// apply all the objects and add toolchaincluster as owner reference
+	newLabels := map[string]string{
+		toolchainv1alpha1.ProviderLabelKey: toolchainv1alpha1.ProviderLabelValue,
+		toolchainv1alpha1.OwnerLabelKey:    toolchainCluster.GetName(),
```
mfrancisc, line 64: I thought about adding the toolchaincluster as owner of these resources , so that we can identify who created them.

controllers/toolchaincluster_resources/toolchaincluster_resources_controller.go
MatousJobanek: There are two options:
1. either don't create the whole `Reconciler` in toolchain-common, but let each caller (host/member) operator define what they want to watch. Toolchain-common could contain either some generic logic that would create the `Reconciler` based on the given values, or just the content of the `Reconcile` function
2. or parse the content of the template, go over all kinds that are in the template and add watch for each of the kind.

controllers/toolchaincluster_resources/toolchaincluster_resources_controller.go
MatousJobanek: TBH, I'm not sure if the kubebuilder reads the content from the dependencies. In other words, I don't think that this will have any result on the manifest generated by the kubebuilder. 
This should be rather placed in the operator code - we do that also with the ToolchainCluster controller (and other stuff), the marker is added to the main.go:  https://github.com/codeready-toolchain/host-operator/blob/f27b2b90c5132b3346a8eb13254307fb209c46d4/main.go#L92-L94

controllers/toolchaincluster_resources/testdata/membersa.yaml
```diff
@@ -0,0 +1,19 @@
...
+metadata:
+  name: toolchaincluster-member
+  namespace: {{.Namespace}}
+rules:
+- apiGroups:
+  - toolchain.dev.openshift.com
+  resources:
+  - "*"
+  verbs:
+  - "*"
```
MatousJobanek, line 19: There are already some templates in https://github.com/codeready-toolchain/toolchain-common/tree/master/pkg/template/testdata so couldn't we use them?

controllers/toolchaincluster_resources/toolchaincluster_resources_controller.go
```diff
@@ -0,0 +1,68 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&toolchainv1alpha1.ToolchainCluster{}).
```
MatousJobanek, line 22: The manifests should be applied in the cluster even before the very first ToolchainCluster CR is created in the namespace. The creation of the ToolchainCluster CR relies on the existence of these resources.
I would watch ServiceAccounts as the main driver for this controller. We know that there is gonna be always at least one SA in the namespace, so we know that the controller will be always triggered when the operator is started for the first time.

Watch for ToolchainCluster CRs will be added later as part of this story [KUBESAW-44](https://issues.redhat.com/browse/KUBESAW-44)

controllers/toolchaincluster_resources/toolchaincluster_resources_controller.go
```diff
@@ -0,0 +1,68 @@
...
+	// Fetch the ToolchainCluster instance
+	toolchainCluster := &toolchainv1alpha1.ToolchainCluster{}
+	err := r.client.Get(ctx, request.NamespacedName, toolchainCluster)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
+	}
```
MatousJobanek, line 54: As mentioned above, the controller doesn't care about ToolchainCluster for now. It will become ToolchainCluster aware later [KUBESAW-44](https://issues.redhat.com/browse/KUBESAW-44), but not that it would be driven by the CR, it would create the CRs instead.

pkg/client/client.go
```diff
@@ -261,3 +262,54 @@ func MergeAnnotations(toolchainObject client.Object, newAnnotations map[string]s
...
+// ApplyUnstructuredObjects applies the given Unstructured objects on the cluster.
+func ApplyUnstructuredObjects(ctx context.Context, cl client.Client, unstructuredObjects []*unstructured.Unstructured, newLabels map[string]string) (bool, error) {
+	applyClient := NewApplyClient(cl)
+	anyApplied := false
+
+	for _, unstructuredObj := range unstructuredObjects {
+		var object = unstructuredObj.DeepCopyObject().(client.Object)
+		if err := runtime.DefaultUnstructuredConverter.FromUnstructured(unstructuredObj.Object, object); err != nil {
+			return false, err
+		}
```
MatousJobanek, line 275: What do we need this for? What does this do?

pkg/client/client.go
```diff
@@ -261,3 +262,54 @@ func MergeAnnotations(toolchainObject client.Object, newAnnotations map[string]s
...
+		// Special handling of ServiceAccounts is required because if a ServiceAccount is reapplied when it already exists, it causes Kubernetes controllers to
+		// automatically create new Secrets for the ServiceAccounts. After enough time the number of Secrets created will hit the Secrets quota and then no new
+		// Secrets can be created. To prevent this from happening, we fetch the already existing SA, update labels and annotations only, and then call update using the same object (keeping the refs to secrets).
+		if strings.EqualFold(object.GetObjectKind().GroupVersionKind().Kind, "ServiceAccount") {
```
MatousJobanek, line 279: do you think that we could move it directly to the `applyObject` function?

controllers/toolchaincluster_resources/toolchaincluster_resources_controller.go
```diff
@@ -0,0 +1,68 @@
...
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	// apply all the objects and add toolchaincluster as owner reference
+	newLabels := map[string]string{
+		toolchainv1alpha1.ProviderLabelKey: toolchainv1alpha1.ProviderLabelValue,
+		toolchainv1alpha1.OwnerLabelKey:    toolchainCluster.GetName(),
+	}
+	_, err = applycl.ApplyUnstructuredObjects(ctx, r.client, allObjects, newLabels)
+	return reconcile.Result{}, err
```
MatousJobanek, line 67: We should also take care of the deletion right? For example if one item is removed from the template or is renamed.
We could simply list all objects (of the given kind) with some added label `provider: toolchaincluster-resources-controller` or something like that, and delete all that are not in the template. 

This can be done in a separate PR, though.

MatousJobanek (commented):
It's not specific for your PR per se, but rather to the logic that is already there. 
We should take a look at libraries that could take care of the updates instead of writing our own create-or-update logic. I know that there is `CreateOrUpdate` by the controller-runtime library, not sure if it's something we could use in a generic way. Kubectl has its own logic for apply to, ArgoCD, and Helm charts most likely as well - let's keep whatever is there as it is for now, and take a look at our options as part of the NSTemplateTier decoupling effort.

controllers/toolchaincluster_resources/toolchaincluster_resources_controller.go
mfrancisc: Thanks for your suggestion, I'm implementing 2. in https://github.com/codeready-toolchain/toolchain-common/pull/373/commits/6c014d7e44734b2694b2c3f6fd0b3bbb695e7490

controllers/toolchaincluster_resources/testdata/membersa.yaml
mfrancisc: I've updated the tests to reuse the same templates in https://github.com/codeready-toolchain/toolchain-common/pull/373/commits/a6e5f881e6cf1fb6adba401619de36bf701aed82

mfrancisc: I haven't done this initially because I exposing the fs variables would make Go to include those in the final binary ( unless there's some other way to do it) . But I think it's not a big issue.

controllers/toolchaincluster_resources/toolchaincluster_resources_controller.go
mfrancisc: Yup, I've updated the code to watch for SA and all the other types from the templates. Thanks for your pointers!

controllers/toolchaincluster_resources/toolchaincluster_resources_controller.go
mfrancisc: Got it! Removed https://github.com/codeready-toolchain/toolchain-common/pull/373/commits/b5015a53eb80030dc9c79e03aad19c82073fe6d5

pkg/client/client.go
mfrancisc: Yeah, this is a left over 🤦‍♂️ 

Removed https://github.com/codeready-toolchain/toolchain-common/pull/373/commits/b1606562340e1711c01b821400d1bb1e55e06c9d

thanks

pkg/client/client.go
mfrancisc: Good point, I've implemented the functionality in `applyObject` in https://github.com/codeready-toolchain/toolchain-common/pull/373/commits/8ee46aacee2e9c2a0bb3a24f7c0027556cc34dfc . 

It's not using the existing object but just copying the secrets and annotations, please take a look and let me know if I've missed anything.

controllers/toolchaincluster_resources/toolchaincluster_resources_controller.go
mfrancisc: Yeah I though about that, and we can add it as a follow up PR, in order to keep this smaller, if that makes sense.

controllers/toolchaincluster_resources/testdata/membersa.yaml
mfrancisc: I've moved those back to the test directory in https://github.com/codeready-toolchain/toolchain-common/pull/373/commits/8ee46aacee2e9c2a0bb3a24f7c0027556cc34dfc , as discussed , so that the templates are not bundled with the production binary. 

but maybe there's some other way to avoid duplicating those 🤔

controllers/toolchaincluster_resources/toolchaincluster_resources_controller.go
```diff
@@ -0,0 +1,68 @@
...
+func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
+	reqLogger := log.FromContext(ctx)
+	reqLogger.Info("Reconciling ToolchainCluster resources controller")
+	// check for required templates FS directory
+	if r.templates == nil {
+		return reconcile.Result{}, fmt.Errorf("no templates FS configured")
+	}
+
+	// read the template fs
+	allObjects, err := template.LoadObjectsFromEmbedFS(r.templates, &template.Variables{Namespace: request.Namespace})
```
MatousJobanek, line 58: btw, you already read, parse, and process the template when you are initializing the controller, we pass only one parameter (namespace) that won't change during the execution, so you could store the retrieved objects in the `Reconciler` type and avoid doing the same template process over and over again.
WDYT?

mfrancisc: that's a good point. It will avoid to reload the templates every time ( even if this is done in memory , so it will be very quick ). 
The only "down side" of this is that we don't actually unit test the `SetupWithManager` function right ? So the loading part will be uncovered. But I think that's fine. 

WDYT?

MatousJobanek: yeah, you already test the `template.LoadObjectsFromEmbedFS` function in a unit-test, so we should be fine :+1:

controllers/toolchaincluster_resources/toolchaincluster_resources_controller.go
```diff
@@ -0,0 +1,68 @@
...
+func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, operatorNamespace string) error {
+	build := ctrl.NewControllerManagedBy(mgr).
+		For(&v1.ServiceAccount{})
+
+	// add watcher for all kinds from given templates
+	allObjects, err := template.LoadObjectsFromEmbedFS(r.templates, &template.Variables{Namespace: operatorNamespace})
+	if err != nil {
+		return err
+	}
+	mapToOwnerByLabel := handler.EnqueueRequestsFromMapFunc(commoncontroller.MapToOwnerByLabel("", toolchainv1alpha1.ProviderLabelKey))
```
MatousJobanek, line 41: why don't you set the namespace?
```suggestion
	mapToOwnerByLabel := handler.EnqueueRequestsFromMapFunc(commoncontroller.MapToOwnerByLabel(operatorNamespace, toolchainv1alpha1.ProviderLabelKey))
```

MatousJobanek: apart from that, I'm thinking if it wouldn't make sense to do the mapping based on some special key (or value) label used for this controller only. To act on the events of the resources managed by this controller only.

controllers/toolchaincluster_resources/toolchaincluster_resources_controller.go
MatousJobanek: look at what I mentioned in the other comment - regarding the namespace, you didn't set the namespace there, so it would be empty here.

controllers/toolchaincluster_resources/toolchaincluster_resources_controller_test.go
```diff
@@ -0,0 +1,91 @@
...
+func TestToolchainClusterResources(t *testing.T) {
+	// given
+	status := test.NewClusterStatus(toolchainv1alpha1.ToolchainClusterReady, v1.ConditionTrue)
+	toolchainCluster, sec := test.NewToolchainCluster("east", test.MemberOperatorNs, "secret", status, nil)
+	toolchainCluster.Spec.CABundle = "ZHVtbXk="
+	cl := test.NewFakeClient(t, toolchainCluster, sec)
+	service := cluster.NewToolchainClusterServiceWithClient(cl, logf.Log, test.MemberOperatorNs, 3*time.Second, func(config *rest.Config, options client.Options) (client.Client, error) {
+		return client.New(config, options)
+	})
+	defer service.DeleteToolchainCluster("east")
```
MatousJobanek: why do we need all these toolchainCluster stuff? it looks like a copy-paste.
A simple empty fake client should be enough:
```suggestion
	cl := test.NewFakeClient(t)
```

controllers/toolchaincluster_resources/toolchaincluster_resources_controller_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+	t.Run("controller should create service account resource", func(t *testing.T) {
+		// then
+		controller, req := prepareReconcile(toolchainCluster, cl, &serviceAccountFS)
```
MatousJobanek: you forgot to update the tests based on the changes in the controller - it shouldn't act based on the toolchainCluster, but based on some dummy/default SA in the operator namespace

controllers/toolchaincluster_resources/toolchaincluster_resources_controller_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+		// when
+		_, err := controller.Reconcile(context.TODO(), req)
+		require.NoError(t, err)
+		sa := &v1.ServiceAccount{}
+		err = cl.Get(context.TODO(), types.NamespacedName{
+			Namespace: test.MemberOperatorNs,
+			Name:      "toolchaincluster-host",
+		}, sa)
+		require.NoError(t, err)
```
MatousJobanek, line 53: let's verify all other objects from the template

pkg/client/client.go
```diff
@@ -145,6 +145,16 @@ func (c ApplyClient) applyObject(ctx context.Context, obj client.Object, options
 	originalGeneration := existing.GetGeneration()
 	obj.SetResourceVersion(existing.GetResourceVersion())
 
+	// Special handling of ServiceAccounts is required because if a ServiceAccount is reapplied when it already exists, it causes Kubernetes controllers to
+	// automatically create new Secrets for the ServiceAccounts. After enough time the number of Secrets created will hit the Secrets quota and then no new
+	// Secrets can be created. To prevent this from happening, we keep the existing refs to secrets.
+	if strings.EqualFold(obj.GetObjectKind().GroupVersionKind().Kind, "ServiceAccount") {
+		MergeAnnotations(obj, existing.GetAnnotations()) // copy existing annotations
```
MatousJobanek, line 152: shouldn't we merge also labels?
```suggestion
		MergeAnnotations(obj, existing.GetAnnotations()) // copy existing annotations
		MergeLabels(obj, existing.GetLabels())
```

controllers/toolchaincluster_resources/toolchaincluster_resources_controller.go
mfrancisc: I've moved the initialization of the templates in https://github.com/codeready-toolchain/toolchain-common/pull/373/commits/fda181c169cfeb20493405a036d363a5dbac558b 

Thanks!

pkg/client/client.go
```diff
@@ -145,6 +145,16 @@ func (c ApplyClient) applyObject(ctx context.Context, obj client.Object, options
...
+		if err := RetainSecrets(obj, existing); err != nil {
+			return false, err
```
MatousJobanek, line 154: btw, there are also `imagePullSecrets`. 
I wouldn't make it so complicated - was there anything wrong with the original code when we just simply updated the labels & annotations in the existing one and used that object to update the resource?

controllers/toolchaincluster_resources/toolchaincluster_resources_controller.go
MatousJobanek: ok, I just submitted a review for the original code - before you pushed the commit :man_facepalming:

mfrancisc: yeah, sorry about that! I wanted to make sure we agree on this comment before updating the code 😬

MatousJobanek: just go through my comments - you already addressed some of them in your commit

pkg/client/client.go
mfrancisc: I wanted to reuse the existing Get/Create/Update calls from this function, I'm merging the labels and annotations on the existing object in https://github.com/codeready-toolchain/toolchain-common/pull/373/commits/cf4f268fefbf595507d343c3d97b36e593143235

I'm unsure about the ResourceVerision that gets overwritten, not sure if that's an issue.

pkg/client/client.go
mfrancisc: Makes sense, somehow I lost that. Done in https://github.com/codeready-toolchain/toolchain-common/pull/373/commits/cf4f268fefbf595507d343c3d97b36e593143235

Thanks!

controllers/toolchaincluster_resources/toolchaincluster_resources_controller_test.go
mfrancisc: Done in https://github.com/codeready-toolchain/toolchain-common/pull/373/commits/cf4f268fefbf595507d343c3d97b36e593143235

controllers/toolchaincluster_resources/toolchaincluster_resources_controller.go
mfrancisc: I've introduced a custom label value `ResourceControllerLabelValue` in https://github.com/codeready-toolchain/toolchain-common/pull/373/commits/cf4f268fefbf595507d343c3d97b36e593143235. We can move it later to API, or I can do it right away if it looks good.

controllers/label_event_handler.go
```diff
@@ -26,3 +26,23 @@ func MapToOwnerByLabel(namespace, label string) func(object client.Object) []rec
 		return []reconcile.Request{}
 	}
 }
+
+// MapToOwnerByLabelValue returns an event handler will convert events on a resource to requests
+// if the resource matches a given label key and value
+// (if the label exists)
+func MapToOwnerByLabelValue(namespace, labelKey, labelValue string) func(object client.Object) []reconcile.Request {
```
MatousJobanek, line 33: This doesn't map to any owner since the returned `Request` contains the name of the object.
I would maybe rename the func to `MapToControllerNsWhenMatchingLabel` or something like that.

pkg/client/client.go
```diff
@@ -261,3 +271,22 @@ func MergeAnnotations(toolchainObject client.Object, newAnnotations map[string]s
...
 	toolchainObject.SetAnnotations(annotations)
 }
+
+// ApplyUnstructuredObjects applies the given Unstructured objects on the cluster.
+func ApplyUnstructuredObjects(ctx context.Context, cl client.Client, unstructuredObjects []*unstructured.Unstructured, newLabels map[string]string) (bool, error) {
+	applyClient := NewApplyClient(cl)
+	anyApplied := false
+
+	for _, unstructuredObj := range unstructuredObjects {
+		var object = unstructuredObj.DeepCopyObject().(client.Object)
```
MatousJobanek, line 281: why do we need to deep-copy the object? this is already done in the `applyObject` function.
Apart from that `*unstructured.Unstructured` is of the type `client.Object` so you can pass it directly to the `ApplyObject` function.

pkg/client/client.go
```diff
@@ -261,3 +271,22 @@ func MergeAnnotations(toolchainObject client.Object, newAnnotations map[string]s
...
+		log.Info("applying object", "object_namespace", object.GetNamespace(), "object_name", object.GetObjectKind().GroupVersionKind().Kind+"/"+object.GetName())
+		MergeLabels(object, newLabels)
+		_, err := applyClient.ApplyObject(ctx, object)
+		if err != nil {
+			return anyApplied, err
+		}
+		anyApplied = true
```
MatousJobanek, line 288: btw, in this case the `anyApplied` field doesn't make much sense - it will be always true as soon as there is at least one object in the given slice. There is no other option

MatousJobanek (commented):
Looks great, just a few cosmetic suggestions

controllers/label_event_handler.go
mfrancisc: makes sense, renamed in https://github.com/codeready-toolchain/toolchain-common/pull/373/commits/28cde680248bb0135cbec44ad1e98ab9813605e9 . TBH I avoided the `Ns` part, but let me know WDYT. 

Thanks!

pkg/client/client.go
mfrancisc: Good catch! removed in https://github.com/codeready-toolchain/toolchain-common/pull/373/commits/f95ba17b06d5cb6ec1673f3cda2c33b6fb5bf220 . 

Probably a left-over of some old version of the code.

pkg/client/client.go
mfrancisc: if it fails while applying the first object then it returns `false, err`. I mean if it enters the error case at line 285 on the first iteration.

mfrancisc: Maybe I should get the bool from `applyClient.ApplyObject` at line 283 ? 🤔

controllers/label_event_handler.go
MatousJobanek: Looks good :+1:

pkg/client/client.go
MatousJobanek: if it fails with error, then we return an error anyway. 
Let's either drop it or use the boolean returned from `ApplyObject`

mfrancisc: yeah let's drop it for now, and we can introduce it if needed later.

See https://github.com/codeready-toolchain/toolchain-common/pull/373/commits/34595ac1a48392faea9925e454e62c37501512e8 

Thanks for the suggestion.

MatousJobanek (approved):
Cool, thanks for the additional changes :pray: :+1: :rocket:

controllers/toolchaincluster_resources/toolchaincluster_resources_controller.go
```diff
@@ -0,0 +1,77 @@
...
+
+	// add watcher for all kinds from given templates
+	var err error
+	r.templateObjects, err = template.LoadObjectsFromEmbedFS(r.templates, &template.Variables{Namespace: operatorNamespace})
+	if err != nil {
+		return err
+	}
+	mapToOwnerByLabel := handler.EnqueueRequestsFromMapFunc(commoncontroller.MapToControllerByMatchingLabel(operatorNamespace, toolchainv1alpha1.ProviderLabelKey, ResourceControllerLabelValue))
+	for _, obj := range r.templateObjects {
+		build = build.Watches(&source.Kind{Type: obj.DeepCopyObject().(runtimeclient.Object)}, mapToOwnerByLabel, builder.WithPredicates(commonpredicates.LabelsAndGenerationPredicate{}))
```
metlos, line 47: Is there some way to set up the controller to only watch for resources from a single namespace? I think the total number of serviceaccounts+roles+clusterroles+rolebindings in the cluster can get quite big, so we should make our controller as cache-friendly as possible.

controllers/label_event_handler.go
```diff
@@ -26,3 +26,23 @@ func MapToOwnerByLabel(namespace, label string) func(object client.Object) []rec
...
+// MapToControllerByMatchingLabel returns an event handler will convert events on a resource to requests
+// if the resource matches a given label key and value
+// (if the label exists)
+func MapToControllerByMatchingLabel(namespace, labelKey, labelValue string) func(object client.Object) []reconcile.Request {
+	return func(obj client.Object) []reconcile.Request {
+		if labelValueFound, exists := obj.GetLabels()[labelKey]; exists && labelValue == labelValueFound {
+			return []reconcile.Request{
+				{
+					NamespacedName: types.NamespacedName{
+						Namespace: namespace,
```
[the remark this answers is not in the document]
mfrancisc: Good point, the object namespace is fine, no need to pass that as param.

Removed in https://github.com/codeready-toolchain/toolchain-common/pull/373/commits/8a4a23dc9c9656479a50953c8b213b96a296dee6 

Thanks for pointing this out.

I've also noticed that I had to rename test.

controllers/toolchaincluster_resources/toolchaincluster_resources_controller.go
mfrancisc: This is configured in the manager , so for example for host operator it will watch and cache objects in the toolchain-host-operator namespace https://github.com/codeready-toolchain/host-operator/blob/de95481affea496ceb043113a347f64a61abe42c/main.go#L205. 

This is my understanding. So it should not watch and cache objects in all namespaces.

controllers/label_event_handler.go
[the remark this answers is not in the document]
MatousJobanek: Don't forget that the templates can contain also cluster-scoped resources and since the controller will watch all kinds from the template, then you can get events for objects that won't have any namespace set. In other words, we need to use the operator's namespace.

controllers/toolchaincluster_resources/toolchaincluster_resources_controller.go
MatousJobanek: Yes, by default we use a cache scoped to the operator's namespace only, and only when it's really necessary we use a client with a cache scoped for all namespaces - eg: https://github.com/codeready-toolchain/member-operator/blob/cacdbe4681bccddaf7c0d1c9957e97eb8fcfee92/main.go#L188

controllers/toolchaincluster_resources/toolchaincluster_resources_controller.go
```diff
@@ -0,0 +1,77 @@
...
+		For(&v1.ServiceAccount{})
+
+	// add watcher for all kinds from given templates
+	var err error
+	r.templateObjects, err = template.LoadObjectsFromEmbedFS(r.templates, &template.Variables{Namespace: operatorNamespace})
+	if err != nil {
+		return err
+	}
+	mapToOwnerByLabel := handler.EnqueueRequestsFromMapFunc(commoncontroller.MapToControllerByMatchingLabel(toolchainv1alpha1.ProviderLabelKey, ResourceControllerLabelValue))
+	for _, obj := range r.templateObjects {
```
MatousJobanek, line 46: just one idea - what if the template contains more objects of the same kind? wouldn't the builder complain that you are trying to watch the same kind multiple times?

controllers/label_event_handler.go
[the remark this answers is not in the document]
mfrancisc: But triggering a reconcile for a Cluster Scope resource using the operator namespace ? Would this cause a not found error in the reconcile since the object needs to be searched without the namespace set ?

controllers/toolchaincluster_resources/toolchaincluster_resources_controller.go
mfrancisc: Good point!

TBH I'm not sure what will happen in this case , by quickly looking at the builder code I don't think it will error, but maybe worth adding an e2e test case for this ?

controllers/label_event_handler.go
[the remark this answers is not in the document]
MatousJobanek: never mind - I did not realize (after the PTO :smile:) that the template is actually already processed (the namespace variable is already replaced) when the controller is being started. 
All good, sorry for confusion :wink:

controllers/toolchaincluster_resources/toolchaincluster_resources_controller.go
MatousJobanek: I don't think that we need an e2e for this. After all, if we change the template and it would be a problem for the controller/builder, then it would start failing and the current e2e tests would fail. 
It was just thought, if the builder seems to be ok if we call it more times with the same kind, then keep it as it is :wink:

mfrancisc: Right, it should start failing right away with the existing e2e tests, good point!

metlos (approved):

MatousJobanek (approved):

pkg/template/template_test.go
```diff
@@ -17,19 +17,19 @@ import (
 var EFS embed.FS
 
 //go:embed testdata/host/*
-var hostFS embed.FS
+var HostFS embed.FS
```
filariow, line 20: this exported variable seems not used outside this file

pkg/template/template_test.go
```diff
@@ -17,19 +17,19 @@ import (
...
 
 //go:embed testdata/member/*
-var memberFS embed.FS
+var MemberFS embed.FS
```
filariow, line 23: ditto

pkg/client/client.go
```diff
@@ -261,3 +271,18 @@ func MergeAnnotations(toolchainObject client.Object, newAnnotations map[string]s
 	}
 	toolchainObject.SetAnnotations(annotations)
 }
+
+// ApplyUnstructuredObjects applies the given Unstructured objects on the cluster.
+func ApplyUnstructuredObjects(ctx context.Context, cl client.Client, unstructuredObjects []*unstructured.Unstructured, newLabels map[string]string) error {
```
filariow, line 276: I'd suggest to rename this function so that it's more explicit what `newLabels` is used for. Maybe something like `ApplyUnstructuredObjectsWithNewLabels`. WDYT?

controllers/toolchainclusterresources/toolchaincluster_resources_controller_test.go
```diff
@@ -0,0 +1,122 @@
...
+	cl := test.NewFakeClient(t, sa)
+
+	t.Run("controller should create service account resource", func(t *testing.T) {
+		// then
+		controller, req := prepareReconcile(sa, cl, &serviceAccountFS)
+
+		// when
+		_, err := controller.Reconcile(context.TODO(), req)
+		require.NoError(t, err)
+		checkExpectedServiceAccountResources(t, cl)
```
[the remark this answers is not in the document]
mfrancisc: good catch! Thanks!

pkg/template/template_test.go
mfrancisc: good point, see https://github.com/codeready-toolchain/toolchain-common/pull/373/commits/0501aa9c6ce1538834ba6c8b8b159174e01d1527

pkg/template/template_test.go
mfrancisc: see https://github.com/codeready-toolchain/toolchain-common/pull/373/commits/0501aa9c6ce1538834ba6c8b8b159174e01d1527

pkg/client/client.go
mfrancisc: Makes sense. Renamed in https://github.com/codeready-toolchain/toolchain-common/pull/373/commits/59712631cdb878aaaceb7eeb1474dce8fc12e191

controllers/toolchainclusterresources/toolchaincluster_resources_controller_test.go
```diff
@@ -0,0 +1,124 @@
...
+		// when
+		_, err := controller.Reconcile(context.TODO(), req)
+		
+		// then
+		require.NoError(t, err)
+		checkExpectedServiceAccountResources(t, cl)
+	})
+
+	t.Run("controller should create cluster role resource", func(t *testing.T) {
+		// then
```
filariow, line 55: Also here and in the next tests the comments seems misplaced

filariow (commented):
LGTM, just a last nit

controllers/toolchainclusterresources/toolchaincluster_resources_controller_test.go
mfrancisc: I saw now that there was a //given in the parent test, that's why was missing in the subtests, but the //then and //when was inverted indeed. 

Anyways I think it's fine to have //given twice.

Done https://github.com/codeready-toolchain/toolchain-common/pull/373/commits/6f6a0c01bb0bcd7fbcaf173a879d576940bd0cad

filariow (approved):

PR #374, opened 2024-03-14
KSPACE-46: Deprecations and new utility methods for SPCs

metlos:
Deprecate a couple of functions for working with capacity thresholds in ToolchainConfig that is being replaced by the SpaceProvisionerConfig.

Some new utility methods for initializing the SpaceProvisionerConfigs in the tests.


pkg/test/config/toolchainconfig.go
```diff
@@ -85,17 +85,20 @@ func (o AutomaticApprovalOption) Enabled(value bool) AutomaticApprovalOption {
 	return o
 }
 
+// Deprecated: This is no longer used for anything.
```
mfrancisc, line 88: nice - thanks for adding those comments. I guess you'll delete those in follow up PRs, once we switch to the new ones ?

pkg/test/spaceprovisionerconfig/spaceprovisionerconfig.go
```diff
@@ -24,6 +25,38 @@ func NewSpaceProvisionerConfig(name string, namespace string, opts ...CreateOpti
 	return spc
 }
 
+func NewEnabledValidSpaceProvisionerConfigWithTenantRole(name string, namespace string, referencedToolchainCluster string, opts ...CreateOption) *toolchainv1alpha1.SpaceProvisionerConfig {
```
mfrancisc, line 28: minor thing - but TBH I find it easier to read the content of the function, then reading the long name. 

My question is - what is the use case/benefit for having those `New..` constructor functions, couldn't we just avoid adding those permutations and just rely on `NewSpaceProvisionerConfig` with the needed options?

mfrancisc (commented):
Looks Good 👍

I have one comment related to the new utility functions.

pkg/test/config/toolchainconfig.go
metlos: Yes, that's the intention... There's already an issue for that: https://issues.redhat.com/browse/KUBESAW-36

pkg/test/spaceprovisionerconfig/spaceprovisionerconfig.go
metlos: I'm also not a huge fan of long names but at the same time, this function is used on *85* places (and the "overrides" of it on some more), so I thought it would be good to have 1 way that consistently sets up the SPC the same way everywhere rather specifying the 4 options on every one of those 80+ places.

I wanted to be explicit about what the function does instead of calling it "NewInitializedSpc" or something.

MatousJobanek: how about `NewEnabledValidTenantSPC`?

mfrancisc (approved):
Nice Job 👍 

Thanks for addressing my comments!

MatousJobanek (approved):
Looks good :+1:

pkg/test/spaceprovisionerconfig/spaceprovisionerconfig.go
```diff
@@ -24,6 +25,38 @@ func NewSpaceProvisionerConfig(name string, namespace string, opts ...CreateOpti
...
+}
+
+func NewEnabledValidSPC(name string, namespace string, referencedToolchainCluster string, opts ...CreateOption) *toolchainv1alpha1.SpaceProvisionerConfig {
+	return NewSpaceProvisionerConfig(name, namespace,
+		append(opts,
+			Enabled(true),
+			WithReadyConditionValid(),
+			ReferencingToolchainCluster(referencedToolchainCluster))...,
+	)
+}
```
MatousJobanek, line 54: just going through the other PRs - if I'm not mistaken, all the SPCs are created only in host-operator namespace, so you could simplify it, drop the "namespace" parameter, and set the `HostOperatorNs` automatically

fbm3307: IMO keeping it as a parameter would be a better option . It would be hidden, but clean i guess

fbm3307 (approved):
Looks Good

pkg/test/spaceprovisionerconfig/spaceprovisionerconfig.go
metlos: That's a good observation. I thought about it and I think it's best to just move these utility functions into `host-operator` and only leave the most generic function here, because it is also used in e2e tests.

The latest commit here and in https://github.com/codeready-toolchain/host-operator/pull/991 should have these functions moved.

MatousJobanek: good idea, thanks for moving it & simplifying it :+1:

PR #376, opened 2024-03-20
Update k8s dependencies to 4.14 and controller-runtime to v0.15

ranakan19:
Updates k8s dependencies to 0.27 and controller-runtime to v0.15.
With the update to controller-runtime v0.15 (changes in release detailed [here](https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.15.0)), this PR has following changes to combat the breaking changes:
- add context to event handlers - https://github.com/kubernetes-sigs/controller-runtime/pull/2139
- Add statusSubResource for the changes in fake client status handling - https://github.com/kubernetes-sigs/controller-runtime/pull/2259 - affects Update and Patch
- Add all toolchain resources as StatusSubresource in FakeClient

Changes wrt - https://issues.redhat.com/browse/SANDBOX-558 
DO NOT MERGE (All the PRs related to version updates need to be merged at once, but these PR are for early feedback so that there are not a lot of changes to review at once)


xcoulon (approved):

pkg/test/client.go
```diff
@@ -16,13 +16,14 @@ import (
 )
 
 // NewFakeClient creates a fake K8s client with ability to override specific Get/List/Create/Update/StatusUpdate/Delete functions
-func NewFakeClient(t T, initObjs ...runtime.Object) *FakeClient {
+func NewFakeClient(t T, initObjs ...client.Object) *FakeClient {
```
ranakan19, line 19: changes initObjs type from runtime.Object to client.Object - see discussion [here](https://redhat-internal.slack.com/archives/CHK0J6HT6/p1709236452856659)
WithStatusSubresource can only accept client.Object was a major reason to consider making this change.

controllers/label_event_handler_test.go
```diff
@@ -15,6 +16,7 @@ func TestLabelMapper(t *testing.T) {
 
 	t.Run("resource with expected label", func(t *testing.T) {
 		// given
+		ctx := context.TODO()
```
mfrancisc, line 19: minor - we could move the context creation up in the parent test

pkg/test/client_test.go
```diff
@@ -109,11 +109,24 @@ func TestNewClient(t *testing.T) {
...
-			assert.EqualValues(t, 1, retrieved.Generation) // Generation not changed
+			assert.EqualValues(t, 1, retrieved.Generation) // Generation not changed, since blank update
+		})
+
+		t.Run("status update with actual update in status", func(t *testing.T) {
+			created, retrieved := createAndGetDeployment(t, fclient)
+			created.Status.Replicas = 2
+			require.NoError(t, fclient.Status().Update(context.TODO(), created))
+			require.NoError(t, fclient.Get(context.TODO(), types.NamespacedName{Namespace: "somenamespace", Name: created.Name}, retrieved))
+			assert.EqualValues(t, 2, retrieved.Status.Replicas) // replicas count changed to 2
```
mfrancisc, line 124: should we check that generation is 1 or it doesn't bring any benefit ?

mfrancisc (approved):
Nice Job! 

I have only few minor comments.

pkg/test/client.go
```diff
@@ -16,13 +16,14 @@ import (
...
 	s := scheme.Scheme
 	err := toolchainv1alpha1.AddToScheme(s)
 	require.NoError(t, err)
 	cl := fake.NewClientBuilder().
 		WithScheme(s).
-		WithRuntimeObjects(initObjs...).
+		WithObjects(initObjs...).
+		WithStatusSubresource(initObjs...).
```
MatousJobanek, line 26: Related to our discussion in the host-operator PR, we could try to improve this logic to register status sub-resource for all "toolchain" resources. In this way we would know that, the status sub-resource is registered for all objects we manage. 
So, could you go through all toolchain GVKs and add them to the client builder as objects with the status sub-resource? Try to use a generic logic using the scheme and GVKs, we don't want to update the logic as soon as we add or remove  a CRD.

alexeykazakov (commented):
Looks good overall. Just one minor comment. And +1 to @MatousJobanek's suggestion regarding registering status sub resources for all our API.

pkg/test/client.go
ranakan19: added in [6dfea01](https://github.com/codeready-toolchain/toolchain-common/pull/376/commits/6dfea0122719082c1bfd61eafd6361989534c551)

pkg/test/client_test.go
```diff
@@ -109,11 +109,24 @@ func TestNewClient(t *testing.T) {
...
+		})
+
+		t.Run("status update fails when the objects being updated doesn't have status", func(t *testing.T) {
+			created, _ := createAndGetSecret(t, fclient)
+			require.Error(t, fclient.Status().Update(context.TODO(), created))
```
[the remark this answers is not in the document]
ranakan19: done in [e3932d3](https://github.com/codeready-toolchain/toolchain-common/pull/376/commits/e3932d3357197b02463aa766ff0344a29008dc35)

alexeykazakov (approved):
Nice!

pkg/configuration/cache_test.go
```diff
@@ -343,11 +343,11 @@ func TestMultipleExecutionsInParallel(t *testing.T) {
...
 
 			// then
-			require.NoError(t, err)
+			assert.NoError(t, err)
 			toolchaincfg, ok := config.(*toolchainv1alpha1.ToolchainConfig)
-			require.True(t, ok)
+			assert.True(t, ok)
 			assert.NotEmpty(t, toolchaincfg.Spec)
-			require.NotEmpty(t, secrets)
+			assert.NotEmpty(t, secrets)
```
MatousJobanek, line 350: is there any reason for changing it to `assert`? IIRC, the best practice is to use `require` for some specific assertions like the error one.

pkg/test/client.go
```diff
@@ -182,6 +192,22 @@ func toMap(obj runtime.Object) (map[string]interface{}, error) {
 	return m, nil
 }
 
+func getAllToolchainResources(s *runtime.Scheme) []client.Object {
+	toolchainObjs := make([]client.Object, 0)
+	KindToTypeMap := s.KnownTypes(toolchainv1alpha1.GroupVersion)
+	var kinds []string
+	for key := range KindToTypeMap {
```
MatousJobanek, line 199: any reason for starting with the capital letter?
```suggestion
	kindToTypeMap := s.KnownTypes(toolchainv1alpha1.GroupVersion)
	var kinds []string
	for key := range kindToTypeMap {
```

pkg/test/client.go
```diff
@@ -182,6 +192,22 @@ func toMap(obj runtime.Object) (map[string]interface{}, error) {
...
+		kinds = append(kinds, key)
+	}
+	for _, k := range kinds {
+		obj := &unstructured.Unstructured{}
+		gvk := schema.GroupVersionKind{Group: toolchainv1alpha1.GroupVersion.Group, Version: toolchainv1alpha1.GroupVersion.Version, Kind: k}
+		obj.SetGroupVersionKind(gvk)
+		toolchainObjs = append(toolchainObjs, obj)
+	}
+	return toolchainObjs
+}
```
MatousJobanek, line 209: you can simplify it to:
```suggestion
func getAllToolchainResources(s *runtime.Scheme) []client.Object {
	kindToTypeMap := s.KnownTypes(toolchainv1alpha1.GroupVersion)
	toolchainObjs := make([]client.Object, 0, len(kindToTypeMap))
	for kind := range kindToTypeMap {
		obj := &unstructured.Unstructured{}
		obj.SetGroupVersionKind(toolchainv1alpha1.GroupVersion.WithKind(kind))
		toolchainObjs = append(toolchainObjs, obj)
	}
	return toolchainObjs
}
```

pkg/configuration/cache_test.go
ranakan19: you're right,
linter check in host suggested to use require instead of assert.
but I was changing it to assert here because of linter complaining previously - in this commit https://github.com/codeready-toolchain/toolchain-common/pull/376/commits/0986552a344431731de2d9c8cf0c3484a60c5553

Let me check what's going on. I expect the linter to complain uniformly across all repos.  but in general i do agree require is the convention to usr instead of assert with error especially.

ranakan19: its because of `"go-require: require must only be used in the goroutine running the test function (testifylint)"` added a comment beside it for explanation

MatousJobanek: ah, ok. This is a bit different since we run the verification inside our goroutine :+1: 
Thanks for taking a look at it

pkg/test/client_test.go
```diff
@@ -139,6 +155,11 @@ func TestNewClient(t *testing.T) {
 			require.NoError(t, fclient.Status().Patch(context.TODO(), retrieved, depPatch))
 		})
 
+		t.Run("status create fails", func(t *testing.T) {
+			_, retrieved := createAndGetDeployment(t, fclient)
+			require.EqualError(t, fclient.Status().Create(context.TODO(), retrieved, retrieved), "fakeSubResourceWriter does not support create for status")
```
ranakan19, line 160: Error returned from fakeclient - https://github.com/kubernetes-sigs/controller-runtime/blob/release-0.15/pkg/client/fake/client.go#L1116 
Only substatus resource supported is eviction

PR #377, opened 2024-03-22
add PublicViewerConfig utils

filariow:


pkg/configuration/publicviewer_config.go
```diff
@@ -0,0 +1,25 @@
...
+	return c.Config.Username
+}
+
+func (c PublicViewerConfig) IsPublicViewer(username string) bool {
+	if !c.Enabled() {
+		return false
+	}
+
+	return c.Config.Username == username
+}
```
mfrancisc, line 25: should we check that c.Config.Username is not empty ? I mean could it happen that we enable the config without setting a username? WDYT?

filariow: the RegEx validation should prevent this from happening

```
// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`
```

filariow: no, you're right, this can happen. Let me look into it better, great catch!

filariow: Do you think [this approach](https://github.com/codeready-toolchain/host-operator/pull/998#discussion_r1537753050) could help also here?

mfrancisc: As pointed out in the comment https://github.com/codeready-toolchain/host-operator/pull/998#discussion_r1538759675 , I would remove the field.

PR #381, opened 2024-04-03
Make NSTemplateTier generator generic

MatousJobanek:
This PR moves the logic of nstemplatetier_generator.go from host-operator to toolchain-common and make it generic so it can be used by a ksctl command for generating NSTemplateTier & TierTemplate manifests locally in a GitOps repository.
There is almost no change in the logic of the generator, it should work in the same way as it was before, just with the difference that it doesn't load templates from the assets, but takes them from the map passed as a parameter.

[KUBESAW-59](https://issues.redhat.com/browse/KUBESAW-59)

related PR in host-operator repo https://github.com/codeready-toolchain/host-operator/pull/1006

### DIFF
--- a/pkg/template/nstemplatetiers/nstemplatetier_generator.go
+++ b/pkg/template/nstemplatetiers/nstemplatetier_generator.go
@@ -21,7 +21,7 @@ import (
 
 var log = logf.Log.WithName("templates")
 
-type EnsureObject func(toEnsure runtimeclient.Object, canUpdate bool) (bool, error)
+type EnsureObject func(toEnsure runtimeclient.Object, canUpdate bool, tierName string) (bool, error)
 
 type TierGenerator struct {
 	ensureObject    EnsureObject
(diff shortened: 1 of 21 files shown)


alexeykazakov (approved):

pkg/template/nstemplatetiers/nstemplatetier_generator.go
```diff
@@ -0,0 +1,466 @@
...
+}
+
+// createTierTemplates creates all TierTemplate resources from the tier map
+func (t *TierGenerator) createTierTemplates() error {
+	// create the templates
+	for _, tierTmpls := range t.templatesByTier {
+		for _, tierTmpl := range tierTmpls.tierTemplates {
+			log.Info("creating TierTemplate", "namespace", tierTmpl.Namespace, "name", tierTmpl.Name)
+			// using the "standard" client since we don't need to support updates on such resources, they should be immutable
+			if _, err := t.ensureObject(tierTmpl, false); err != nil {
```
mfrancisc, line 282: the second parameter wasn't immediately clear to me, but I like the new `ensureObject` function.

mfrancisc (approved):
Nice Job 👍 

I like that we switched to `embed.FS` and the new `ensureObject` function.

fbm3307 (approved):

PR #384, opened 2024-04-04
fix some linter's complaints

filariow:
Signed-off-by: Francesco Ilario <filario@redhat.com>

### DIFF
--- a/pkg/cluster/service.go
+++ b/pkg/cluster/service.go
@@ -234,7 +234,7 @@ func ListToolchainClusterConfigs(cl client.Client, namespace string, timeout tim
 	}
 	var configs []*Config
 	for _, cluster := range toolchainClusters.Items {
-		cluster := cluster
+		cluster := cluster // TODO We won't need it after upgrading to go 1.22: https://go.dev/blog/loopvar-preview
 		clusterConfig, err := NewClusterConfig(cl, &cluster, timeout)
 		if err != nil {
 			return nil, err
--- a/pkg/cluster/service.go
+++ b/pkg/cluster/service.go
@@ -234,7 +234,8 @@ func ListToolchainClusterConfigs(cl client.Client, namespace string, timeout tim
 	}
 	var configs []*Config
 	for _, cluster := range toolchainClusters.Items {
-		clusterConfig, err := NewClusterConfig(cl, &cluster, timeout) // nolint:gosec
+		cluster := cluster
+		clusterConfig, err := NewClusterConfig(cl, &cluster, timeout)
 		if err != nil {
 			return nil, err
 		}
--- a/pkg/template/processor.go
+++ b/pkg/template/processor.go
@@ -2,7 +2,7 @@ package template
 
 import (
 	"fmt"
-	"math/rand" //nolint:gosec
+	"math/rand"
 	"time"
 
 	templatev1 "github.com/openshift/api/template/v1"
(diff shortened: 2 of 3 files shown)


pkg/cluster/service.go
```diff
@@ -234,7 +234,8 @@ func ListToolchainClusterConfigs(cl client.Client, namespace string, timeout tim
 	}
 	var configs []*Config
 	for _, cluster := range toolchainClusters.Items {
-		clusterConfig, err := NewClusterConfig(cl, &cluster, timeout) // nolint:gosec
+		cluster := cluster
```
mfrancisc, line 237: I see that this solves the gosec issue, but looks a bit confusing to me 😕

mfrancisc (commented):
Thanks a lot for looking at those linting issues and making those improvements 🙏 

I have one small comment.

pkg/cluster/service.go
filariow: That's ugly, I agree. Btw, this is needed because of how go handles for loop variables until go 1.22. It basically reuses the same memory behind the variable, so using the variable as pointer is not safe. 

Alternatively, it can be fixed by using the array and the index provided by the loop.

I find this approach better because 
1. code will remain cleaner: no `&clusters[i]` around when you could have had/used the `cluster` variable from the for loop
2. The `cluster := cluster` will be around only until the project uses go < 1.22 
3. It will be easy after upgrading to go 1.22 to find these lines and delete them

alexeykazakov: I agree that it's safer to use something like `varName := varName` rather than `copyOfVarName := varName` but it might look confusing for people who are not aware of that golang problem. So I would just add a comment to the code explaining why we need it.

mfrancisc: @filariow thanks for your explanation, tbh I do prefer using the `array[i]` syntax, but it's just personal preference. 
I'm fine with adding a comment and/or TODO so that we can explain why we need it and a reminder to remove those lines when we upgrade go version > 1.22

rajivnathan (approved):

pkg/cluster/service.go
rajivnathan: If we use `array[i]` then we wouldn't need to update the code later for 1.22 right? If so I think that'd be preferable but I think as long as we have a common comment that can be searched used to find these lines it'll make the upgrade to 1.22 easier.

filariow: @rajivnathan As per my understanding, whatever choice we make here we are not going to require code changes when upgrading to go 1.22. At the same time, if after the upgrade to go 1.22 the team would like to have the following notation across the codebase a refactor would be needed either code style we chose now: 

```golang
for _, cluster := range clusters {
   doSomethingOnCluster(&cluster)
   doSomeOtherThingOnCluster(&cluster)
}
```

In the following a recap of the options we have:

**Accessing the array by index**
```golang
for i := range clusters {
   doSomethingOnCluster(&clusters[i])
   doSomeOtherThingOnCluster(&clusters[i])
}

// Or 

for i := range clusters {
   cluster := cluster[i]

   doSomethingOnCluster(&cluster)
   doSomeOtherThingOnCluster(&cluster)
}
```

**using the `range` variable**
```golang
for _, cluster := range clusters {
   cluster := cluster

   doSomethingOnCluster(&cluster)
   doSomeOtherThingOnCluster(&cluster)
}
```

I'd propose to use the notation the team finds more readable now, and decide whether to refactor the for loops again when bumping to go 1.22. WDYT?

alexeykazakov: I personally like using `range` :) But it's just personal preferences. But anyway it's all cosmetic changes. I would keep your current PR as is (keep using the current range but introduce `varName := varName`). Just add a comment about why we need it. When we migrate to 1.22 we can clean it up but as you rightly said we won't have to.

filariow: done in 207c3e39fb7815199ff8224b55d0ad98b26abf04

mfrancisc (approved):

alexeykazakov (approved):

PR #385, opened 2024-04-05
clear last applied configuration

mfrancisc:
see https://redhat-internal.slack.com/archives/C06MJ2DVBU4/p1712310833089889

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -113,10 +113,6 @@ func (c ApplyClient) applyObject(ctx context.Context, obj client.Object, options
 	if config.saveConfiguration {
 		// set current object as annotation
 		annotations := obj.GetAnnotations()
-		// reset the previous config to avoid recursive embedding of the object
-		if _, found := obj.GetAnnotations()[LastAppliedConfigurationAnnotationKey]; found {
-			delete(obj.GetAnnotations(), LastAppliedConfigurationAnnotationKey)
-		}
 		newConfiguration = GetNewConfiguration(obj)
 		if annotations == nil {
 			annotations = map[string]string{}
@@ -209,7 +205,12 @@ func clusterIP(obj runtime.Object) (string, bool, error) {
 	}
 }
 
-func GetNewConfiguration(newResource runtime.Object) string {
+func GetNewConfiguration(newResource client.Object) string {
+	// reset the previous config to avoid recursive embedding of the object
+	copyResource := newResource.DeepCopyObject().(client.Object)
+	if _, found := copyResource.GetAnnotations()[LastAppliedConfigurationAnnotationKey]; found {
+		delete(copyResource.GetAnnotations(), LastAppliedConfigurationAnnotationKey)
+	}
 	newJSON, err := marshalObjectContent(newResource)
 	if err != nil {
 		log.Error(err, "unable to marshal the object", "object", newResource)
@@ -282,7 +283,7 @@ func ApplyUnstructuredObjectsWithNewLabels(ctx context.Context, cl client.Client
 	for _, unstructuredObj := range unstructuredObjects {
 		log.Info("applying object", "object_namespace", unstructuredObj.GetNamespace(), "object_name", unstructuredObj.GetObjectKind().GroupVersionKind().Kind+"/"+unstructuredObj.GetName())
 		MergeLabels(unstructuredObj, newLabels)
-		_, err := applyClient.ApplyObject(ctx, unstructuredObj)
+		_, err := applyClient.ApplyObject(ctx, unstructuredObj, SaveConfiguration(false))
 		if err != nil {
 			return err
 		}
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -124,6 +124,7 @@ func (c ApplyClient) applyObject(ctx context.Context, obj client.Object, options
 	namespacedName := types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetName()}
 	if err := c.Client.Get(ctx, namespacedName, existing); err != nil {
 		if apierrors.IsNotFound(err) {
+			obj.SetResourceVersion("") // reset resource version when creating to avoid error: resourceVersion should not be set on objects to be created
 			return true, c.createObj(ctx, obj, config.owner)
 		}
 		return false, errors.Wrapf(err, "unable to get the resource '%v'", existing)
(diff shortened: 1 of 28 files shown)


pkg/client/client.go
```diff
@@ -113,7 +113,11 @@ func (c ApplyClient) applyObject(ctx context.Context, obj client.Object, options
 	if config.saveConfiguration {
 		// set current object as annotation
 		annotations := obj.GetAnnotations()
-		newConfiguration = getNewConfiguration(obj)
+		// reset the previous config to avoid recursive embedding of the object
+		if _, found := obj.GetAnnotations()[LastAppliedConfigurationAnnotationKey]; found {
+			delete(obj.GetAnnotations(), LastAppliedConfigurationAnnotationKey)
```
alexeykazakov, line 118: I would rather move this to the `getNewConfiguration()` function. So the function always returns a clean object to be used in the last applied configuration annotation. And also would create a deep copy of the object in that `getNewConfiguration()` function before modifying it (removing the annotation). It won't break anything in the current code if we don't do it but it would make the function less error prone in case we modify the code in the future.

mfrancisc: Good point, done in https://github.com/codeready-toolchain/toolchain-common/pull/385/commits/aacc2c6479568c7e2dfba7bef4749098ea3336f6 

I've also added the annotation to the existing obj in unit tests.

pkg/client/client.go
```diff
@@ -278,7 +287,7 @@ func ApplyUnstructuredObjectsWithNewLabels(ctx context.Context, cl client.Client
 	for _, unstructuredObj := range unstructuredObjects {
 		log.Info("applying object", "object_namespace", unstructuredObj.GetNamespace(), "object_name", unstructuredObj.GetObjectKind().GroupVersionKind().Kind+"/"+unstructuredObj.GetName())
 		MergeLabels(unstructuredObj, newLabels)
-		_, err := applyClient.ApplyObject(ctx, unstructuredObj)
+		_, err := applyClient.ApplyObject(ctx, unstructuredObj, SaveConfiguration(false))
```
mfrancisc, line 290: Unfortunately for Unstructured objects it is still not working, so I'm keeping the saveConfiguration functionality disabled. I've tried multiple things:

- removing the `LastAppliedConfigurationAnnotationKey` annotation from both new object before creating it
- removing the`LastAppliedConfigurationAnnotationKey` annotation from the existing object before merging the annotations with the new object
- removing the special handling of `Unstructured` objects in `marshalObjectContent`:
````
if newRes, ok := newResource.(runtime.Unstructured); ok {
		return json.Marshal(newRes.UnstructuredContent())
}
````

but nothing seems to work. The `toolchain.dev.openshift.com/last-applied-configuration` annotation is still being  recursively embedded.

Comment (mfrancisc) [reply pkg/client/client.go:290]: I'm also wondering if we really need this functionality at all. We could

a. trigger the Update of the object anyway , that should do anything if there are no changes on the object. 
b. rely on `kubectl.kubernetes.io/last-applied-configuration` , how is this different from our custom annotation ?

Comment (alexeykazakov) [reply pkg/client/client.go:290]: Good points @mfrancisc. Not sure why we started using our annotation instead. Maybe @MatousJobanek remembers why.

File: pkg/client/client.go [hunk 3]
```diff
@@ -124,6 +124,7 @@ func (c ApplyClient) applyObject(ctx context.Context, obj client.Object, options
 	namespacedName := types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetName()}
 	if err := c.Client.Get(ctx, namespacedName, existing); err != nil {
 		if apierrors.IsNotFound(err) {
+			obj.SetResourceVersion("") // reset resource version when creating to avoid error: resourceVersion should not be set on objects to be created
```
MatousJobanek, line 127: Do you know in which cases the provided object contained the `resourceVersion`?

pkg/client/client_test.go
```diff
@@ -339,6 +339,106 @@ func TestApplySingle(t *testing.T) {
 			assert.Equal(t, "second-value", configMap.Data["first-param"])
 		})
 	})
+
+	t.Run("updates of ServiceAccount", func(t *testing.T) {
+
+		t.Run("should update service account with last applied configuration", func(t *testing.T) {
```
MatousJobanek, line 345: since the "last-appplied-configuration" is used and is set to the same value, then there the SA shouldn't be updated.

pkg/client/client_test.go
```diff
@@ -339,6 +339,106 @@ func TestApplySingle(t *testing.T) {
...
+					Namespace: existingSA.Namespace,
+				},
+			}
+			existingSA.Secrets = secretRefs
+			existingLastAppliedAnnotation := map[string]string{
+				client.LastAppliedConfigurationAnnotationKey: client.GetNewConfiguration(existingSA),
+			}
+			existingSA.SetAnnotations(existingLastAppliedAnnotation) // let's set the last applied annotation
+			cl, cli := newClient(t)
+			_, err := cl.ApplyRuntimeObject(context.TODO(), existingSA)
```
MatousJobanek, line 361: I noticed that we should pass a deep-copy of the object, otherwise, the generation as well as resourceVersion values would be set and would influence the late execution when the object that is passed to the function again: 
```suggestion
			_, err := cl.ApplyRuntimeObject(context.TODO(), existingSA.DeepCopyObject())
```

MatousJobanek (approved):
Looks good, just some comments related to the tests

pkg/client/client.go
MatousJobanek: it was an attempt to get closer to what kubectl does, to optimize the number of updates we do from our controllers and to have some control over what was updated and what not. 
But I really doubt that it works as we hoped that it could work, so I'm for dropping the annotation :+1: to simplify the logic. But let's do it in a separate PR.

pkg/client/client.go
mfrancisc: It was happening with the new toolchaincluster_resource_controller. Unfortunately I don't have the logs anymore but creation of the objects was failing due to this ResourceVersion field being set. TBH I was confused about why this was happening, since the objects should be "new" from the templates directly.

pkg/client/client_test.go
mfrancisc: done in https://github.com/codeready-toolchain/toolchain-common/pull/385/commits/91ec6712cad89744600f375013fff20bcbadbe55

pkg/client/client.go
MatousJobanek: didn't you see the failure as part of the unit tests? see the comment below - it could be the case.

mfrancisc: I wasn't able to reproduce the failure in unit tests. you mean that we should pass a deepcopy of the objects here https://github.com/codeready-toolchain/toolchain-common/pull/385/files#diff-2895f30116602d8fe6b0545c186d4f81247fa4eaab049dab174c6c91c0036e08R291 ? Or something else ..

PR #395, opened 2024-05-02
use Google UUID

ranakan19:
instead of upgrading gofrs to the new version, choosing to go ahead with google uuid instead. No functional reason of choosing one over the other, just that google library has a wider base of users (although gofrs base wasn't small). This way only one library is used for uuid and can be in sync with ksctl as well.

### DIFF
--- a/pkg/test/env_test.go
+++ b/pkg/test/env_test.go
@@ -138,10 +138,8 @@ func TestUnsetEnvVarAndRestore(t *testing.T) {
 }
 
 func unsetVariable(t *testing.T) string {
-	u, err := uuid.NewRandom()
-	require.NoError(t, err)
-	varName := u.String()
-	err = os.Unsetenv(varName)
+	varName := uuid.NewString()
+	err := os.Unsetenv(varName)
 	require.NoError(t, err)
 	_, present := os.LookupEnv(varName)
 	require.False(t, present)
@@ -149,15 +147,10 @@ func unsetVariable(t *testing.T) string {
 }
 
 func setVariable(t *testing.T) (string, string) {
-	u, err := uuid.NewRandom()
-	require.NoError(t, err)
-	key := u.String()
+	key := uuid.NewString()
+	value := uuid.NewString()
 
-	u, err = uuid.NewRandom()
-	require.NoError(t, err)
-	value := u.String()
-
-	err = os.Setenv(key, value)
+	err := os.Setenv(key, value)
 	require.NoError(t, err)
 	_, present := os.LookupEnv(key)
 	require.True(t, present)


pkg/test/env_test.go
```diff
@@ -149,11 +149,11 @@ func unsetVariable(t *testing.T) string {
 }
 
 func setVariable(t *testing.T) (string, string) {
-	u, err := uuid.NewV4()
+	u, err := uuid.NewRandom()
 	require.NoError(t, err)
 	key := u.String()
 
-	u, err = uuid.NewV4()
+	u, err = uuid.NewRandom()
```
rajivnathan, line 156: Same here

pkg/test/env_test.go
```diff
@@ -138,7 +138,7 @@ func TestUnsetEnvVarAndRestore(t *testing.T) {
 }
 
 func unsetVariable(t *testing.T) string {
-	u, err := uuid.NewV4()
+	u, err := uuid.NewRandom()
```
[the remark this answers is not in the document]
ranakan19: I used `NewRandom()` so as to not disturb the test, as it returns `(uuid,err)`, while `uuid.New()` return `(uuid)`
`uuid.New()` is equivalent to `uuid.Must(uuid.NewRandom()) `
hope it helps?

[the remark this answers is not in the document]
MatousJobanek: but is it a problem to drop the error and require the generation of the UUID pass? I mean, do you expect that it would fail?
Personally, I would vote for `uuid.New()` to make the test a bit simpler and not deal with the error.

pkg/test/env_test.go
```diff
@@ -138,7 +138,7 @@ func TestUnsetEnvVarAndRestore(t *testing.T) {
...
 	require.NoError(t, err)
 	varName := u.String()
```
MatousJobanek, line 143: this whole thing could be simplified to:
```suggestion
	varName := uuid.NewString()
```
right?

pkg/test/env_test.go
```diff
@@ -149,11 +149,11 @@ func unsetVariable(t *testing.T) string {
 }
 
 func setVariable(t *testing.T) (string, string) {
-	u, err := uuid.NewV4()
+	u, err := uuid.NewRandom()
 	require.NoError(t, err)
 	key := u.String()
```
MatousJobanek, line 154: the same
```suggestion
	key := uuid.NewString()
```

pkg/test/env_test.go
```diff
@@ -149,11 +149,11 @@ func unsetVariable(t *testing.T) string {
...
 
-	u, err = uuid.NewV4()
+	u, err = uuid.NewRandom()
 	require.NoError(t, err)
 	value := u.String()
```
MatousJobanek, line 158:
```suggestion
	value := uuid.NewString()
```

pkg/test/env_test.go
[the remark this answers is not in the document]
ranakan19: No it's not a problem, I can change it now.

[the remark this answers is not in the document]
ranakan19: done in 96f0276

pkg/test/env_test.go
ranakan19: done in 96f0276

pkg/test/env_test.go
ranakan19: yes done in 96f0276

pkg/test/env_test.go
ranakan19: yes done in 96f0276

pkg/test/env_test.go
ranakan19: yep, done in 96f0276

MatousJobanek (approved):
Thanks

metlos:
What about using `k8s.io/apimachinery/pkg/util/uuid` so that we don't have to pick the uuid library at all?

MatousJobanek:
Looking at https://github.com/kubernetes/apimachinery/blob/master/pkg/util/uuid/uuid.go they use `github.com/google/uuid` as well, so it technically  doesn't matter if we use `github.com/google/uuid` directly or via apimachinery

rajivnathan (approved):

PR #397, opened 2024-05-03
add ConditionsMatch function

sbryzak:
Refactor to move conditions testing functions from the test package to the conditions package so that they may be used outside of tests


alexeykazakov (commented):
Looks good. However we need unit tests :)

sbryzak:
@metlos thanks for the review - this PR is not actually new code, it's just a refactoring to move existing code to a different package in order to allow it to be used outside of the tests.  It is already used extensively throughout the dev sandbox codebase and very well tested, so i'm hesitant to change any of it at this point.

MatousJobanek:
@sbryzak you are right - this code was used in unit tests, but could you please provide more context for why you need to use it in production? As @metlos mentioned in his comment, there is no point in checking the message content in production.

pkg/condition/condition.go
```diff
@@ -146,3 +146,32 @@ func HasConditionReason(conditions []toolchainv1alpha1.Condition, conditionType
...
+func ConditionsMatch(first, second []toolchainv1alpha1.Condition) bool {
+	if len(first) != len(second) {
+		return false
+	}
+	for _, c := range first {
+		if !ContainsCondition(second, c) {
+			return false
+		}
+	}
+	for _, c := range second {
```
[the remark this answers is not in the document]
alexeykazakov: I agree. And while, as you said, you are not introducing a new code but rather moving it to a different package it's still OK to make such optimization. We just need to make sure that the code is properly covered by the tests.

alexeykazakov:
Regarding the `Message` thing. I agree that matching `Message` in runtime code is at least questionable. However it may make sense in tests. Depending on how it's actually used in tests. So I would be safer to keep it for the tests.
We could do something like this:
```
func containsCondition(conditions []toolchainv1alpha1.Condition, contains toolchainv1alpha1.Condition, ignoreMessage bool) bool {
    ...
}

...

func ContainsCondition(conditions []toolchainv1alpha1.Condition, contains toolchainv1alpha1.Condition) bool {
    return containsConditions(conditions, contains, true)
}

func ContainsConditionWithMessage(conditions []toolchainv1alpha1.Condition, contains toolchainv1alpha1.Condition) bool {
    return containsConditions(conditions, contains, false)
}
```

And in the test package:
```
func ContainsCondition(conditions []toolchainv1alpha1.Condition, contains toolchainv1alpha1.Condition) bool {
    return ContainsConditionWithMessage(conditions, contains)
}
```

MatousJobanek:
Sure, it makes completely sense to match the message in the tests, but why moving it out of the pkg/test/ package? All these functions that are there can be (and are) already used by unit tests in host-operator see https://github.com/codeready-toolchain/host-operator/blob/423ebbae4392a3752562f9f975e8db3532b31b2d/controllers/usersignup/usersignup_controller_test.go#L33 
Let's keep it in the pkg/test/ package to make it clear that it's supposed to be used in tests only. There is no need to match the message in the production code.

alexeykazakov:
> Let's keep it in the pkg/test/ package to make it clear that it's supposed to be used in tests only. There is no need to match the message in the production code.

Shane wants to use it in runtime too. See the new predicate code in the host operator PR: https://github.com/codeready-toolchain/host-operator/pull/988/files#diff-686778ddb77570eda2cfd3b717a28573e5b4399c35d7f85936143fe0feff0dedR52

pkg/condition/condition.go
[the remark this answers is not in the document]
sbryzak: I've removed the second loop from (now) both places

pkg/condition/condition.go
```diff
@@ -146,3 +146,32 @@ func HasConditionReason(conditions []toolchainv1alpha1.Condition, conditionType
...
+			return false
+		}
+	}
+	return true
+}
+
+// ContainsCondition returns true if the specified list of conditions contains the specified condition and the statuses of the conditions match.
+// LastTransitionTime is ignored.
+func ContainsCondition(conditions []toolchainv1alpha1.Condition, contains toolchainv1alpha1.Condition) bool {
+	for _, c := range conditions {
```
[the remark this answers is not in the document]
sbryzak: This has now been refactored as per Alexey's suggestion, and now allows a flag to be passed in specifying whether the message property should be considered in the comparison.

MatousJobanek:
ok, this is exactly what I was looking for, thanks for the link.
TBH, I don't see any reason for using any predicate in this particular case at all. I mean, what is the point of the predicate? But let's move this discussion to the other PR.

pkg/condition/condition.go
```diff
@@ -146,3 +146,39 @@ func HasConditionReason(conditions []toolchainv1alpha1.Condition, conditionType
...
+			if ignoreMessage {
+				return contains.Status == c.Status && contains.Reason == c.Reason
+			}
+			return contains.Status == c.Status && contains.Reason == c.Reason && contains.Message == c.Message
+		}
+	}
+	return false
+}
+
+// ContainsCondition returns true if the specified list of conditions contains the specified condition and the statuses of the conditions match.
```
alexeykazakov, line 176: Let's add a comment stating that it ignores condition's messages.

pkg/condition/condition.go
```diff
@@ -146,3 +146,39 @@ func HasConditionReason(conditions []toolchainv1alpha1.Condition, conditionType
...
+// LastTransitionTime is ignored.
+func ContainsCondition(conditions []toolchainv1alpha1.Condition, contains toolchainv1alpha1.Condition) bool {
+	return containsConditions(conditions, contains, true)
+}
+
+func ContainsConditionWithMessage(conditions []toolchainv1alpha1.Condition, contains toolchainv1alpha1.Condition) bool {
```
alexeykazakov, line 182: Let's add a comment to the function explaining what it does.

alexeykazakov (commented):
Looks good overall. Just a couple of minor comments regarding missing function descriptions. And also it looks like you can add a bit more tests.

pkg/condition/condition.go
sbryzak: Done

pkg/condition/condition.go
sbryzak: Done

PR #398, opened 2024-05-06
KUBESAW-43: Introduce LastUpdatedTime as prefered one

fbm3307:
After making the `LastProbeTime` optional, introduce `LastUpdatedTime` as the preferred one

This is the next step after https://github.com/codeready-toolchain/api/pull/420 
This is related to 
- Member-Operator - https://github.com/codeready-toolchain/member-operator/pull/564
- Host-Operator - https://github.com/codeready-toolchain/host-operator/pull/1025


mfrancisc (approved):
Looks good to me 👍

pkg/status/toolchaincluster.go
```diff
@@ -47,9 +47,15 @@ func GetToolchainClusterConditions(logger logr.Logger, attrs ToolchainClusterAtt
 	foundLastProbeTime := false
 	for _, condition := range toolchainCluster.ClusterStatus.Conditions {
 		if condition.Type == toolchainv1alpha1.ToolchainClusterReady {
-			lastProbeTime = condition.LastProbeTime
+			lastProbeTime = func() metav1.Time {
+				if condition.LastUpdatedTime != nil {
+					return *condition.LastUpdatedTime
+				}
+				return condition.LastProbeTime
```
MatousJobanek, line 54: Could you please add/update the unit tests so it verifies that it give the LastUpdatedTime preference and if not set then it takes the LastProbeTime?
I mean, let's verify that it works when either from these two fields are not set. If I'm not mistaken, the current unit tests sets both fields

fbm3307: added it

MatousJobanek (approved):

PR #400, opened 2024-06-05
Remove the migration step to add labels to creds secret for ToolchainClusters.

metlos:
`$TITLE`.

With all the clusters already updated, we can remove this migration step. IMHO, we should KEEP the test for the presence of the labels in `toolchain-e2e` (https://github.com/codeready-toolchain/toolchain-e2e/blob/master/test/e2e/toolchaincluster_test.go#L43) until we actually flip the logic and create the `ToolchainCluster`s based on the secrets.


MatousJobanek (approved):
cool :+1:

MatousJobanek:
>  IMHO, we should KEEP the test for the presence of the labels in toolchain-e2e (https://github.com/codeready-toolchain/toolchain-e2e/blob/master/test/e2e/toolchaincluster_test.go#L43) until we actually flip the logic and create the ToolchainClusters based on the secrets.

fully agree with you :+1:

PR #386, opened 2024-04-10
KUBESAW-12: Convert the health-check goroutine into ToolchainCluster controller

fbm3307:
This is to convert the health-check goroutine into ToolchainCluster controller

The related PRs: 

- Host-operator - https://github.com/codeready-toolchain/host-operator/pull/1010
- Member-Operator - https://github.com/codeready-toolchain/member-operator/pull/559

The Other PRs where only Toolchain common dependency is updated 
- Toolchain-e2e - https://github.com/codeready-toolchain/toolchain-e2e/pull/949
- Registration-service - https://github.com/codeready-toolchain/registration-service/pull/422
- Sandbox-sre - https://github.com/codeready-toolchain/sandbox-sre/pull/1698

PR #401, opened 2024-06-06
KUBESAW-100: Refactoring toolchaincluster controller

fbm3307:
This PR is to refactor the ToolchainCluster controller .

this is related to this [discussion](https://github.com/codeready-toolchain/toolchain-common/pull/386#discussion_r1571030991) and [this ](https://redhat-internal.slack.com/archives/C06MJ2DVBU4/p1713430967057339?thread_ts=1713362133.858979&cid=C06MJ2DVBU4) . 

Related PR where only change is updating the dependency But paired E2e has changes as Offline condition is dropped
- Host-operator - https://github.com/codeready-toolchain/host-operator/pull/1049
   - Paired e2e - https://github.com/codeready-toolchain/toolchain-e2e/pull/1005
- Member-operator - https://github.com/codeready-toolchain/member-operator/pull/579
  - Paired e2e - https://github.com/codeready-toolchain/toolchain-e2e/pull/1006
- Registration-Service - https://github.com/codeready-toolchain/registration-service/pull/437
  - Paired-e2e - https://github.com/codeready-toolchain/toolchain-e2e/pull/1000
- Sandbox-SRE - https://github.com/codeready-toolchain/sandbox-sre/pull/1830
- KSCTL - https://github.com/kubesaw/ksctl/pull/42

### DIFF
--- a/controllers/toolchaincluster/healthchecker.go
+++ b/controllers/toolchaincluster/healthchecker.go
@@ -19,8 +19,8 @@ const (
 	clusterReachableMsg    = "cluster is reachable"
 )
 
-// getClusterHealthStatus gets the kubernetes cluster health status by requesting "/healthz"
-func getClusterHealthStatus(ctx context.Context, remoteClusterClientset *kubeclientset.Clientset) []v1alpha1.Condition {
+// GetClusterHealthStatus gets the kubernetes cluster health status by requesting "/healthz"
+func GetClusterHealthStatus(ctx context.Context, remoteClusterClientset *kubeclientset.Clientset) []v1alpha1.Condition {
 	lgr := log.FromContext(ctx)
 	conditions := []v1alpha1.Condition{}
 	body, err := remoteClusterClientset.DiscoveryClient.RESTClient().Get().AbsPath("/healthz").Do(ctx).Raw()
--- a/controllers/toolchaincluster/healthchecker.go
+++ b/controllers/toolchaincluster/healthchecker.go
@@ -4,9 +4,9 @@ import (
 	"context"
 	"strings"
 
+	"github.com/codeready-toolchain/api/api/v1alpha1"
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
 	"github.com/go-logr/logr"
-	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kubeclientset "k8s.io/client-go/kubernetes"
@@ -27,41 +27,21 @@ type HealthChecker struct {
 	logger                 logr.Logger
 }
 
-func (hc *HealthChecker) updateIndividualClusterStatus(ctx context.Context, toolchainCluster *toolchainv1alpha1.ToolchainCluster) error {
-
-	currentClusterStatus := hc.getClusterHealthStatus(ctx)
-
-	for index, currentCond := range currentClusterStatus.Conditions {
-		for _, previousCond := range toolchainCluster.Status.Conditions {
-			if currentCond.Type == previousCond.Type && currentCond.Status == previousCond.Status {
-				currentClusterStatus.Conditions[index].LastTransitionTime = previousCond.LastTransitionTime
-			}
-		}
-	}
-
-	toolchainCluster.Status = *currentClusterStatus
-	if err := hc.localClusterClient.Status().Update(ctx, toolchainCluster); err != nil {
-		return errors.Wrapf(err, "Failed to update the status of cluster %s", toolchainCluster.Name)
-	}
-	return nil
-}
-
 // getClusterHealthStatus gets the kubernetes cluster health status by requesting "/healthz"
-func (hc *HealthChecker) getClusterHealthStatus(ctx context.Context) *toolchainv1alpha1.ToolchainClusterStatus {
-	clusterStatus := toolchainv1alpha1.ToolchainClusterStatus{}
+func (hc *HealthChecker) getClusterHealthStatus(ctx context.Context) []v1alpha1.Condition {
+	conditions := []v1alpha1.Condition{}
 	body, err := hc.remoteClusterClientset.DiscoveryClient.RESTClient().Get().AbsPath("/healthz").Do(ctx).Raw()
 	if err != nil {
 		hc.logger.Error(err, "Failed to do cluster health check for a ToolchainCluster")
-		clusterStatus.Conditions = append(clusterStatus.Conditions, clusterOfflineCondition())
+		conditions = append(conditions, clusterOfflineCondition())
 	} else {
 		if !strings.EqualFold(string(body), "ok") {
-			clusterStatus.Conditions = append(clusterStatus.Conditions, clusterNotReadyCondition(), clusterNotOfflineCondition())
+			conditions = append(conditions, clusterNotReadyCondition(), clusterNotOfflineCondition())
 		} else {
-			clusterStatus.Conditions = append(clusterStatus.Conditions, clusterReadyCondition())
+			conditions = append(conditions, clusterReadyCondition())
 		}
 	}
-
-	return &clusterStatus
+	return conditions
 }
 
 func clusterReadyCondition() toolchainv1alpha1.Condition {
--- a/controllers/toolchaincluster/healthchecker.go
+++ b/controllers/toolchaincluster/healthchecker.go
@@ -9,10 +9,8 @@ import (
 )
 
 const (
-	healthzOk              = "/healthz responded with ok"
-	healthzNotOk           = "/healthz responded without ok"
-	clusterNotReachableMsg = "cluster is not reachable"
-	clusterReachableMsg    = "cluster is reachable"
+	healthzOk    = "/healthz responded with ok"
+	healthzNotOk = "/healthz responded without ok"
 )
 
 // getClusterHealth gets the kubernetes cluster health status by requesting "/healthz"
--- a/controllers/toolchaincluster/healthchecker.go
+++ b/controllers/toolchaincluster/healthchecker.go
@@ -15,8 +15,8 @@ const (
 	clusterReachableMsg    = "cluster is reachable"
 )
 
-// GetClusterHealth gets the kubernetes cluster health status by requesting "/healthz"
-func GetClusterHealth(ctx context.Context, remoteClusterClientset *kubeclientset.Clientset) (bool, error) {
+// getClusterHealth gets the kubernetes cluster health status by requesting "/healthz"
+func getClusterHealthStatus(ctx context.Context, remoteClusterClientset *kubeclientset.Clientset) (bool, error) {
 
 	lgr := log.FromContext(ctx)
 	body, err := remoteClusterClientset.DiscoveryClient.RESTClient().Get().AbsPath("/healthz").Do(ctx).Raw()
--- a/controllers/toolchaincluster/healthchecker.go
+++ b/controllers/toolchaincluster/healthchecker.go
@@ -4,10 +4,6 @@ import (
 	"context"
 	"strings"
 
-	"github.com/codeready-toolchain/api/api/v1alpha1"
-	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
-	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kubeclientset "k8s.io/client-go/kubernetes"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
@@ -19,68 +15,19 @@ const (
 	clusterReachableMsg    = "cluster is reachable"
 )
 
-// GetClusterHealthStatus gets the kubernetes cluster health status by requesting "/healthz"
-func GetClusterHealthStatus(ctx context.Context, remoteClusterClientset *kubeclientset.Clientset) []v1alpha1.Condition {
+// GetClusterHealth gets the kubernetes cluster health status by requesting "/healthz"
+func GetClusterHealth(ctx context.Context, remoteClusterClientset *kubeclientset.Clientset) (bool, error) {
+
 	lgr := log.FromContext(ctx)
-	conditions := []v1alpha1.Condition{}
 	body, err := remoteClusterClientset.DiscoveryClient.RESTClient().Get().AbsPath("/healthz").Do(ctx).Raw()
 	if err != nil {
 		lgr.Error(err, "Failed to do cluster health check for a ToolchainCluster")
-		conditions = append(conditions, clusterOfflineCondition())
+		return false, err
 	} else {
 		if !strings.EqualFold(string(body), "ok") {
-			conditions = append(conditions, clusterNotReadyCondition(), clusterNotOfflineCondition())
+			return false, nil
 		} else {
-			conditions = append(conditions, clusterReadyCondition())
+			return true, nil
 		}
-	}
-	return conditions
-}
-
-func clusterReadyCondition() toolchainv1alpha1.Condition {
-	currentTime := metav1.Now()
-	return toolchainv1alpha1.Condition{
-		Type:               toolchainv1alpha1.ConditionReady,
-		Status:             corev1.ConditionTrue,
-		Reason:             toolchainv1alpha1.ToolchainClusterClusterReadyReason,
-		Message:            healthzOk,
-		LastUpdatedTime:    &currentTime,
-		LastTransitionTime: currentTime,
-	}
-}
-
-func clusterNotReadyCondition() toolchainv1alpha1.Condition {
-	currentTime := metav1.Now()
-	return toolchainv1alpha1.Condition{
-		Type:               toolchainv1alpha1.ConditionReady,
-		Status:             corev1.ConditionFalse,
-		Reason:             toolchainv1alpha1.ToolchainClusterClusterNotReadyReason,
-		Message:            healthzNotOk,
-		LastUpdatedTime:    &currentTime,
-		LastTransitionTime: currentTime,
-	}
-}
-
-func clusterOfflineCondition() toolchainv1alpha1.Condition {
-	currentTime := metav1.Now()
-	return toolchainv1alpha1.Condition{
-		Type:               toolchainv1alpha1.ToolchainClusterOffline,
-		Status:             corev1.ConditionTrue,
-		Reason:             toolchainv1alpha1.ToolchainClusterClusterNotReachableReason,
-		Message:            clusterNotReachableMsg,
-		LastUpdatedTime:    &currentTime,
-		LastTransitionTime: currentTime,
-	}
-}
-
-func clusterNotOfflineCondition() toolchainv1alpha1.Condition {
-	currentTime := metav1.Now()
-	return toolchainv1alpha1.Condition{
-		Type:               toolchainv1alpha1.ToolchainClusterOffline,
-		Status:             corev1.ConditionFalse,
-		Reason:             toolchainv1alpha1.ToolchainClusterClusterReachableReason,
-		Message:            clusterReachableMsg,
-		LastUpdatedTime:    &currentTime,
-		LastTransitionTime: currentTime,
 	}
 }
--- a/controllers/toolchaincluster/healthchecker.go
+++ b/controllers/toolchaincluster/healthchecker.go
@@ -23,11 +23,7 @@ func getClusterHealthStatus(ctx context.Context, remoteClusterClientset *kubecli
 	if err != nil {
 		lgr.Error(err, "Failed to do cluster health check for a ToolchainCluster")
 		return false, err
-	} else {
-		if !strings.EqualFold(string(body), "ok") {
-			return false, nil
-		} else {
-			return true, nil
-		}
 	}
+	return strings.EqualFold(string(body), "ok"), nil
+
 }
--- a/controllers/toolchaincluster/healthchecker.go
+++ b/controllers/toolchaincluster/healthchecker.go
@@ -6,11 +6,10 @@ import (
 
 	"github.com/codeready-toolchain/api/api/v1alpha1"
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
-	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kubeclientset "k8s.io/client-go/kubernetes"
-	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 const (
@@ -20,19 +19,13 @@ const (
 	clusterReachableMsg    = "cluster is reachable"
 )
 
-type HealthChecker struct {
-	localClusterClient     client.Client
-	remoteClusterClient    client.Client
-	remoteClusterClientset *kubeclientset.Clientset
-	logger                 logr.Logger
-}
-
 // getClusterHealthStatus gets the kubernetes cluster health status by requesting "/healthz"
-func (hc *HealthChecker) getClusterHealthStatus(ctx context.Context) []v1alpha1.Condition {
+func getClusterHealthStatus(ctx context.Context, remoteClusterClientset *kubeclientset.Clientset) []v1alpha1.Condition {
+	lgr := log.FromContext(ctx)
 	conditions := []v1alpha1.Condition{}
-	body, err := hc.remoteClusterClientset.DiscoveryClient.RESTClient().Get().AbsPath("/healthz").Do(ctx).Raw()
+	body, err := remoteClusterClientset.DiscoveryClient.RESTClient().Get().AbsPath("/healthz").Do(ctx).Raw()
 	if err != nil {
-		hc.logger.Error(err, "Failed to do cluster health check for a ToolchainCluster")
+		lgr.Error(err, "Failed to do cluster health check for a ToolchainCluster")
 		conditions = append(conditions, clusterOfflineCondition())
 	} else {
 		if !strings.EqualFold(string(body), "ok") {
--- a/controllers/toolchaincluster/healthchecker_test.go
+++ b/controllers/toolchaincluster/healthchecker_test.go
@@ -97,7 +97,7 @@ func TestClusterHealthChecks(t *testing.T) {
 			require.NoError(t, err)
 
 			//when
-			hcond := getClusterHealthStatus(context.TODO(), cacheclient)
+			hcond := GetClusterHealthStatus(context.TODO(), cacheclient)
 
 			//then
 			assert.Len(t, tc.clusterconditions, len(hcond))
--- a/controllers/toolchaincluster/healthchecker_test.go
+++ b/controllers/toolchaincluster/healthchecker_test.go
@@ -5,22 +5,21 @@ import (
 	"testing"
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
+	"github.com/codeready-toolchain/toolchain-common/pkg/cluster"
 	"github.com/codeready-toolchain/toolchain-common/pkg/test"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"gopkg.in/h2non/gock.v1"
 	corev1 "k8s.io/api/core/v1"
+	kubeclientset "k8s.io/client-go/kubernetes"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
-
-var logger = logf.Log.WithName("toolchaincluster_healthcheck")
 
 func TestClusterHealthChecks(t *testing.T) {
 
 	// given
 	defer gock.Off()
-	//tcNs := "test-namespace"
+	tcNs := "test-namespace"
 	gock.New("http://cluster.com").
 		Get("healthz").
 		Persist().
@@ -43,89 +42,86 @@ func TestClusterHealthChecks(t *testing.T) {
 		status            toolchainv1alpha1.ToolchainClusterStatus
 	}{
 		//ToolchainCluster.status doesn't contain any conditions
-		"UnstableNoCondition": {
-			tctype:            "unstable",
-			apiendpoint:       "http://unstable.com",
-			clusterconditions: []toolchainv1alpha1.Condition{unhealthy(), notOffline()},
-			status:            toolchainv1alpha1.ToolchainClusterStatus{},
-		},
+		// "UnstableNoCondition": {
+		// 	tctype:            "unstable",
+		// 	apiendpoint:       "http://unstable.com",
+		// 	clusterconditions: []toolchainv1alpha1.Condition{unhealthy(), notOffline()},
+		// 	status:            toolchainv1alpha1.ToolchainClusterStatus{},
+		// },
 		"StableNoCondition": {
 			tctype:            "stable",
 			apiendpoint:       "http://cluster.com",
 			clusterconditions: []toolchainv1alpha1.Condition{healthy()},
 			status:            toolchainv1alpha1.ToolchainClusterStatus{},
 		},
-		"NotFoundNoCondition": {
-			tctype:            "not-found",
-			apiendpoint:       "http://not-found.com",
-			clusterconditions: []toolchainv1alpha1.Condition{offline()},
-			status:            toolchainv1alpha1.ToolchainClusterStatus{},
-		},
+		// "NotFoundNoCondition": {
+		// 	tctype:            "not-found",
+		// 	apiendpoint:       "http://not-found.com",
+		// 	clusterconditions: []toolchainv1alpha1.Condition{offline()},
+		// 	status:            toolchainv1alpha1.ToolchainClusterStatus{},
+		// },
 		//ToolchainCluster.status already contains conditions
-		"UnstableContainsCondition": {
-			tctype:            "unstable",
-			apiendpoint:       "http://unstable.com",
-			clusterconditions: []toolchainv1alpha1.Condition{unhealthy(), notOffline()},
-			status:            withStatus(healthy()),
-		},
-		"StableContainsCondition": {
-			tctype:            "stable",
-			apiendpoint:       "http://cluster.com",
-			clusterconditions: []toolchainv1alpha1.Condition{healthy()},
-			status:            withStatus(offline()),
-		},
-		"NotFoundContainsCondition": {
-			tctype:            "not-found",
-			apiendpoint:       "http://not-found.com",
-			clusterconditions: []toolchainv1alpha1.Condition{offline()},
-			status:            withStatus(healthy()),
-		},
-		//if the connection cannot be established at beginning, then it should be offline
-		"OfflineConnectionNotEstablished": {
-			tctype:            "failing",
-			apiendpoint:       "http://failing.com",
-			clusterconditions: []toolchainv1alpha1.Condition{offline()},
-			status:            toolchainv1alpha1.ToolchainClusterStatus{},
-		},
-		//if no zones nor region is retrieved, then keep the current
-		"NoZoneKeepCurrent": {
-			tctype:            "stable",
-			apiendpoint:       "http://cluster.com",
-			clusterconditions: []toolchainv1alpha1.Condition{healthy()},
-			status:            withStatus(offline()),
-		},
+		// "UnstableContainsCondition": {
+		// 	tctype:            "unstable",
+		// 	apiendpoint:       "http://unstable.com",
+		// 	clusterconditions: []toolchainv1alpha1.Condition{unhealthy(), notOffline()},
+		// 	status:            withStatus(healthy()),
+		// },
+		// "StableContainsCondition": {
+		// 	tctype:            "stable",
+		// 	apiendpoint:       "http://cluster.com",
+		// 	clusterconditions: []toolchainv1alpha1.Condition{healthy()},
+		// 	status:            withStatus(offline()),
+		// },
+		// "NotFoundContainsCondition": {
+		// 	tctype:            "not-found",
+		// 	apiendpoint:       "http://not-found.com",
+		// 	clusterconditions: []toolchainv1alpha1.Condition{offline()},
+		// 	status:            withStatus(healthy()),
+		// },
+		// //if the connection cannot be established at beginning, then it should be offline
+		// "OfflineConnectionNotEstablished": {
+		// 	tctype:            "failing",
+		// 	apiendpoint:       "http://failing.com",
+		// 	clusterconditions: []toolchainv1alpha1.Condition{offline()},
+		// 	status:            toolchainv1alpha1.ToolchainClusterStatus{},
+		// },
+		// //if no zones nor region is retrieved, then keep the current
+		// "NoZoneKeepCurrent": {
+		// 	tctype:            "stable",
+		// 	apiendpoint:       "http://cluster.com",
+		// 	clusterconditions: []toolchainv1alpha1.Condition{healthy()},
+		// 	status:            withStatus(offline()),
+		// },
 	}
-	for k, _ := range tests {
+	for k, tc := range tests {
 		t.Run(k, func(t *testing.T) {
-			// tctype, sec := newToolchainCluster(tc.tctype, tcNs, tc.apiendpoint, tc.status)
-			// cl := test.NewFakeClient(t, tctype, sec)
-			// reset := setupCachedClusters(t, cl, tctype)
-			// defer reset()
-			// cachedtc, found := cluster.GetCachedToolchainCluster(tctype.Name)
-			// require.True(t, found)
-			// cacheclient, err := kubeclientset.NewForConfig(cachedtc.RestConfig)
-			// require.NoError(t, err)
-			// healthChecker := &HealthChecker{
-			// 	localClusterClient:     cl,
-			// 	remoteClusterClient:    cachedtc.Client,
-			// 	remoteClusterClientset: cacheclient,
-			// 	logger:                 logger,
-			// }
-			// when
-			//err = healthChecker.updateIndividualClusterStatus(context.TODO(), tctype)
+			tctype, sec := newToolchainCluster(tc.tctype, tcNs, tc.apiendpoint, tc.status)
+			cl := test.NewFakeClient(t, tctype, sec)
+			reset := setupCachedClusters(t, cl, tctype)
+			defer reset()
+			cachedtc, found := cluster.GetCachedToolchainCluster(tctype.Name)
+			require.True(t, found)
+			cacheclient, err := kubeclientset.NewForConfig(cachedtc.RestConfig)
+			require.NoError(t, err)
 
+			//when
+			hcond := getClusterHealthStatus(context.TODO(), cacheclient)
+			//fmt.Printf("hcond %+v", hcond)
 			//then
-			// require.NoError(t, err)
-			// assertClusterStatus(t, cl, tc.tctype, tc.clusterconditions...)
+			require.NoError(t, err)
+			//fmt.Printf("tc cond %+v", tc.clusterconditions)
+			//assertClusterStatus(t, cl, tc.tctype, hcond...)
+			assert.Equal(t, hcond[0].Status, tc.clusterconditions[0].Status)
 		})
 	}
 }
 
-func withStatus(conditions ...toolchainv1alpha1.Condition) toolchainv1alpha1.ToolchainClusterStatus {
-	return toolchainv1alpha1.ToolchainClusterStatus{
-		Conditions: conditions,
-	}
-}
+//	func withStatus(conditions ...toolchainv1alpha1.Condition) toolchainv1alpha1.ToolchainClusterStatus {
+//		return toolchainv1alpha1.ToolchainClusterStatus{
+//			Conditions: conditions,
+//		}
+//	}
 func assertClusterStatus(t *testing.T, cl client.Client, clusterName string, clusterConds ...toolchainv1alpha1.Condition) {
 	tc := &toolchainv1alpha1.ToolchainCluster{}
 	err := cl.Get(context.TODO(), test.NamespacedName("test-namespace", clusterName), tc)
--- a/controllers/toolchaincluster/healthchecker_test.go
+++ b/controllers/toolchaincluster/healthchecker_test.go
@@ -34,49 +34,49 @@ func TestClusterHealthChecks(t *testing.T) {
 		Reply(404)
 
 	tests := map[string]struct {
-		tctype      string
-		apiendpoint string
-		healthcheck bool
-		errh        error
+		tcType      string
+		apiEndPoint string
+		healthCheck bool
+		err         error
 	}{
 		"HealthOkay": {
-			tctype:      "stable",
-			apiendpoint: "http://cluster.com",
-			healthcheck: true,
+			tcType:      "stable",
+			apiEndPoint: "http://cluster.com",
+			healthCheck: true,
 		},
 		"HealthNotOkayButNoError": {
-			tctype:      "unstable",
-			apiendpoint: "http://unstable.com",
-			healthcheck: false,
+			tcType:      "unstable",
+			apiEndPoint: "http://unstable.com",
+			healthCheck: false,
 		},
 		"ErrorWhileDoingHealth": {
-			tctype:      "Notfound",
-			apiendpoint: "http://not-found.com",
-			healthcheck: false,
-			errh:        fmt.Errorf("the server could not find the requested resource"),
+			tcType:      "Notfound",
+			apiEndPoint: "http://not-found.com",
+			healthCheck: false,
+			err:         fmt.Errorf("the server could not find the requested resource"),
 		},
 	}
 	for k, tc := range tests {
 		t.Run(k, func(t *testing.T) {
 			//given
-			tctype, sec := newToolchainCluster(tc.tctype, tcNs, tc.apiendpoint, toolchainv1alpha1.ToolchainClusterStatus{})
-			cl := test.NewFakeClient(t, tctype, sec)
-			reset := setupCachedClusters(t, cl, tctype)
+			tcType, sec := newToolchainCluster(tc.tcType, tcNs, tc.apiEndPoint, toolchainv1alpha1.ToolchainClusterStatus{})
+			cl := test.NewFakeClient(t, tcType, sec)
+			reset := setupCachedClusters(t, cl, tcType)
 			defer reset()
-			cachedtc, found := cluster.GetCachedToolchainCluster(tctype.Name)
+			cachedTC, found := cluster.GetCachedToolchainCluster(tcType.Name)
 			require.True(t, found)
-			cacheclient, err := kubeclientset.NewForConfig(cachedtc.RestConfig)
+			cacheClient, err := kubeclientset.NewForConfig(cachedTC.RestConfig)
 			require.NoError(t, err)
 
 			//when
-			healthcheck, errh := getClusterHealthStatus(context.TODO(), cacheclient)
+			healthCheck, err := getClusterHealthStatus(context.TODO(), cacheClient)
 
 			//then
-			require.Equal(t, tc.healthcheck, healthcheck)
-			if tc.errh != nil {
-				require.EqualError(t, errh, tc.errh.Error())
+			require.Equal(t, tc.healthCheck, healthCheck)
+			if tc.err != nil {
+				require.EqualError(t, err, tc.err.Error())
 			} else {
-				require.NoError(t, errh)
+				require.NoError(t, err)
 			}
 
 		})
--- a/controllers/toolchaincluster/healthchecker_test.go
+++ b/controllers/toolchaincluster/healthchecker_test.go
@@ -7,7 +7,6 @@ import (
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
 	"github.com/codeready-toolchain/toolchain-common/pkg/cluster"
 	"github.com/codeready-toolchain/toolchain-common/pkg/test"
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"gopkg.in/h2non/gock.v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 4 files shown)


fbm3307:
Raised this PR to get reviews on the refactoring of controller. Would fix the test cases once we have finalised the refactoring

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -75,20 +78,52 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+		return r.overwriteRunHealthcheck(ctx, lcl, rcl, rclset, tc, lgr)
+	}
+	healthChecker := &HealthChecker{
+		localClusterClient:     lcl,
+		remoteClusterClient:    rcl,
+		remoteClusterClientset: rclset,
+		logger:                 lgr,
+	}
+	// update the status of the individual cluster.
+	clstatus := healthChecker.getClusterHealthStatus(ctx)
```
MatousJobanek, line 106: is there any reason for keeping the `HealthChecker` type and not doing the check directly here in this method?

MatousJobanek: apart from that, some of the parameters are not needed anymore

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -75,20 +78,52 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
 	// update the status of the individual cluster.
-	if err := healthChecker.updateIndividualClusterStatus(ctx, toolchainCluster); err != nil {
+	if err := r.updateStatus(ctx, toolchainCluster, healthcheckResult); err != nil {
 		reqLogger.Error(err, "unable to update cluster status of ToolchainCluster")
 		return reconcile.Result{}, err
 	}
-
 	return reconcile.Result{RequeueAfter: r.RequeAfter}, nil
 }
+func (r *Reconciler) runHealthcheck(ctx context.Context, lcl client.Client, rcl client.Client, rclset *kubeclientset.Clientset, tc *toolchainv1alpha1.ToolchainCluster, lgr logr.Logger) ([]toolchainv1alpha1.Condition, error) {
```
MatousJobanek, line 95: the `tc *toolchainv1alpha1.ToolchainCluster` is not used anywhere.
the logger can be also dropped - we can always construct the logger from the context

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -19,9 +21,10 @@ import (
 
 // Reconciler reconciles a ToolchainCluster object
 type Reconciler struct {
-	Client     client.Client
-	Scheme     *runtime.Scheme
-	RequeAfter time.Duration
+	Client                  client.Client
+	Scheme                  *runtime.Scheme
+	RequeAfter              time.Duration
+	overwriteRunHealthcheck func(ctx context.Context, lcl client.Client, rcl client.Client, rclset *kubeclientset.Clientset, tc *toolchainv1alpha1.ToolchainCluster, lgr logr.Logger) ([]toolchainv1alpha1.Condition, error)
```
MatousJobanek, line 27: let's not introduce any "overwrite" logic; instead of it:
* convert the `runHealthcheck` method into a function 
* rename this parameter to `CheckHealth` or something like that
* either in main.go pass the `runHealthcheck` function as a value of the `CheckHealth` param of the Reconciler, or set it as the default value as part of the `SetupWithManager` method.

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -75,20 +78,52 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+
+func (r *Reconciler) updateStatus(ctx context.Context, toolchainCluster *toolchainv1alpha1.ToolchainCluster, currentconditions []toolchainv1alpha1.Condition) error {
+
+	for index, currentCond := range currentconditions {
+		for _, previousCond := range toolchainCluster.Status.Conditions {
+			if currentCond.Type == previousCond.Type && currentCond.Status == previousCond.Status {
+				currentconditions[index].LastTransitionTime = previousCond.LastTransitionTime
+			}
+		}
+	}
```
MatousJobanek, line 119: why do we need this logic? it looks like you could use the common logic instead  https://github.com/codeready-toolchain/toolchain-common/blob/6a80f659afcfe1652aa4c44e8f4b5da01f38f95f/pkg/condition/condition.go#L14

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -75,20 +78,52 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 		reqLogger.Error(err, "cannot create ClientSet for the ToolchainCluster")
 		return reconcile.Result{}, err
```
MatousJobanek, line 79: if we fail to create the client, then we should make sure that the ready condition is set to false

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -75,20 +78,52 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+	}
+	healthChecker := &HealthChecker{
+		localClusterClient:     lcl,
+		remoteClusterClient:    rcl,
+		remoteClusterClientset: rclset,
+		logger:                 lgr,
+	}
+	// update the status of the individual cluster.
+	clstatus := healthChecker.getClusterHealthStatus(ctx)
+	return clstatus, nil
```
MatousJobanek, line 107: the only returned value for the error seems to be `nil`, so I guess that you can remove it

MatousJobanek (commented):
apart from the comments that I left there, it looks like you still keep resetting the conditions if the cached client wasn't found 
https://github.com/codeready-toolchain/toolchain-common/blob/466372121c853170a21c7874694c0d41dd53d94f/controllers/toolchaincluster/toolchaincluster_controller.go#L69
let's update the conditions instead (and also make sure that the "Ready" condition is set to false)

controllers/toolchaincluster/toolchaincluster_controller.go
fbm3307: okay removed them

controllers/toolchaincluster/toolchaincluster_controller.go
fbm3307: yep done

controllers/toolchaincluster/toolchaincluster_controller.go
fbm3307: Okay This would be easier, 
i used the overwrite for ease of testing though

controllers/toolchaincluster/toolchaincluster_controller.go
fbm3307: done

controllers/toolchaincluster/toolchaincluster_controller.go
fbm3307: okay

controllers/toolchaincluster/toolchaincluster_controller.go
fbm3307: i have tried to use `AddorUpdateStatusConditionsWithLastUpdatedTimestamp` instead

controllers/toolchaincluster/toolchaincluster_controller.go
fbm3307: I did update it, but testing this functionality doesn't make sense, as we never encounter this scenario.

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -54,29 +57,45 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
 			reqLogger.Error(err, "failed to update the status of ToolchainCluster")
 		}
 		return reconcile.Result{}, err
 	}
 
+	//Not testing as this as this is being tested in cluster package
 	clientSet, err := kubeclientset.NewForConfig(cachedCluster.RestConfig)
 	if err != nil {
 		reqLogger.Error(err, "cannot create ClientSet for the ToolchainCluster")
+		toolchainCluster.Status.Conditions = condition.AddOrUpdateStatusConditionsWithLastUpdatedTimestamp(toolchainCluster.Status.Conditions, clusterNotReadyCondition())
```
filariow, line 71: Updating the conditions in the structure without updating the status won't have any effect. So, we should either persist the change with `r.Client.Status().Update(...)` (maybe `r.updateStatus`?) or we can remove this line.

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -54,29 +57,45 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	cachedCluster, ok := cluster.GetCachedToolchainCluster(toolchainCluster.Name)
 	if !ok {
 		err := fmt.Errorf("cluster %s not found in cache", toolchainCluster.Name)
-		toolchainCluster.Status.Conditions = []toolchainv1alpha1.Condition{clusterOfflineCondition()}
+		toolchainCluster.Status.Conditions = condition.AddOrUpdateStatusConditionsWithLastUpdatedTimestamp(toolchainCluster.Status.Conditions, clusterOfflineCondition())
```
filariow, line 60: would it make sense to use `r.updateStatus(ctx, toolchainCluster, clusterOfflineCondition)` here?

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -54,29 +57,45 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+
+func (r *Reconciler) runCheckHealthOrDefault(ctx context.Context, rcc *kubeclientset.Clientset) []toolchainv1alpha1.Condition {
+	if r.CheckHealth != nil {
+		return r.CheckHealth(ctx, rcc)
+	}
+	hcond := getClusterHealthStatus(ctx, rcc)
+	return hcond
+}
+
+func (r *Reconciler) updateStatus(ctx context.Context, toolchainCluster *toolchainv1alpha1.ToolchainCluster, currentconditions []toolchainv1alpha1.Condition) error {
```
filariow, line 94: for applying changes proposed in previous comments, introducing here a `params` may be helpful
```suggestion
func (r *Reconciler) updateStatus(ctx context.Context, toolchainCluster *toolchainv1alpha1.ToolchainCluster, currentconditions ...toolchainv1alpha1.Condition) error {
```

controllers/toolchaincluster/toolchaincluster_controller_test.go
```diff
@@ -89,6 +90,7 @@ func TestClusterControllerChecks(t *testing.T) {
 
 		cl := test.NewFakeClient(t, stable, sec)
 		reset := setupCachedClusters(t, cl, stable)
+		stable.Status.Conditions = condition.AddOrUpdateStatusConditionsWithLastUpdatedTimestamp(stable.Status.Conditions, clusterReadyCondition())
```
filariow, line 93: is this required?

controllers/toolchaincluster/toolchaincluster_controller_test.go
```diff
@@ -147,6 +207,21 @@ func newToolchainCluster(name, tcNs string, apiEndpoint string, status toolchain
 }
 
 func prepareReconcile(toolchainCluster *toolchainv1alpha1.ToolchainCluster, cl *test.FakeClient, requeAfter time.Duration) (Reconciler, reconcile.Request) {
+	controller := Reconciler{
+		Client:     cl,
+		Scheme:     scheme.Scheme,
+		RequeAfter: requeAfter,
+		CheckHealth: func(context.Context, *kubeclientset.Clientset) []toolchainv1alpha1.Condition {
+			return toolchainCluster.Status.Conditions
+		},
```
filariow, line 216: This is a bit misleading to me. I read it as "return the conditions the cluster already has".

I'd suggest to remove the `prepareCheckHealthDefaultReconcile` function and to update the `prepareReconcile` function to require the function to use as `CheckHealth`. This will make harder to inadvertently test the healthchecker logic from this suite.

WDYT?

controllers/toolchaincluster/toolchaincluster_controller_test.go
```diff
@@ -101,23 +103,81 @@ func TestClusterControllerChecks(t *testing.T) {
 		assertClusterStatus(t, cl, "stable", healthy())
 	})
 
-	t.Run("toolchain cluster cache not found", func(t *testing.T) {
+	t.Run("Checking the run check health default ", func(t *testing.T) {
```
filariow, line 106: I feel like this should not be tested here as it is testing the Healthchecker code.
Indeed, if the `CheckHealth` field is not set, we are testing the Healthcheker code that's executed at `controllers/toolchaincluster/toolchaincluster_controller.go#L86`.

To me in this suite we should just test that (1) if we got an error we are forwarding the error and (2) if we got a set of conditions we are setting the condition in the status.

Another thing that we'd be interested in testing is "if CheckHealth is not tested, we are invoking the healthchecker", but it's seems to me is a quite hard to test in unit tests and not being that useful. E2E can easily cover this scenario.

filariow (commented):
well done, just a couple of ideas for tests 👍🏾

controllers/toolchaincluster/toolchaincluster_controller_test.go
fbm3307: no i think i can set it while creating the cluster

controllers/toolchaincluster/toolchaincluster_controller_test.go
fbm3307: This is to check second the scenario in reconciler, where we would pass the conditions when we call the reconciler. i wouldnt wont to remove this, probably i can update the return conditions

controllers/toolchaincluster/toolchaincluster_controller.go
fbm3307: Can you elaborate more..? 
My understanding here is we update the status every time we poll to check healtth,  but inbetween we still want to have the condition updated if something breaks while checking the health. so we are updating the condition and not status.

controllers/toolchaincluster/toolchaincluster_controller_test.go
fbm3307: i think CheckHealth is a part of controller and checking it from somewhere else wouldn't make sense i think

filariow: yes, but for how the reconciler is designed if its `CheckHealth` field is not set, it would use the Healthchecker logic. This scenario inadvertently checks the Healthchecker logic.  

To verify it, you can 
```bash
go test ./controllers/toolchaincluster/... -run 'TestClusterControllerChecks/Checking_the_run_check_health_default_' -coverprofile cover.out
go tool cover -html=cover.out
```

the output should be similar to the following:
```diff
+func (r *Reconciler) runCheckHealthOrDefault(ctx context.Context, rcc *kubeclientset.Clientset) []toolchainv1alpha1.Condition {
+        if r.CheckHealth != nil {
-                return r.CheckHealth(ctx, rcc)
+        }
+        hcond := getClusterHealthStatus(ctx, rcc)
+        return hcond
+}
```

this means that we are invoking -hence testing- the healthchecker code. 

To me the healthchecker logic should be tested in isolation, in its own unit tests. Here we should just test that the reconciler handles correctly the different kind of result the func `r.runCheckHealthOrDefault(ctx, clientSet)` can return, i.e. empty set of conditions, conditions already in status, conditions not in status, etc.

To avoid testing the healthchecker code in the reconciler tests we should enforce the `CheckHealth` field to be always set.

controllers/toolchaincluster/toolchaincluster_controller.go
filariow: sure, basically L71 is updating the data in memory but we're never reaching out the database to persist this change because L72 returns. To persist the data in the database we need to do something like L60-L64.

It seems to me we have an auxiliary function that implements this logic (`r.updateStatus`) we can simplify and just invoke `r.updateStatus(ctx, toolchainCluster, clusterNotReadyCondition())`.

fbm3307: thanks for pointing this out,  i have changed it

controllers/toolchaincluster/toolchaincluster_controller.go
fbm3307: done

controllers/toolchaincluster/toolchaincluster_controller.go
fbm3307: done

controllers/toolchaincluster/toolchaincluster_controller_test.go
fbm3307: i think i have updated it now

controllers/toolchaincluster/toolchaincluster_controller_test.go
fbm3307: updated it

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -54,29 +57,38 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	cachedCluster, ok := cluster.GetCachedToolchainCluster(toolchainCluster.Name)
 	if !ok {
 		err := fmt.Errorf("cluster %s not found in cache", toolchainCluster.Name)
-		toolchainCluster.Status.Conditions = []toolchainv1alpha1.Condition{clusterOfflineCondition()}
-		if err := r.Client.Status().Update(ctx, toolchainCluster); err != nil {
-			reqLogger.Error(err, "failed to update the status of ToolchainCluster")
+		if err := r.updateStatus(ctx, toolchainCluster, clusterOfflineCondition()); err != nil {
+			reqLogger.Error(err, "unable to update cluster status of ToolchainCluster")
```
filariow, line 61: we can add this test for testing L61:

```go
	t.Run("error while updating a toolchain cluster on cache miss", func(t *testing.T) {
		// given
		unstable, _ := newToolchainCluster("unstable", tcNs, "http://cluster.com", toolchainv1alpha1.ToolchainClusterStatus{})
		unstable.Status.Conditions = condition.AddOrUpdateStatusConditionsWithLastUpdatedTimestamp(unstable.Status.Conditions, healthy())

		cl := test.NewFakeClient(t, unstable)
		suerr := fmt.Errorf("my test error")
		cl.MockStatusUpdate = func(ctx context.Context, obj runtimeclient.Object, opts ...runtimeclient.UpdateOption) error {
			return suerr
		}
		unstable.Status.Conditions = condition.AddOrUpdateStatusConditionsWithLastUpdatedTimestamp(unstable.Status.Conditions, clusterOfflineCondition())
		controller, req := prepareReconcile(unstable, cl, requeAfter)

		// when
		recresult, err := controller.Reconcile(context.TODO(), req)

		// then
		require.EqualError(t, err, "cluster unstable not found in cache")
		require.Equal(t, reconcile.Result{}, recresult)

		actualtoolchaincluster := &toolchainv1alpha1.ToolchainCluster{}
		err = cl.Client.Get(context.TODO(), types.NamespacedName{Name: "unstable", Namespace: tcNs}, actualtoolchaincluster)
		require.NoError(t, err)
		assertClusterStatus(t, cl, "unstable", healthy())
	})
```

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -54,29 +57,38 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
 	}
-
 	return reconcile.Result{RequeueAfter: r.RequeAfter}, nil
 }
+
+func (r *Reconciler) updateStatus(ctx context.Context, toolchainCluster *toolchainv1alpha1.ToolchainCluster, currentconditions ...toolchainv1alpha1.Condition) error {
+
+	toolchainCluster.Status.Conditions = condition.AddOrUpdateStatusConditionsWithLastUpdatedTimestamp(toolchainCluster.Status.Conditions, currentconditions...)
+	if err := r.Client.Status().Update(ctx, toolchainCluster); err != nil {
+		return errors.Wrapf(err, "Failed to update the status of cluster %s", toolchainCluster.Name)
```
filariow, line 91: we can add this test for covering L91:
```go
	t.Run("error while updating a toolchain cluster", func(t *testing.T) {
		// given
		stable, sec := newToolchainCluster("stable", tcNs, "http://cluster.com", withStatus(healthy()))
		suerr := fmt.Errorf("my test error")
		cl := test.NewFakeClient(t, stable, sec)
		cl.MockStatusUpdate = func(ctx context.Context, obj runtimeclient.Object, opts ...runtimeclient.UpdateOption) error {
			return suerr
		}
		reset := setupCachedClusters(t, cl, stable)

		defer reset()
		controller, req := prepareReconcile(stable, cl, requeAfter)

		// when
		recresult, err := controller.Reconcile(context.TODO(), req)

		// then
		require.EqualError(t, err, fmt.Sprintf("Failed to update the status of cluster %s: %v", stable.Name, suerr))
		require.Equal(t, reconcile.Result{}, recresult)
		assertClusterStatus(t, cl, "stable", healthy())
	})
```

controllers/toolchaincluster/toolchaincluster_controller_test.go
```diff
@@ -151,9 +153,36 @@ func prepareReconcile(toolchainCluster *toolchainv1alpha1.ToolchainCluster, cl *
 		Client:     cl,
 		Scheme:     scheme.Scheme,
 		RequeAfter: requeAfter,
+		CheckHealth: func(context.Context, *kubeclientset.Clientset) []toolchainv1alpha1.Condition {
+			return toolchainCluster.Status.Conditions
+		},
```
filariow, line 158: This is still hard to read to me. I still read it as "return the conditions the cluster already has". It feels a bit hacky. What if we pass this conditions in `prepareReconcile` instead? Something like the following:

```go
func prepareReconcile(toolchainCluster *toolchainv1alpha1.ToolchainCluster, cl *test.FakeClient, requeAfter time.Duration, healthConditions ...toolchainv1alpha1.Condition) (Reconciler, reconcile.Request) {
	controller := Reconciler{
		Client:     cl,
		Scheme:     scheme.Scheme,
		RequeAfter: requeAfter,
		CheckHealth: func(context.Context, *kubeclientset.Clientset) []toolchainv1alpha1.Condition {
			return healthConditions
		},
	}
	req := reconcile.Request{
		NamespacedName: test.NamespacedName(toolchainCluster.Namespace, toolchainCluster.Name),
	}
	return controller, req
}
```

WDYT?

filariow (changes requested):
largely LGTM, I'd like to propose a couple more test cases and a little refactor

controllers/toolchaincluster/toolchaincluster_controller.go
fbm3307: Thanks for this, i was trying to have this check in the same test as cache not found.. but i guess keep this separate would be good

controllers/toolchaincluster/toolchaincluster_controller.go
fbm3307: same here

controllers/toolchaincluster/toolchaincluster_controller_test.go
fbm3307: wouldnt it be the same..? 
we are going to pass conditions and i am passing the already available condition of the cluster..

filariow: the main difference is that right now the cluster status is equal to the status returned by the healthcheck. After this change the actual status of the cluster can differ from the value returned by the healthcheck.

filariow: basically we can not properly test the following scenarios:

```gherkin
Scenario: Conditions are updated
    Given a cluster with Condition C1
    When the healthcheck returns Condition C2
    Then the cluster is updated to have Condition C2
```

```gherkin
Scenario: Condition time is not updated
    Given a cluster with Condition C1
    When the healthcheck returns Condition C1
    Then the cluster has Condition C1
    And cluster Condition C1 LastTransitionTime is not updated
```

I think we should add these test case. I can not find them in the actual test suite.

fbm3307: this will be a integration test , as this cannot be tested in unit tests, as heathcheck will be invoked from reconciler..?

fbm3307: I have tried to add them

mfrancisc (commented):
Nice job 👍 

I have only some cosmetic comments.

controllers/toolchaincluster/healthchecker_test.go
```diff
@@ -38,140 +32,59 @@ func TestClusterHealthChecks(t *testing.T) {
...
-			require.NoError(t, err)
-			healthChecker := &HealthChecker{
-				localClusterClient:     cl,
-				remoteClusterClient:    cachedtc.Client,
-				remoteClusterClientset: cacheclient,
-				logger:                 logger,
-			}
-			// when
-			err = healthChecker.updateIndividualClusterStatus(context.TODO(), tctype)
+	t.Run("When cluster health is ok", func(t *testing.T) {
```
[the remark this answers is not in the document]
fbm3307: yeah.. updated it. now

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -69,26 +73,110 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 		return reconcile.Result{}, err
 	}
 
+	//Not testing as this as this is being tested in cluster package
```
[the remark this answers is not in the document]
fbm3307: yes

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -69,26 +73,110 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
 	clientSet, err := kubeclientset.NewForConfig(cachedCluster.RestConfig)
 	if err != nil {
 		reqLogger.Error(err, "cannot create ClientSet for the ToolchainCluster")
+		if err := r.updateStatus(ctx, toolchainCluster, clusterNotReadyCondition()); err != nil {
```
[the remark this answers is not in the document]
fbm3307: offline means, cluster is unreachable, while notready means, cluster is reachable but some problems while it is getting ready/healthy, Hence here the condition not ready, as cluster is reachable, but not completely healthy/ready

controllers/toolchaincluster/healthchecker.go
```diff
@@ -20,94 +15,19 @@ const (
...
-
-	for index, currentCond := range currentClusterStatus.Conditions {
-		for _, previousCond := range toolchainCluster.Status.Conditions {
-			if currentCond.Type == previousCond.Type && currentCond.Status == previousCond.Status {
-				currentClusterStatus.Conditions[index].LastTransitionTime = previousCond.LastTransitionTime
-			}
-		}
-	}
+// GetClusterHealth gets the kubernetes cluster health status by requesting "/healthz"
+func GetClusterHealth(ctx context.Context, remoteClusterClientset *kubeclientset.Clientset) (bool, error) {
```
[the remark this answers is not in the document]
fbm3307: done

controllers/toolchaincluster/toolchaincluster_controller_test.go
```diff
@@ -160,6 +236,50 @@ func TestClusterControllerChecks(t *testing.T) {
 	})
 }
 
+func TestGetClusterHealth(t *testing.T) {
+	t.Run("Checkhealth deafualt", func(t *testing.T) {
```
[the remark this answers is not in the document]
fbm3307: done

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -69,26 +73,110 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+		return nil
+	}
+	toolchainCluster.Status.Conditions = resultconditions
+	if err := r.Client.Status().Update(ctx, toolchainCluster); err != nil {
+		return errors.Wrapf(err, "Failed to update the status of cluster %s", toolchainCluster.Name)
+	}
+	return nil
+}
+
+func (r *Reconciler) GetClusterHealthCondition(ctx context.Context, remoteClusterClientset *kubeclientset.Clientset) []v1alpha1.Condition {
```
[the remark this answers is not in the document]
fbm3307: done

controllers/toolchaincluster/healthchecker.go
```diff
@@ -20,94 +15,19 @@ const (
...
+		return false, err
 	} else {
 		if !strings.EqualFold(string(body), "ok") {
-			clusterStatus.Conditions = append(clusterStatus.Conditions, clusterNotReadyCondition(), clusterNotOfflineCondition())
+			return false, nil
 		} else {
-			clusterStatus.Conditions = append(clusterStatus.Conditions, clusterReadyCondition())
+			return true, nil
 		}
 	}
```
filariow, line 32: if you like, this can be simplified as follows:

```go
	if err != nil {
		lgr.Error(err, "Failed to do cluster health check for a ToolchainCluster")
		return false, err
	}

	return strings.EqualFold(string(body), "ok"), nil
```

controllers/toolchaincluster/toolchaincluster_controller_test.go
```diff
@@ -107,22 +108,97 @@ func TestClusterControllerChecks(t *testing.T) {
...
 		// given
 		stable, _ := newToolchainCluster("stable", tcNs, "http://cluster.com", toolchainv1alpha1.ToolchainClusterStatus{})
+		stable.Status.Conditions = condition.AddOrUpdateStatusConditionsWithLastUpdatedTimestamp(stable.Status.Conditions, healthy())
 
 		cl := test.NewFakeClient(t, stable)
-
+		cl.MockStatusUpdate = func(ctx context.Context, obj runtimeclient.Object, opts ...runtimeclient.UpdateOption) error {
+			return fmt.Errorf("mock error")
+		}
+		stable.Status.Conditions = condition.AddOrUpdateStatusConditionsWithLastUpdatedTimestamp(stable.Status.Conditions, clusterOfflineCondition())
```
filariow, line 138: it seems to me we can simplify this as follows:

```go
		stable, _ := newToolchainCluster("stable", tcNs, "http://cluster.com", withStatus(healthy(), clusterOfflineCondition())

		cl := test.NewFakeClient(t, stable)
		cl.MockStatusUpdate = func(ctx context.Context, obj runtimeclient.Object, opts ...runtimeclient.UpdateOption) error {
			return fmt.Errorf("mock error")
		}
```

controllers/toolchaincluster/toolchaincluster_controller_test.go
```diff
@@ -107,22 +108,97 @@ func TestClusterControllerChecks(t *testing.T) {
...
+		// when
+		recresult, err := controller.Reconcile(context.TODO(), req)
+
+		// then
+		require.EqualError(t, err, fmt.Sprintf("Failed to update the status of cluster %s: %v", stable.Name, serr))
+		require.Equal(t, reconcile.Result{}, recresult)
+		assertClusterStatus(t, cl, "stable", healthy())
+	})
+
+	t.Run("Last transition time is not updated for same condition", func(t *testing.T) {
```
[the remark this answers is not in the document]
fbm3307: but i think i removed it now, as it doesnt help in any check

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -72,23 +76,106 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+		conditions = append(conditions, clusterOfflineCondition())
+	} else {
+		if !healthcheck {
+			conditions = append(conditions, clusterNotReadyCondition(), clusterNotOfflineCondition())
+		} else {
+			conditions = append(conditions, clusterReadyCondition())
+		}
+	}
+	return conditions
+}
```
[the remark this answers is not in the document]
fbm3307: done

controllers/toolchaincluster/toolchaincluster_controller_test.go
```diff
@@ -209,3 +329,57 @@ func prepareReconcile(toolchainCluster *toolchainv1alpha1.ToolchainCluster, cl *
...
+		Message: "cluster is not reachable",
+	}
+}
+func notOffline() toolchainv1alpha1.Condition {
+	return toolchainv1alpha1.Condition{Type: toolchainv1alpha1.ToolchainClusterOffline,
+		Status:  corev1.ConditionFalse,
+		Reason:  "ClusterReachable",
+		Message: "cluster is reachable",
+	}
+}
```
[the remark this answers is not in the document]
fbm3307: We do want to keep the testing func different, also the main difference is in spec, the test ones, do not have 'Lastupdatedtime'. and 'LastTransitionTime', as it seems that comparing them is difficult, it does change in miliseconds part Date(2024, time.July, 3, 15, 28, 1, **301334000**, time.Local) the bold part is something which keeps on changing while doing a unit test , that being said, i have removed the test case of last transition time(if we are not keeping 'last transition time' in spec , there is no point in testing it

controllers/toolchaincluster/toolchaincluster_controller_test.go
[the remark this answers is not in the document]
mfrancisc: right, I don't remember what was the behavior we want to guard with this test ? Is it really needed ?

mfrancisc (commented):
Looks great! 👍  

Thanks for addressing my comments, I have only one observation regarding the healthcheck test.

controllers/toolchaincluster/toolchaincluster_controller_test.go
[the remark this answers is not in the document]
fbm3307: removed it

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -72,23 +76,99 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+func (r *Reconciler) getClusterHealthCondition(ctx context.Context, remoteClusterClientset *kubeclientset.Clientset) []v1alpha1.Condition {
+
+	healthcheck, errhealth := r.getClusterHealth(ctx, remoteClusterClientset)
+	if errhealth != nil {
+		return []v1alpha1.Condition{clusterOfflineCondition()}
+	}
+	if !healthcheck {
+		return []v1alpha1.Condition{clusterNotReadyCondition(), clusterNotOfflineCondition()}
+	}
+	return []v1alpha1.Condition{clusterReadyCondition()}
```
MatousJobanek, line 113: if the cluster gets ever into the offline state, then since you don't reset it at the end of the function, it will stay there forever. 
TBH, we could simply drop the offline condition and use the "Ready" condition only (as we do in all other controllers). In other words replace the offline condition with another Reason of the "ready" condition

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -72,23 +76,99 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+
+func clusterReadyCondition() toolchainv1alpha1.Condition {
+	currentTime := metav1.Now()
+	return toolchainv1alpha1.Condition{
+		Type:               toolchainv1alpha1.ConditionReady,
+		Status:             corev1.ConditionTrue,
+		Reason:             toolchainv1alpha1.ToolchainClusterClusterReadyReason,
+		Message:            healthzOk,
+		LastUpdatedTime:    &currentTime,
+		LastTransitionTime: currentTime,
```
MatousJobanek, line 132: if you use the generic functions from toolchain-common, then you don't need to define the times - they should be set by the functions automatically.

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -72,23 +76,99 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+	}
+}
+
+func clusterOfflineCondition() toolchainv1alpha1.Condition {
+	currentTime := metav1.Now()
+	return toolchainv1alpha1.Condition{
+		Type:               toolchainv1alpha1.ToolchainClusterOffline,
+		Status:             corev1.ConditionTrue,
```
MatousJobanek, line 152: As I mentioned above, it would be nice to drop this condition to make it more consistent with the other controllers/CRDs
```suggestion
		Type:               toolchainv1alpha1.ConditionReady,
		Status:             corev1.ConditionFalse,
```
it would be also much easier to manage the changes of the states.

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -72,23 +76,99 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
 		return reconcile.Result{}, err
 	}
-
 	return reconcile.Result{RequeueAfter: r.RequeAfter}, nil
 }
 
+func (r *Reconciler) updateStatus(ctx context.Context, toolchainCluster *toolchainv1alpha1.ToolchainCluster, currentconditions ...toolchainv1alpha1.Condition) error {
+	toolchainCluster.Status.Conditions = condition.AddOrUpdateStatusConditionsWithLastUpdatedTimestamp(toolchainCluster.Status.Conditions, currentconditions...)
+	if err := r.Client.Status().Update(ctx, toolchainCluster); err != nil {
+		return errors.Wrapf(err, "Failed to update the status of cluster %s", toolchainCluster.Name)
```
MatousJobanek, line 99: Let's use the recommended `fmt.Errorf` instead. `errors.Wrapf` is an obsolete legacy code we should start dropping.

controllers/toolchaincluster/toolchaincluster_controller.go
fbm3307: it is same as condition Not ready, so you mean to drop clusteroffline condition itself..?

fbm3307: updated it

controllers/toolchaincluster/toolchaincluster_controller.go
fbm3307: done

controllers/toolchaincluster/toolchaincluster_controller.go
fbm3307: okay updated it

controllers/toolchaincluster/healthchecker_test.go
```diff
@@ -39,139 +33,52 @@ func TestClusterHealthChecks(t *testing.T) {
...
-				assert.Equal(t, expCond.Reason, cond.Reason)
-				assert.Equal(t, expCond.Message, cond.Message)
-				continue ExpConditions
+			//then
+			require.Equal(t, tc.healthcheck, healthcheck)
+			if tc.tctype == "Notfound" {
+				require.Error(t, errh)
+			} else {
+				require.Equal(t, tc.errh, errh)
 			}
```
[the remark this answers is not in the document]
fbm3307: updated it

controllers/toolchaincluster/toolchaincluster_controller_test.go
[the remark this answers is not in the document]
fbm3307: i gues now i have removed it as per latest changes

controllers/toolchaincluster/healthchecker_test.go
```diff
@@ -39,139 +33,50 @@ func TestClusterHealthChecks(t *testing.T) {
...
-				assert.Equal(t, expCond.Reason, cond.Reason)
-				assert.Equal(t, expCond.Message, cond.Message)
-				continue ExpConditions
+			//then
+			require.Equal(t, tc.healthcheck, healthcheck)
+			if tc.tctype == "Notfound" {
+				require.Error(t, errh)
+			} else {
+				require.NoError(t, errh)
 			}
```
MatousJobanek, line 78: you could use the `tc.errh` (because you don't use it at all), set it as:
```
		"ErrorWhileDoingHealth": {
			tctype:      "Notfound",
			apiendpoint: "http://not-found.com",
			healthcheck: false,
			errh:        fmt.Errorf("the server could not find the requested resource"),
		},
```

and then run here

```suggestion
			if tc.errh != nil {
				require.EqualError(t, errh, tc.errh.Error())
			} else {
				require.NoError(t, errh)
			}
```

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -22,9 +24,10 @@ import (
 
 // Reconciler reconciles a ToolchainCluster object
 type Reconciler struct {
-	Client     client.Client
-	Scheme     *runtime.Scheme
-	RequeAfter time.Duration
+	Client      client.Client
+	Scheme      *runtime.Scheme
+	RequeAfter  time.Duration
+	CheckHealth func(context.Context, *kubeclientset.Clientset) (bool, error)
```
MatousJobanek, line 30: the `CheckHealth` field doesn't have to be exported - the unit test is in the same package

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -72,23 +74,66 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+		return fmt.Errorf("failed to update the status of cluster - %s "+err.Error(), toolchainCluster.Name)
+	}
+	return nil
+}
+
+func (r *Reconciler) getClusterHealthCondition(ctx context.Context, remoteClusterClientset *kubeclientset.Clientset) []v1alpha1.Condition {
+
+	healthcheck, errhealth := r.getClusterHealth(ctx, remoteClusterClientset)
+	if errhealth != nil || !healthcheck {
+		return []v1alpha1.Condition{clusterNotReadyCondition()}
```
MatousJobanek, line 106: As I mentioned in the previous review and also during the sync call, let's use a proper `Reason` based on the result of the health check.
* if error is not nil, then `ClusterNotReachable` and also pass the error as the message to the condition
* if health check is false, then `ToolchainClusterClusterNotReadyReason`

We wanted to reuse the condition only, not the reason.

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -72,23 +74,66 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
 
+func (r *Reconciler) updateStatus(ctx context.Context, toolchainCluster *toolchainv1alpha1.ToolchainCluster, currentconditions ...toolchainv1alpha1.Condition) error {
+	toolchainCluster.Status.Conditions = condition.AddOrUpdateStatusConditionsWithLastUpdatedTimestamp(toolchainCluster.Status.Conditions, currentconditions...)
+	if err := r.Client.Status().Update(ctx, toolchainCluster); err != nil {
+		return fmt.Errorf("failed to update the status of cluster - %s "+err.Error(), toolchainCluster.Name)
+	}
+	return nil
+}
+
+func (r *Reconciler) getClusterHealthCondition(ctx context.Context, remoteClusterClientset *kubeclientset.Clientset) []v1alpha1.Condition {
```
MatousJobanek, line 102: there is no reason for returning the slice of conditions, right?

controllers/toolchaincluster/toolchaincluster_controller_test.go
```diff
@@ -103,26 +104,72 @@ func TestClusterControllerChecks(t *testing.T) {
...
+
+		cl := test.NewFakeClient(t, stable)
+		cl.MockStatusUpdate = func(ctx context.Context, obj runtimeclient.Object, opts ...runtimeclient.UpdateOption) error {
+			return fmt.Errorf("mock error")
+		}
 		controller, req := prepareReconcile(stable, cl, requeAfter)
 
 		// when
-		_, err := controller.Reconcile(context.TODO(), req)
+		recresult, err := controller.Reconcile(context.TODO(), req)
```
MatousJobanek, line 140: let's use a camelCase for variable names
```suggestion
		recResult, err := controller.Reconcile(context.TODO(), req)
```

controllers/toolchaincluster/toolchaincluster_controller_test.go
```diff
@@ -209,3 +300,27 @@ func prepareReconcile(toolchainCluster *toolchainv1alpha1.ToolchainCluster, cl *
...
+		for _, cond := range tc.Status.Conditions {
+			if expCond.Type == cond.Type {
+				assert.Equal(t, expCond.Status, cond.Status)
+				assert.Equal(t, expCond.Reason, cond.Reason)
+				assert.Equal(t, expCond.Message, cond.Message)
+				continue ExpConditions
+			}
+		}
+		assert.Failf(t, "condition not found", "the list of conditions %v doesn't contain the expected condition %v", tc.Status.Conditions, expCond)
+	}
```
MatousJobanek, line 325: I believe that you can simplify this part of the code since we use only one condition type

controllers/toolchaincluster/toolchaincluster_controller_test.go
```diff
@@ -103,26 +104,72 @@ func TestClusterControllerChecks(t *testing.T) {
...
+		require.Equal(t, reconcile.Result{}, recresult)
+
 		actualtoolchaincluster := &toolchainv1alpha1.ToolchainCluster{}
 		err = cl.Client.Get(context.TODO(), types.NamespacedName{Name: "stable", Namespace: tcNs}, actualtoolchaincluster)
 		require.NoError(t, err)
-		assertClusterStatus(t, cl, "stable", offline())
+		assertClusterStatus(t, cl, "stable", clusterReadyCondition())
+	})
+
+	t.Run("error while updating a toolchain cluster status", func(t *testing.T) {
```
MatousJobanek, line 152: looks like a duplicate of the previous case
```suggestion
	t.Run("error while updating a toolchain cluster status when health-check failed", func(t *testing.T) {
```

controllers/toolchaincluster/toolchaincluster_controller_test.go
```diff
@@ -160,6 +207,50 @@ func TestClusterControllerChecks(t *testing.T) {
 	})
 }
 
+func TestGetClusterHealth(t *testing.T) {
```
MatousJobanek, line 210: aren't these test cases already covered in the previous tests?

fbm3307: These check the checkHealth param usage..(if someday we set it through main(while calling the controller) so  i added them to test that scenario and also they do the integration testing of checking the health check function and toolchain controller  together.

controllers/toolchaincluster/toolchaincluster_controller.go
fbm3307: updated it now

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -72,23 +74,77 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+	return nil
+}
+
+func (r *Reconciler) getClusterHealthCondition(ctx context.Context, remoteClusterClientset *kubeclientset.Clientset) v1alpha1.Condition {
+
+	ishealthy, err := r.getClusterHealth(ctx, remoteClusterClientset)
+	if err != nil {
+		return clusterOfflineCondition(err.Error())
+	} else {
+		if !ishealthy {
```
MatousJobanek, line 108: could you please keep using camelCase for variable names? I've seen that in other cases too.

controllers/toolchaincluster/toolchaincluster_controller_test.go
```diff
@@ -160,6 +207,50 @@ func TestClusterControllerChecks(t *testing.T) {
...
+		reset := setupCachedClusters(t, cl, stable)
+
+		defer reset()
+		controller, req := prepareReconcile(stable, cl, requeAfter)
+		controller.checkHealth = func(context.Context, *kubeclientset.Clientset) (bool, error) {
+			return true, nil
+		}
+
+		// when
+		recresult, err := controller.Reconcile(context.TODO(), req)
```
MatousJobanek, line 225: As I mentioned in the first comment and also in the previous review, please use camelCase
```suggestion
		recResult, err := controller.Reconcile(context.TODO(), req)
```

controllers/toolchaincluster/healthchecker_test.go
```diff
@@ -39,139 +34,51 @@ func TestClusterHealthChecks(t *testing.T) {
...
 
 	tests := map[string]struct {
-		tctype            string
-		apiendpoint       string
-		clusterconditions []toolchainv1alpha1.Condition
-		status            toolchainv1alpha1.ToolchainClusterStatus
+		tctype      string
+		apiendpoint string
+		healthcheck bool
+		errh        error
```
MatousJobanek, line 40: let's keep naming convention
```suggestion
		tcType      string
		apiEndpoint string
		healthCheck bool
		err         error
```

controllers/toolchaincluster/healthchecker_test.go
```diff
@@ -39,139 +34,51 @@ func TestClusterHealthChecks(t *testing.T) {
...
-			err = healthChecker.updateIndividualClusterStatus(context.TODO(), tctype)
 
-			//then
-			require.NoError(t, err)
-			assertClusterStatus(t, cl, tc.tctype, tc.clusterconditions...)
-		})
-	}
-}
+			//when
+			healthcheck, errh := getClusterHealthStatus(context.TODO(), cacheclient)
```
MatousJobanek, line 72: here as well
```suggestion
			healthCheck, err := getClusterHealthStatus(context.TODO(), cacheclient)
```

controllers/toolchaincluster/healthchecker_test.go
```diff
@@ -39,139 +34,51 @@ func TestClusterHealthChecks(t *testing.T) {
...
 		t.Run(k, func(t *testing.T) {
-			tctype, sec := newToolchainCluster(tc.tctype, tcNs, tc.apiendpoint, tc.status)
+			//given
+			tctype, sec := newToolchainCluster(tc.tctype, tcNs, tc.apiendpoint, toolchainv1alpha1.ToolchainClusterStatus{})
 			cl := test.NewFakeClient(t, tctype, sec)
 			reset := setupCachedClusters(t, cl, tctype)
 			defer reset()
 			cachedtc, found := cluster.GetCachedToolchainCluster(tctype.Name)
 			require.True(t, found)
 			cacheclient, err := kubeclientset.NewForConfig(cachedtc.RestConfig)
```
MatousJobanek, line 68:
```suggestion
			cachedTC, found := cluster.GetCachedToolchainCluster(tctype.Name)
			require.True(t, found)
			cacheClient, err := kubeclientset.NewForConfig(cachedTC.RestConfig)
```

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -58,9 +61,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	cachedCluster, ok := cluster.GetCachedToolchainCluster(toolchainCluster.Name)
 	if !ok {
 		err := fmt.Errorf("cluster %s not found in cache", toolchainCluster.Name)
-		toolchainCluster.Status.Conditions = []toolchainv1alpha1.Condition{clusterOfflineCondition()}
-		if err := r.Client.Status().Update(ctx, toolchainCluster); err != nil {
-			reqLogger.Error(err, "failed to update the status of ToolchainCluster")
+		if err := r.updateStatus(ctx, toolchainCluster, clusterNotReadyCondition()); err != nil {
```
MatousJobanek, line 64: correct me if I'm wrong, but we used cluster offline conditon which should correspond to
```suggestion
		if err := r.updateStatus(ctx, toolchainCluster, clusterOfflineCondition(err.Error())); err != nil {
```

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -72,23 +74,77 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	clientSet, err := kubeclientset.NewForConfig(cachedCluster.RestConfig)
 	if err != nil {
 		reqLogger.Error(err, "cannot create ClientSet for the ToolchainCluster")
+		if err := r.updateStatus(ctx, toolchainCluster, clusterNotReadyCondition()); err != nil {
```
MatousJobanek, line 77: As Francisc already pointed out in his comment in the previous review, this should be offline as well:
```suggestion
		if err := r.updateStatus(ctx, toolchainCluster, clusterOfflineCondition(err.Error())); err != nil {
```

controllers/toolchaincluster/toolchaincluster_controller_test.go
```diff
@@ -103,26 +104,72 @@ func TestClusterControllerChecks(t *testing.T) {
...
+		cl := test.NewFakeClient(t, unstable)
+		unstable.Status.Conditions = condition.AddOrUpdateStatusConditionsWithLastUpdatedTimestamp(unstable.Status.Conditions, clusterNotReadyCondition())
+		controller, req := prepareReconcile(unstable, cl, requeAfter)
 
+		// when
+		_, err := controller.Reconcile(context.TODO(), req)
+
+		// then
+		require.EqualError(t, err, "cluster unstable not found in cache")
+		actualtoolchaincluster := &toolchainv1alpha1.ToolchainCluster{}
```
MatousJobanek, line 123:
```suggestion
		actualToolchainCluster := &toolchainv1alpha1.ToolchainCluster{}
```

controllers/toolchaincluster/toolchaincluster_controller_test.go
```diff
@@ -209,3 +300,25 @@ func prepareReconcile(toolchainCluster *toolchainv1alpha1.ToolchainCluster, cl *
...
+ExpConditions:
+	for _, expCond := range clusterConds {
+		for _, cond := range tc.Status.Conditions {
+			assert.Equal(t, expCond.Status, cond.Status)
+			assert.Equal(t, expCond.Reason, cond.Reason)
+			assert.Equal(t, expCond.Message, cond.Message)
+			continue ExpConditions
+		}
+		assert.Failf(t, "condition not found", "the list of conditions %v doesn't contain the expected condition %v", tc.Status.Conditions, expCond)
+	}
```
MatousJobanek, line 323: As I have already mentioned during our sync call, you can use assertion function for that:
```suggestion
	test.AssertConditionsMatch(t, tc.Status.Conditions, clusterConds...)
```

MatousJobanek (commented):
Please, double check all variables in these 4 files and make sure that they are consistent with the rest of our code:
* use camelCase
* `err` for variable names of errors when possible

controllers/toolchaincluster/toolchaincluster_controller.go
fbm3307: yes thanks for this

controllers/toolchaincluster/toolchaincluster_controller.go
fbm3307: i did understood that previous comment differently , but yea it has to be offline

fbm3307:
> Please, double check all variables in these 4 files and make sure that they are consistent with the rest of our code:
> 
> * use camelCase
> * `err` for variable names of errors when possible

I have tried cover everything & updated it to camel case @MatousJobanek the last commit.

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -72,23 +74,77 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
-	if err := healthChecker.updateIndividualClusterStatus(ctx, toolchainCluster); err != nil {
+	if err := r.updateStatus(ctx, toolchainCluster, healthCheckResult); err != nil {
 		reqLogger.Error(err, "unable to update cluster status of ToolchainCluster")
 		return reconcile.Result{}, err
 	}
-
 	return reconcile.Result{RequeueAfter: r.RequeAfter}, nil
 }
 
+func (r *Reconciler) updateStatus(ctx context.Context, toolchainCluster *toolchainv1alpha1.ToolchainCluster, currentconditions ...toolchainv1alpha1.Condition) error {
```
MatousJobanek, line 94:
```suggestion
func (r *Reconciler) updateStatus(ctx context.Context, toolchainCluster *toolchainv1alpha1.ToolchainCluster, currentConditions ...toolchainv1alpha1.Condition) error {
```

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -72,23 +74,77 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+	return nil
+}
+
+func (r *Reconciler) getClusterHealthCondition(ctx context.Context, remoteClusterClientset *kubeclientset.Clientset) v1alpha1.Condition {
+
+	ishealthy, err := r.getClusterHealth(ctx, remoteClusterClientset)
+	if err != nil {
+		return clusterOfflineCondition(err.Error())
+	} else {
+		if !ishealthy {
```
MatousJobanek, line 108: already mentioned that 2 times before
https://github.com/codeready-toolchain/toolchain-common/pull/401#discussion_r1665463555
https://github.com/codeready-toolchain/toolchain-common/pull/401#discussion_r1674127887
```suggestion
	isHealthy, err := r.getClusterHealth(ctx, remoteClusterClientset)
	if err != nil {
		return clusterOfflineCondition(err.Error())
	} else {
		if !isHealthy {
```

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -72,23 +74,77 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+			return clusterNotReadyCondition()
+		}
+		return clusterReadyCondition()
+	}
```
MatousJobanek, line 112: btw, it's just a detail, but you don't need else statement in this situation:
```suggestion
	isHealthy, err := r.getClusterHealth(ctx, remoteClusterClientset)
	if err != nil {
		return clusterOfflineCondition(err.Error())
	}
	if !isHealthy {
		return clusterNotReadyCondition()
	}
	return clusterReadyCondition()
```

controllers/toolchaincluster/toolchaincluster_controller_test.go
```diff
@@ -58,71 +59,115 @@ func TestClusterControllerChecks(t *testing.T) {
...
 		// then
 		require.EqualError(t, err, "mock error")
-		require.Equal(t, reconcile.Result{Requeue: false, RequeueAfter: 0}, recresult)
+		require.Equal(t, reconcile.Result{Requeue: false, RequeueAfter: 0}, recResult)
 	})
 
 	t.Run("reconcile successful and requeued", func(t *testing.T) {
 		// given
-		stable, sec := newToolchainCluster("stable", tcNs, "http://cluster.com", toolchainv1alpha1.ToolchainClusterStatus{})
+		stable, sec := newToolchainCluster("stable", tcNs, "http://cluster.com", withStatus(clusterReadyCondition()))
```
MatousJobanek, line 92: the test is supposed to verify that the ready condition is properly set - keep the original format
```suggestion
		stable, sec := newToolchainCluster("stable", tcNs, "http://cluster.com", toolchainv1alpha1.ToolchainClusterStatus{}))
```

controllers/toolchaincluster/toolchaincluster_controller_test.go
```diff
@@ -58,71 +59,115 @@ func TestClusterControllerChecks(t *testing.T) {
...
-		require.Equal(t, reconcile.Result{RequeueAfter: requeAfter}, recresult)
-		assertClusterStatus(t, cl, "stable", healthy())
+		require.Equal(t, reconcile.Result{RequeueAfter: requeAfter}, recResult)
+		assertClusterStatus(t, cl, "stable", clusterReadyCondition())
 	})
 
 	t.Run("toolchain cluster cache not found", func(t *testing.T) {
 		// given
-		stable, _ := newToolchainCluster("stable", tcNs, "http://cluster.com", toolchainv1alpha1.ToolchainClusterStatus{})
+		unstable, _ := newToolchainCluster("unstable", tcNs, "http://unstable.com", withStatus(clusterOfflineCondition("mock error")))
```
MatousJobanek, line 111: it's the same - don't set the condition - it should be set by the controller
```suggestion
		unstable, _ := newToolchainCluster("unstable", tcNs, "http://unstable.com", toolchainv1alpha1.ToolchainClusterStatus{})
```

controllers/toolchaincluster/toolchaincluster_controller_test.go
```diff
@@ -58,71 +59,115 @@ func TestClusterControllerChecks(t *testing.T) {
...
+		err = cl.Client.Get(context.TODO(), types.NamespacedName{Name: "stable", Namespace: tcNs}, actualToolchainCluster)
 		require.NoError(t, err)
-		assertClusterStatus(t, cl, "stable", offline())
+		assertClusterStatus(t, cl, "stable", clusterReadyCondition())
+	})
+
+	t.Run("error while updating a toolchain cluster status when health-check failed", func(t *testing.T) {
+		// given
+		stable, sec := newToolchainCluster("stable", tcNs, "http://cluster.com", toolchainv1alpha1.ToolchainClusterStatus{})
+		serr := fmt.Errorf("my test error")
```
MatousJobanek, line 153:
```suggestion
		expectedErr := fmt.Errorf("my test error")
```

controllers/toolchaincluster/toolchaincluster_controller_test.go
```diff
@@ -160,6 +205,50 @@ func TestClusterControllerChecks(t *testing.T) {
 	})
 }
 
+func TestGetClusterHealth(t *testing.T) {
+	t.Run("Check health default", func(t *testing.T) {
+		// given
+		stable, sec := newToolchainCluster("stable", "test-namespace", "http://cluster.com", withStatus(clusterReadyCondition()))
```
MatousJobanek, line 211:
```suggestion
		stable, sec := newToolchainCluster("stable", "test-namespace", "http://cluster.com", toolchainv1alpha1.ToolchainClusterStatus{})
```

controllers/toolchaincluster/toolchaincluster_controller.go
fbm3307: its a miss from my side will take extra care next time

controllers/toolchaincluster/toolchaincluster_controller.go
```diff
@@ -72,23 +74,77 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
 		return reconcile.Result{}, err
 	}
-
 	return reconcile.Result{RequeueAfter: r.RequeAfter}, nil
 }
 
+func (r *Reconciler) updateStatus(ctx context.Context, toolchainCluster *toolchainv1alpha1.ToolchainCluster, currentConditions ...toolchainv1alpha1.Condition) error {
+	toolchainCluster.Status.Conditions = condition.AddOrUpdateStatusConditionsWithLastUpdatedTimestamp(toolchainCluster.Status.Conditions, currentConditions...)
+	if err := r.Client.Status().Update(ctx, toolchainCluster); err != nil {
+		return fmt.Errorf("failed to update the status of cluster - %s "+err.Error(), toolchainCluster.Name)
```
MatousJobanek, line 97: let's use proper formatting
```suggestion
		return fmt.Errorf("failed to update the status of cluster - %s: %w", toolchainCluster.Name, err)
```

controllers/toolchaincluster/toolchaincluster_controller_test.go
```diff
@@ -94,35 +93,80 @@ func TestClusterControllerChecks(t *testing.T) {
...
+
+		// when
+		_, err := controller.Reconcile(context.TODO(), req)
+
+		// then
+		require.EqualError(t, err, "cluster unstable not found in cache")
+		actualToolchainCluster := &toolchainv1alpha1.ToolchainCluster{}
+		err = cl.Client.Get(context.TODO(), types.NamespacedName{Name: "unstable", Namespace: tcNs}, actualToolchainCluster)
+		require.NoError(t, err)
+		assertClusterStatus(t, cl, "unstable", clusterOfflineCondition("cluster unstable not found in cache"))
```
MatousJobanek, line 124: there is no need to call explicitly, when it's called inside of the `assertClusterStatus` function:
```suggestion
		assertClusterStatus(t, cl, "unstable", clusterOfflineCondition("cluster unstable not found in cache"))
```

controllers/toolchaincluster/toolchaincluster_controller_test.go
```diff
@@ -94,35 +93,80 @@ func TestClusterControllerChecks(t *testing.T) {
...
 
 		// then
 		require.EqualError(t, err, "cluster stable not found in cache")
-		actualtoolchaincluster := &toolchainv1alpha1.ToolchainCluster{}
-		err = cl.Client.Get(context.TODO(), types.NamespacedName{Name: "stable", Namespace: tcNs}, actualtoolchaincluster)
+		require.Equal(t, reconcile.Result{}, recResult)
+
+		actualToolchainCluster := &toolchainv1alpha1.ToolchainCluster{}
+		err = cl.Client.Get(context.TODO(), types.NamespacedName{Name: "stable", Namespace: tcNs}, actualToolchainCluster)
 		require.NoError(t, err)
```
MatousJobanek, line 146: you call get on the Toolchaincluster CR, but you don't do any verification with it. 
Anyway, you can use this: 
```suggestion
		assertClusterStatus(t, cl, "stable")
```

controllers/toolchaincluster/toolchaincluster_controller_test.go
```diff
@@ -94,35 +93,80 @@ func TestClusterControllerChecks(t *testing.T) {
...
+		controller, req := prepareReconcile(stable, cl, requeAfter)
+		controller.checkHealth = func(context.Context, *kubeclientset.Clientset) (bool, error) {
+			return false, expectedErr
+		}
+		// when
+		recResult, err := controller.Reconcile(context.TODO(), req)
+
+		// then
+		require.EqualError(t, err, fmt.Sprintf("failed to update the status of cluster - %s %v", stable.Name, expectedErr))
+		require.Equal(t, reconcile.Result{}, recResult)
```
MatousJobanek, line 169:
```suggestion
		require.Equal(t, reconcile.Result{}, recResult)
		assertClusterStatus(t, cl, "stable")
```

MatousJobanek (approved):
Looks good, but please take a look at my last comments

PR #403, opened 2024-06-12
fix: don't use CAData in the ToolchainCluster client

MatousJobanek:
[KUBESAW-121](https://issues.redhat.com/browse/KUBESAW-121)

don't read the caBundle and don't set the CAData in the ToolchainCluster client. Instead of it, use the `ToolchainCluster.spec.disabledTLSValidations` to decide if it should use insecure connection or not
see https://github.com/codeready-toolchain/toolchain-cicd/pull/110


xcoulon (approved):

alexeykazakov (approved):
Do we also want to drop the CABandle from the ToolchainCluster config after merging it since it's not used anymore or it will be part of the overall ToolchainCluster thing refactoring?

fbm3307 (approved):

MatousJobanek:
@alexeykazakov yes, there is a separate story for this https://issues.redhat.com/browse/KUBESAW-47

PR #404, opened 2024-06-14
Store kubeconfig in ToolchainCluster secret

metlos:
This introduces a temporary migration step that converts the connection details spread across the ToolchainCluster and its associated secret into the "kubeconfig" field in the secret data that contains the serialized kubeconfig file with the details all included.

Associated PRs:
- toolchain-e2e: https://github.com/codeready-toolchain/toolchain-e2e/pull/997
- host-operator: https://github.com/codeready-toolchain/host-operator/pull/1042

### DIFF
--- a/controllers/toolchaincluster/toolchaincluster_controller_test.go
+++ b/controllers/toolchaincluster/toolchaincluster_controller_test.go
@@ -9,12 +9,17 @@ import (
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
 	"github.com/codeready-toolchain/toolchain-common/pkg/cluster"
 	"github.com/codeready-toolchain/toolchain-common/pkg/test"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"gopkg.in/h2non/gock.v1"
+	"gotest.tools/assert/cmp"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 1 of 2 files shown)


controllers/toolchaincluster/toolchaincluster_controller_test.go
```diff
@@ -119,6 +124,53 @@ func TestClusterControllerChecks(t *testing.T) {
...
+
+	t.Run("migrates connection settings to kubeconfig in secret", func(t *testing.T) {
+		// given
+		tc, secret := newToolchainCluster("tc", tcNs, "http://cluster.com", toolchainv1alpha1.ToolchainClusterStatus{})
+		cl := test.NewFakeClient(t, tc, secret)
+		reset := setupCachedClusters(t, cl, tc)
+		defer reset()
+
+		controller, req := prepareReconcile(tc, cl, requeAfter)
+		expectedKubeConfig := composeKubeConfigFromData([]byte("mycooltoken"), "http://cluster.com", "test-namespace", true)
```
mfrancisc, line 136: should we use a fake/hardcoded kubeconfig ? If there's a bug in the `composeKubeConfigFromData` function this test will not catch it right , since both the reconciler and test are using the same logic for generating the kubeconfig.

controllers/toolchaincluster/toolchaincluster_controller_test.go
```diff
@@ -119,6 +124,53 @@ func TestClusterControllerChecks(t *testing.T) {
...
+		assert.Contains(t, secretAfterReconcile.Data, "kubeconfig")
+
+		// we need to use this more complex equals, because we don't initialize the Extension fields (i.e. they're nil)
+		// while they're initialized and empty after deserialization, which causes the "normal" deep equals to fail.
+		result := cmp.DeepEqual(expectedKubeConfig, *actualKubeConfig,
+			cmpopts.IgnoreFields(clientcmdapi.Config{}, "Extensions"),
+			cmpopts.IgnoreFields(clientcmdapi.Preferences{}, "Extensions"),
+			cmpopts.IgnoreFields(clientcmdapi.Cluster{}, "Extensions"),
+			cmpopts.IgnoreFields(clientcmdapi.AuthInfo{}, "Extensions"),
+			cmpopts.IgnoreFields(clientcmdapi.Context{}, "Extensions"),
```
mfrancisc, line 156: by using a fake/static kubeconfig we could also remove those `IgnoreFields` , and make the kubeconfig exactly as the real one. Just an idea.

controllers/toolchaincluster/toolchaincluster_controller_test.go
mfrancisc: maybe the `TestComposeKubeConfig` would catch it.

controllers/toolchaincluster/toolchaincluster_controller_test.go
```diff
@@ -119,6 +124,53 @@ func TestClusterControllerChecks(t *testing.T) {
...
+		)()
+
+		assert.True(t, result.Success())
+	})
+}
+
+func TestComposeKubeConfig(t *testing.T) {
```
mfrancisc, line 163: should we also check for the keys in kubeconfig: 
- `ctx` ( which is the default context I see )
- `cluster` ( which is the default cluster )
- `auth` ( which holds the authentication info ) 

or those are not going to be referenced anywhere ?

mfrancisc (commented):
Great Job 👍 

I have few comments related to the tests.

controllers/toolchaincluster/toolchaincluster_controller_test.go
metlos: These are all tested by the means of the `CurrentContext`. This is set to "ctx" and the context then references the cluster and auth by name and as such binds everything together.

controllers/toolchaincluster/toolchaincluster_controller_test.go
metlos: The `composeKubeConfigFromData` function is really a glorified struct initialization, so I chose to use that. Having a hardcoded config with all the `Extension` fields would IMHO require the same amount of explanation as the customized DeepEqual. So I chose this way that I think is ever so slightly more legible.

And you're right, the `TestComposeKubeConfig` function is there for checking the validity of the generated kubeconfig, so I think it is OK to have 1 function used both in test and the tested reconcile (if it becomes more complicated, I would agree with you that it could cause a problem).

controllers/toolchaincluster/toolchaincluster_controller_test.go
metlos: See my reply in your other comment above.

controllers/toolchaincluster/toolchaincluster_controller_test.go
mfrancisc: yup not a big problem! 

I was thinking something similar to https://github.com/codeready-toolchain/host-operator/blob/43bf61e3a43ca2444883348059e0fe3d3177dff3/controllers/spacerequest/spacerequest_controller_test.go#L1094 for generating a fake kubeconfig based on what you expect the actual one should look like. I'm not sure I understand which `Extensions` fields you are referring too, but I'm sure I'm missing something here.

PR #406, opened 2024-06-21
Add feature toggles to config test util

alexeykazakov:
Needed for https://issues.redhat.com/browse/SANDBOX-653

### DIFF
--- a/pkg/test/config/toolchainconfig.go
+++ b/pkg/test/config/toolchainconfig.go
@@ -579,6 +579,19 @@ func (o TiersOption) DurationBeforeChangeTierRequestDeletion(value string) Tiers
 	return o
 }
 
+func (o TiersOption) FeatureToggle(name string, weight *uint) TiersOption {
+	o.addFunction(func(config *toolchainv1alpha1.ToolchainConfig) {
+		if config.Spec.Host.Tiers.FeatureToggles == nil {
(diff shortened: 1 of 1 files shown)


ranakan19 (approved):

pkg/test/config/toolchainconfig.go
```diff
@@ -579,6 +579,19 @@ func (o TiersOption) DurationBeforeChangeTierRequestDeletion(value string) Tiers
 	return o
 }
 
+func (o TiersOption) FeatureToggle(name string, weight *uint) TiersOption {
+	o.addFunction(func(config *toolchainv1alpha1.ToolchainConfig) {
+		if config.Spec.Host.Tiers.FeatureToggles == nil {
+			config.Spec.Host.Tiers.FeatureToggles = make([]toolchainv1alpha1.FeatureToggle, 0)
```
rajivnathan, line 585: Shouldn't the default be 100?

rajivnathan: oops, was too quick reading it, you're just initializing the array 🙂

rajivnathan (approved):

pkg/test/config/toolchainconfig.go
alexeykazakov: yeah.. I don't really think we need to initialize the array at all. You can append to a nil slice but just to be 100% safe... :)

PR #407, opened 2024-06-21
add public-viewer configuration

filariow:
This PR adds configuration for Public-Viewer as described in [JIRA [ASC-532](https://issues.redhat.com//browse/ASC-532)](https://issues.redhat.com//browse/ASC-532).

### DIFF
--- a/pkg/configuration/publicviewer_config.go
+++ b/pkg/configuration/publicviewer_config.go
@@ -1,0 +1,13 @@
+package configuration
+
+import (
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
+)
+
+type PublicViewerConfig struct {
+	Config *toolchainv1alpha1.PublicViewerConfiguration
+}
+
+func (c PublicViewerConfig) Enabled() bool {
+	return c.Config != nil && c.Config.Enabled
+}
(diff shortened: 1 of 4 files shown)


pkg/configuration/publicviewer_config.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
+)
+
+type PublicViewerConfig struct {
+	Config *toolchainv1alpha1.PublicViewerConfiguration
+}
+
+func (c PublicViewerConfig) Enabled() bool {
+	return c.Config != nil && c.Config.Enabled
```
mfrancisc, line 12: alternatively you could use:
```suggestion
	return GetBool(c.Config.Enabled, false)
```

pkg/test/config/toolchainconfig.go
```diff
@@ -49,6 +49,22 @@ func PerMemberCluster(name string, value int) PerMemberClusterOptionInt {
 	}
 }
 
+//---PublicViewer Configuration---//
```
mfrancisc, line 52: very minor thing - now it's part of the Host configuration right ?

mfrancisc (approved):
Looks good! 

Just few cosmetics.

pkg/test/config/toolchainconfig.go
filariow: great catch! let me fix this

filariow:
No more required after moving this configuration under Host (#430)

PR #412, opened 2024-07-12
Initialize the cluster client cache from the kubeconfig in secrets

metlos:
This changes the cluster cache to initialize the connections to the clusters from the kubeconfig stored in the toolchain cluster secret.

The information from the cluster cache is also used to populate the new informational fields in the ToolchainCluster status.

Related PRs:
- host-operator: https://github.com/codeready-toolchain/host-operator/pull/1058
- toolchain-e2e: https://github.com/codeready-toolchain/toolchain-e2e/pull/1012

### DIFF
--- a/pkg/cluster/service.go
+++ b/pkg/cluster/service.go
@@ -42,7 +42,7 @@ type ToolchainClusterService struct {
 
 type NewClient func(config *rest.Config, options client.Options) (client.Client, error)
 
-// NewToolchainClusterServiceWithClient creates a new instance of ToolchainClusterService object and assigns the given newClient functione to be used for creating a client
+// NewToolchainClusterServiceWithClient creates a new instance of ToolchainClusterService object and assigns the given newClient function to be used for creating a client
 func NewToolchainClusterServiceWithClient(client client.Client, log logr.Logger, namespace string, timeout time.Duration, newClient NewClient) ToolchainClusterService {
 	service := NewToolchainClusterService(client, log, namespace, timeout)
 	service.newClient = newClient
--- a/pkg/cluster/service.go
+++ b/pkg/cluster/service.go
@@ -160,16 +160,9 @@ func (s *ToolchainClusterService) enrichLogger(cluster *toolchainv1alpha1.Toolch
 
 // NewClusterConfig generate a new cluster config by fetching the necessary info the given ToolchainCluster's associated Secret and taking all data from ToolchainCluster CR
 func NewClusterConfig(cl client.Client, toolchainCluster *toolchainv1alpha1.ToolchainCluster, timeout time.Duration) (*Config, error) {
-	clusterName := toolchainCluster.Name
-
-	apiEndpoint := toolchainCluster.Spec.APIEndpoint
-	if apiEndpoint == "" {
-		return nil, errors.Errorf("the api endpoint of cluster %s is empty", clusterName)
-	}
-
 	secretName := toolchainCluster.Spec.SecretRef.Name
 	if secretName == "" {
-		return nil, errors.Errorf("cluster %s does not have a secret name", clusterName)
+		return nil, errors.Errorf("cluster %s does not have a secret name", toolchainCluster.Name)
 	}
 	secret := &v1.Secret{}
 	name := types.NamespacedName{
@@ -178,7 +171,22 @@ func NewClusterConfig(cl client.Client, toolchainCluster *toolchainv1alpha1.Tool
 	}
 	err := cl.Get(context.TODO(), name, secret)
 	if err != nil {
-		return nil, errors.Wrapf(err, "unable to get secret %s for cluster %s", name, clusterName)
+		return nil, errors.Wrapf(err, "unable to get secret %s for cluster %s", name, toolchainCluster.Name)
+	}
+
+	if _, ok := secret.Data["kubeconfig"]; ok {
+		return loadConfigFromKubeConfig(toolchainCluster, secret)
+	} else {
+		return loadConfigFromLegacyToolchainCluster(toolchainCluster, secret, timeout)
+	}
+}
+
+func loadConfigFromLegacyToolchainCluster(toolchainCluster *toolchainv1alpha1.ToolchainCluster, secret *v1.Secret, timeout time.Duration) (*Config, error) {
+	clusterName := toolchainCluster.Name
+
+	apiEndpoint := toolchainCluster.Spec.APIEndpoint
+	if apiEndpoint == "" {
+		return nil, errors.Errorf("the api endpoint of cluster %s is empty", clusterName)
 	}
 
 	token, tokenFound := secret.Data[toolchainTokenKey]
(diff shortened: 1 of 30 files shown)


pkg/cluster/service.go
```diff
@@ -178,7 +171,22 @@ func NewClusterConfig(cl client.Client, toolchainCluster *toolchainv1alpha1.Tool
...
+	}
+
+	if _, ok := secret.Data["kubeconfig"]; ok {
+		return loadConfigFromKubeConfig(toolchainCluster, secret)
+	} else {
+		return loadConfigFromLegacyToolchainCluster(toolchainCluster, secret, timeout)
+	}
+}
+
+func loadConfigFromLegacyToolchainCluster(toolchainCluster *toolchainv1alpha1.ToolchainCluster, secret *v1.Secret, timeout time.Duration) (*Config, error) {
```
MatousJobanek, line 184: I already updated the add-cluster.sh script logic so it populates the kubeconfig right away. We can now remove the migration logic and load the config directly from the kubeconfig field

pkg/cluster/service.go
```diff
@@ -201,15 +209,48 @@ func NewClusterConfig(cl client.Client, toolchainCluster *toolchainv1alpha1.Tool
...
+	if cluster == nil {
+		// this should theoretically never happen because we should not be able to get the rest config from an invalid kube config.
+		return nil, errors.New("Could not unambiguously identify the cluster configuration to use in the kubeconfig.")
+	}
+
+	return &Config{
+		Name:              toolchainCluster.Name,
+		APIEndpoint:       cluster.Server,
+		RestConfig:        restCfg,
+		OperatorNamespace: context.Namespace,
```
MatousJobanek, line 248: why not using `clientCfg.Namespace()`?

pkg/cluster/service.go
```diff
@@ -201,15 +209,48 @@ func NewClusterConfig(cl client.Client, toolchainCluster *toolchainv1alpha1.Tool
...
+
+	cluster := cfg.Clusters[context.Cluster]
+	if cluster == nil {
+		// this should theoretically never happen because we should not be able to get the rest config from an invalid kube config.
+		return nil, errors.New("Could not unambiguously identify the cluster configuration to use in the kubeconfig.")
+	}
+
+	return &Config{
+		Name:              toolchainCluster.Name,
+		APIEndpoint:       cluster.Server,
```
MatousJobanek, line 246: couldn't we get that from the `restCfg`?

MatousJobanek (commented):
I probably know why you kept the support for the legacy config. There could be a problem as soon as you would merge a support for kubeconfig only. The reason is that we load the ToolchainCluster CRs and their configs inside main.go before we start the controllers. This means that the ToolchainCluster controller that does the migration wouldn't be started yet and it wouldn't be able to migrate the legacy fields in the secret to the kubeconfig one. 
Anyway, at some point we need to make the switch, so I decided to not keep postponing this action and I updated the add-cluster.sh script so it populates the kubeconfig right away: 
https://github.com/codeready-toolchain/toolchain-cicd/pull/117

This means that we are ready to drop the migration logic from the ToolchainCluster controller and support kubeconfig only.

pkg/cluster/service.go
metlos: This has far-reaching consequences in the tests. I would rather do that in a separate PR.

MatousJobanek: Not sure which consequences you had in mind, but I'm fine with doing it in a separate PR.

MatousJobanek (commented):
Looks good, but since the "kubeconfig` field is the preferred one now, could you please update the unit tests so they load the ToolchainCluster clients from kubeconfig instead of from the legacy fields? 
I guess that it should be enough to update this helper function: https://github.com/codeready-toolchain/toolchain-common/blob/14c41a99df18cb1c9b2a941a7715922843783b38/pkg/test/cluster.go#L19-L34

metlos:
> Looks good, but since the "kubeconfig` field is the preferred one now, could you please update the unit tests so they load the ToolchainCluster clients from kubeconfig instead of from the legacy fields? I guess that it should be enough to update this helper function.

For this to make most sense, I guess we should also not use any legacy fields in the code as well as in the tests. When I do that, I get quite a substantial diff (on top of the changes in this PR):

```
$ git diff --stat
 controllers/toolchaincluster/healthchecker_test.go               | 22 ++++++++++------------
 controllers/toolchaincluster/toolchaincluster_controller.go      | 41 -----------------------------------------
 controllers/toolchaincluster/toolchaincluster_controller_test.go | 65 ++++++++++++++---------------------------------------------------
 pkg/cluster/service.go                                           | 43 +------------------------------------------
 pkg/cluster/service_test.go                                      | 80 ++++++++------------------------------------------------------------------------
 pkg/cluster/service_whitebox_test.go                             |  8 +++-----
 pkg/test/cluster.go                                              | 53 ++++++++++++++++++++++++++++++++++++++++++++++++-----
 pkg/test/verify/cluster.go                                       | 80 +++++++++++++++++---------------------------------------------------------------
 8 files changed, 101 insertions(+), 291 deletions(-)
```

I think it makes more sense to merge this PR as is and move the test updates and removal of the migration logic to a new PR.. WDYT?

MatousJobanek (approved):

pkg/cluster/service_test.go
```diff
@@ -161,3 +165,128 @@ func TestListToolchainClusterConfigs(t *testing.T) {
...
+		cfg, err := cluster.NewClusterConfig(cl, tc, 1*time.Second)
+		require.NoError(t, err)
+
+		assert.Equal(t, "https://over.the.rainbow", cfg.APIEndpoint)
+		assert.Equal(t, "operatorns", cfg.OperatorNamespace)
+		assert.Equal(t, "token", cfg.RestConfig.BearerToken)
+	})
+
+	t.Run("uses kubeconfig in precedence over legacy fields", func(t *testing.T) {
+		tc := newFormTc()
```
mfrancisc, line 277: do we need to add a test case that uses the `legacyTc`? 

basically:
```
                 tc := legacyTc()
		// Combine the kubeconfig and the token in the same secret.
		// We should see auth from the kubeconfig used...
		secret := kubeconfigSecret(t)
		secret.Data["token"] = []byte("not-the-token-we-want")
```

mfrancisc (approved):
Looks good ! 

I have one minor comment only.

pkg/cluster/service_test.go
metlos: This essentially checks that if we detect a `kubeconfig` data field in the secret then we use that instead of the data in the `ToolchainCluster` and the token in the secret.

So as soon as we see the data coming from the kubeconfig, we know that happened.

Now, you have a very good idea for a "defensive test" that would make sure any future changes wouldn't change that behavior. But we plan to completely remove the loading from the legacy fields in the upcoming PRs so I think we can afford to not have that and instead just make sure we use the new over the old.

mfrancisc: Right, I was asking since as soon as you push this PR to prod, the tc will still have the `legacy` format, and later will have the new one. But I guess it's not a concern since the only thing that matters is the loading mechanism and not the tc format.

PR #417, opened 2024-07-19
KUBESAW-129: Makefile target to check compatibility of local version with other repos

fbm3307:
This second part of dependency check 

There is a makefile target that validates compatibility of the local version in other repositories that rely on them

Similar-Related PR in API - https://github.com/codeready-toolchain/api/pull/437


verify-replace.sh
```diff
@@ -0,0 +1,43 @@
...
+TMP_DIR=/tmp/
+BASE_REPO_PATH=$(mktemp -d ${TMP_DIR}replace-verify.XXX)
+GH_BASE_URL_KS=https://github.com/kubesaw/
+GH_BASE_URL_CRT=https://github.com/codeready-toolchain/
+GH_KSCTL=${GH_BASE_URL_KS}ksctl
+GH_HOST=${GH_BASE_URL_CRT}host-operator
+GH_MEMBER=${GH_BASE_URL_CRT}member-operator
+GH_REGSVC=${GH_BASE_URL_CRT}registration-service
+GH_E2E=${GH_BASE_URL_CRT}toolchain-e2e
+GH_TC=${GH_BASE_URL_CRT}toolchain-common
```
mfrancisc, line 11: alternatively you could define an array:
```suggestion
declare -a repos=("${GH_BASE_URL_KS}ksctl" "${GH_BASE_URL_CRT}host-operator" "${GH_BASE_URL_CRT}member-operator", "${GH_BASE_URL_CRT}registration-service", ${GH_BASE_URL_CRT}toolchain-e2e, ${GH_BASE_URL_CRT}toolchain-common)
```

and loop over it with :
```
for repo in "${repos[@]}"
```

mfrancisc, line 11: I'm also not sure if toolchain-common should be here. This would download the repo in tmp and validate it which is not needed in this case right ?

mfrancisc (commented):
Looks good! I've tried it and it seems to be working now 👍 

I was wondering if we should place this script in toolchain-cicd and download it here and in api at runtime, since the [api pr](https://github.com/codeready-toolchain/api/pull/437/files) contains similar/same code if I'm not wrong.

verify-replace.sh
ranakan19: You could also separate repos for ksctl and toolchain for better readability like:
`KSCTL_REPOS=(${GH_BASE_URL_KS}ksctl)
TOOLCHAIN_REPOS=(${GH_BASE_URL_CRT}host-operator GH_MEMBER=${GH_BASE_URL_CRT}member-operator GH_REGSVC=${GH_BASE_URL_CRT}registration-service GH_E2E=${GH_BASE_URL_CRT}toolchain-e2e)
ALL_REPOS=("${KSCTL_REPOS[@]}" "${TOOLCHAIN_REPOS[@]}")`

and then loop over as francisc suggested

ranakan19, line 11:
```suggestion

```
I also don't see this being used in the for loop, so looks like this can be removed?

verify-replace.sh
```diff
@@ -0,0 +1,43 @@
...
+    echo =========================================================================================
+    echo  
+    echo $(basename $repo)
+    echo                                                                     
+    echo =========================================================================================                                            
+    REPO_PATH=$BASE_REPO_PATH/$(basename $repo)
+    echo Cloning repo in /tmp
+    git clone --depth=1 $repo $REPO_PATH
+    echo Repo cloned successfully
+    cd $REPO_PATH
```
ranakan19, line 27:
```suggestion
    cd $REPO_PATH || exit
```
or return? Incase cd fails?

verify-replace.sh
```diff
@@ -0,0 +1,43 @@
...
+GH_TC=${GH_BASE_URL_CRT}toolchain-common
+C_PATH=${PWD}
+ERRORLIST=()
+
+echo Initiating verify-replace on dependent repos
+for repo in ${GH_HOST} ${GH_REGSVC} ${GH_KSCTL} ${GH_MEMBER} ${GH_E2E}
+do
+    echo =========================================================================================
+    echo  
+    echo $(basename $repo)
```
ranakan19, line 20: My knowledge of bashscript is limited, so i don't know if this is helpful, but I see a warning in my IDE for this line:
"Quote this to prevent word splitting."

verify-replace.sh
```diff
@@ -0,0 +1,43 @@
...
+    echo                                                                     
+    echo =========================================================================================                                            
+    REPO_PATH=$BASE_REPO_PATH/$(basename $repo)
+    echo Cloning repo in /tmp
+    git clone --depth=1 $repo $REPO_PATH
+    echo Repo cloned successfully
+    cd $REPO_PATH
+    echo Initiating 'go mod replace' of current toolchain common version in dependent repos
+    go mod edit -replace github.com/codeready-toolchain/toolchain-common=$C_PATH
+    make verify-dependencies || ERRORLIST+=($(basename $repo))
```
ranakan19, line 30: another warning i see in my IDE, if it helps:
"Prefer mapfile or read -a to split command output (or quote to avoid splitting)."
I believe its for the ErrorList

ranakan19 (commented):
Looks good overall, left a few comments.
Does it delete the tmp folder after running the command? if not, can we consider adding it?

verify-replace.sh
fbm3307: yep added quotes

verify-replace.sh
fbm3307: added quotes

verify-replace.sh
fbm3307: its a good point added it

verify-replace.sh
fbm3307: removed it

fbm3307: removed it

fbm3307: i have changed it now..

fbm3307:
> Looks good overall, left a few comments. Does it delete the tmp folder after running the command? if not, can we consider adding it?

it is automatically removed/deleted once the system restarts.. i did consider  adding a delete command.. but that becomes a bit risky, incase that folder is not found .. and hence wanted to avoid it and use this..

fbm3307:
> I was wondering if we should place this script in toolchain-cicd and download it here and in api at runtime, since the [api pr](https://github.com/codeready-toolchain/api/pull/437/files) contains similar/same code if I'm not wrong.

The code is almost same, the difference is in the replace command and api has one more repo toolchain common to run the verify on .. but thought about keeping them in their respective repos for now

mfrancisc:
> The code is almost same, the difference is in the replace command and api has one more repo toolchain common to run the verify on .. but thought about keeping them in their respective repos for now

Ok makes sense for now to keep them both in their respective repos, and we can see how they evolve later on. Thanks 👍

mfrancisc (commented):
The code looks good and works fine on my machine 👍 

I have a couple of questions:
- should we place the script inside some folder , maybe create a `scripts` folder or a subfolder in `make/scripts/verify-replace.sh` if it's simpler to execute from there.

- what is the plan for adding this check to the GH CI, will that be done in follow up PRs ?

fbm3307:
> * should we place the script inside some folder , maybe create a `scripts` folder or a subfolder in `make/scripts/verify-replace.sh` if it's simpler to execute from there.
That can be done.

> * what is the plan for adding this check to the GH CI, will that be done in follow up PRs ?
this same script should work when called from github actions.. i intend to test this for CI and add this to the CI for each new PR

ranakan19 (approved):
Thank you for addressing my comments, looks good to me

mfrancisc:
> this same script should work when called from github actions.. i intend to test this for CI and add this to the CI for each new PR

thanks for the clarification. Thus you are planning to update the GH CI job in this PR.

fbm3307:
> thanks for the clarification. Thus you are planning to update the GH CI job in this PR.

I was thinking to take that as a separate PR.

mfrancisc (approved):
Looks good 👍 

Thanks for addressing my comments and answering my questions 🙏

scripts/verify-replace.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+TMP_DIR=/tmp/
+BASE_REPO_PATH=$(mktemp -d ${TMP_DIR}replace-verify.XXX)
+GH_BASE_URL_KS=https://github.com/kubesaw/
+GH_BASE_URL_CRT=https://github.com/codeready-toolchain/
+declare -a repos=("${GH_BASE_URL_KS}ksctl" "${GH_BASE_URL_CRT}host-operator" "${GH_BASE_URL_CRT}member-operator" "${GH_BASE_URL_CRT}registration-service" "${GH_BASE_URL_CRT}toolchain-e2e")
```
MatousJobanek, line 6: it's just a detail, but global variables and constants are usually ALL_CAPS,  while local can be lower_caps
```suggestion
declare -a REPOS=("${GH_BASE_URL_KS}ksctl" "${GH_BASE_URL_CRT}host-operator" "${GH_BASE_URL_CRT}member-operator" "${GH_BASE_URL_CRT}registration-service" "${GH_BASE_URL_CRT}toolchain-e2e")
```

scripts/verify-replace.sh
```diff
@@ -0,0 +1,38 @@
...
+
+echo Initiating verify-replace on dependent repos
+for repo in "${repos[@]}"
+do
+    echo =========================================================================================
+    echo  
+    echo                        "$(basename $repo)"
+    echo                                                                     
+    echo =========================================================================================                                            
+    REPO_PATH=$BASE_REPO_PATH/$(basename $repo)
```
MatousJobanek, line 18: the best practice is to always use curly brackets for variables in bash to avoid unexpected wrong expansion of variables
```suggestion
    REPO_PATH=${BASE_REPO_PATH}/$(basename ${repo})
```
could you please update it in the rest of the script?

scripts/verify-replace.sh
```diff
@@ -0,0 +1,38 @@
...
+    echo "Cloning repo in /tmp"
+    git clone --depth=1 $repo $REPO_PATH
+    echo "Repo cloned successfully"
+    cd $REPO_PATH || exit
```
MatousJobanek, line 22: I don't see any benefit in having `|| exit`

scripts/verify-replace.sh
```diff
@@ -0,0 +1,38 @@
...
+    echo                                                          
+    echo =========================================================================================
+    echo                                                           
+done
+if [ ${#ERRORLIST[@]} -ne 0 ]; then
+    echo "Below are the repos with error: "
+    for e in ${ERRORLIST[*]}
+    do
+        echo "$e"
+    done
```
MatousJobanek, line 35: we should exit with a non-zero value if something failed, right?

MatousJobanek (commented):
Looks good, a few comments

scripts/verify-replace.sh
fbm3307: removed

scripts/verify-replace.sh
fbm3307: done

scripts/verify-replace.sh
fbm3307: updated it

scripts/verify-replace.sh
fbm3307: not sure if we require the non zero exit here ?, 
here we are just displaying the error  repo names, and error list is already being checked if its empty or not , so my understanding was we dont require exit here.? 
Probably i am missing something here 🤔

verify-replace.sh
MatousJobanek: @ranakan19 could you please clarify why you would add exit?
The repo should be created by now, so not sure why cd would fail. And if it fails, then it would exit the script anyway.

scripts/verify-replace.sh
MatousJobanek: if one of the verification failed, then we want the whole makefile target to return non-zero value (in other words let it fail as well)

fbm3307: updated it

metlos (approved):

MatousJobanek (approved):
Nice :+1:

PR #426, opened 2024-08-26
Remove the migration step to kubeconfig

metlos:
Remove the migration step to kubeconfig and update the tests to not test for the legacy fields in ToolchainCluster.

This a follow-up of https://github.com/codeready-toolchain/toolchain-common/pull/412 that updates the testsuite to not rely on the legacy fields that defined the connection settings for the clusters.

This is safe to do, because all of our clusters have already been upgraded and have the kubeconfigs generated in the connection secrets.

Related PRs:
- toolchain-common: https://github.com/codeready-toolchain/toolchain-common/pull/412
- host-operator: https://github.com/codeready-toolchain/host-operator/pull/1058
- toolchain-e2e: https://github.com/codeready-toolchain/toolchain-e2e/pull/1012

### DIFF
--- a/pkg/cluster/service.go
+++ b/pkg/cluster/service.go
@@ -19,16 +19,10 @@ import (
 )
 
 const (
-	labelNamespace        = "namespace"
 	labelOwnerClusterName = "ownerClusterName"
 	LabelType             = "type"
 	// labelClusterRolePrefix is the prefix that defines the cluster role as label key
 	labelClusterRolePrefix = "cluster-role"
-
-	// These are not used
-	// toolchainAPIQPS   = 20.0
-	// toolchainAPIBurst = 30
-	// toolchainTokenKey = "token"
 )
 
 // ToolchainClusterService manages cached cluster kube clients and related ToolchainCluster CRDs
--- a/pkg/cluster/service.go
+++ b/pkg/cluster/service.go
@@ -174,48 +174,7 @@ func NewClusterConfig(cl client.Client, toolchainCluster *toolchainv1alpha1.Tool
 		return nil, errors.Wrapf(err, "unable to get secret %s for cluster %s", name, toolchainCluster.Name)
 	}
 
-	if _, ok := secret.Data["kubeconfig"]; ok {
-		return loadConfigFromKubeConfig(toolchainCluster, secret, timeout)
-	} else {
-		return loadConfigFromLegacyToolchainCluster(toolchainCluster, secret, timeout)
-	}
-}
-
-func loadConfigFromLegacyToolchainCluster(toolchainCluster *toolchainv1alpha1.ToolchainCluster, secret *v1.Secret, timeout time.Duration) (*Config, error) {
-	clusterName := toolchainCluster.Name
-
-	apiEndpoint := toolchainCluster.Spec.APIEndpoint
-	if apiEndpoint == "" {
-		return nil, errors.Errorf("the api endpoint of cluster %s is empty", clusterName)
-	}
-
-	token, tokenFound := secret.Data[toolchainTokenKey]
-	if !tokenFound || len(token) == 0 {
-		return nil, errors.Errorf("the secret for cluster %s is missing a non-empty value for %q", clusterName, toolchainTokenKey)
-	}
-
-	restConfig, err := clientcmd.BuildConfigFromFlags(apiEndpoint, "")
-	if err != nil {
-		return nil, err
-	}
-
-	if len(toolchainCluster.Spec.DisabledTLSValidations) == 1 &&
-		toolchainCluster.Spec.DisabledTLSValidations[0] == toolchainv1alpha1.TLSAll {
-		restConfig.Insecure = true
-	}
-	restConfig.BearerToken = string(token)
-	restConfig.QPS = toolchainAPIQPS
-	restConfig.Burst = toolchainAPIBurst
-	restConfig.Timeout = timeout
-
-	return &Config{
-		Name:              clusterName,
-		APIEndpoint:       apiEndpoint,
-		RestConfig:        restConfig,
-		OperatorNamespace: toolchainCluster.Labels[labelNamespace],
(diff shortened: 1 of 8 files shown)


pkg/cluster/service.go
```diff
@@ -25,9 +25,10 @@ const (
 	// labelClusterRolePrefix is the prefix that defines the cluster role as label key
 	labelClusterRolePrefix = "cluster-role"
 
-	toolchainAPIQPS   = 20.0
-	toolchainAPIBurst = 30
-	toolchainTokenKey = "token"
+	// These are not used
+	// toolchainAPIQPS   = 20.0
+	// toolchainAPIBurst = 30
+	// toolchainTokenKey = "token"
```
mfrancisc, line 31: let's maybe remove those if not used
```suggestion
```

pkg/test/verify/cluster.go
```diff
@@ -43,19 +43,22 @@ var testCases = map[string]struct {
...
 	// given
 	defer gock.Off()
 	status := test.NewClusterStatus(toolchainv1alpha1.ConditionReady, corev1.ConditionTrue)
 
 	t.Run("add member ToolchainCluster with namespace label set", func(t *testing.T) {
 		for testName, data := range testCases {
 			t.Run(testName, func(t *testing.T) {
-
-				toolchainCluster, sec := test.NewToolchainCluster("east", test.HostOperatorNs, "secret", status, Labels("member-ns", test.NameHost))
+				toolchainCluster, sec := test.NewToolchainCluster(t, "east", test.HostOperatorNs, "member-ns", "secret", status, data.insecure)
```
mfrancisc, line 54: TBH I was a bit confused by the usage of the new `data.insecure` parameter. It seems to drive the `InsecureSkipTLSVerify` flag in the kubeconfig, even if it's not explicitly/directly being checked in the tests. 

It also doesn't drive the configuration of the `DisabledTLSValidations` parameter in the `toolchaincluster` resource.

mfrancisc (commented):
Looks good 👍  

I have left only few minor comments.

pkg/cluster/service.go
metlos: Yeah... I only stumbled upon these because the linter complained about them, so I was a little bit hesitant to remove unrelated code that I know nothing about... But you're right, we should get rid of those...

pkg/test/verify/cluster.go
metlos: As far as I am aware `DisabledTLSValidations` is not used and is going away in favor of the configuration provided in the kubeconfig.

I will take a closer look at the testcases though.. I was just desperately trying to make them pass without paying too much attention to whether they actually make sense in the shape they are now.

pkg/cluster/service.go
MatousJobanek: +1 to removing them

pkg/test/verify/cluster.go
MatousJobanek: that being said, you can drop https://github.com/codeready-toolchain/toolchain-common/blob/a718700447063d91735563f4c2d2b83ee5615e05/pkg/test/verify/cluster.go#L25 as well

pkg/test/verify/cluster.go
```diff
@@ -43,19 +43,22 @@ var testCases = map[string]struct {
...
 				// the caBundle should be always ignored
 				toolchainCluster.Spec.CABundle = "ZHVtbXk="
 				toolchainCluster.Spec.DisabledTLSValidations = data.disabledTLSValidations
 
+				if toolchainCluster.Labels == nil {
+					toolchainCluster.Labels = map[string]string{}
+				}
 				toolchainCluster.Labels[cluster.RoleLabel(cluster.Tenant)] = ""
```
MatousJobanek, line 62: I guess that this can be removed - we don't rely on cluster role labels anymore, they are in SPC. 
it can be done as a separate PR since there seems to be a few more places where it is still used.

MatousJobanek (approved):
Looks good from my side, but please address @mfrancisc's comments

pkg/cluster/service_test.go
```diff
@@ -75,10 +75,10 @@ func TestListToolchainClusterConfigs(t *testing.T) {
...
 	require.NoError(t, toolchainv1alpha1.AddToScheme(scheme.Scheme))
 
-	m1, sec1 := test.NewToolchainClusterWithEndpoint("east", test.HostOperatorNs, "secret1", "http://m1.com", status, map[string]string{"ownerClusterName": "m1ClusterName", "namespace": test.MemberOperatorNs, cluster.RoleLabel(cluster.Tenant): ""})
-	m2, sec2 := test.NewToolchainClusterWithEndpoint("west", test.HostOperatorNs, "secret2", "http://m2.com", status, map[string]string{"ownerClusterName": "m2ClusterName", "namespace": test.MemberOperatorNs, cluster.RoleLabel(cluster.Tenant): ""})
-	host, secHost := test.NewToolchainCluster("host", test.MemberOperatorNs, "secretHost", status, verify.Labels(test.HostOperatorNs, "hostClusterName"))
-	noise, secNoise := test.NewToolchainCluster("noise", "noise-namespace", "secretNoise", status, verify.Labels(test.MemberOperatorNs, "noiseClusterName"))
+	m1, sec1 := test.NewToolchainClusterWithEndpoint(t, "east", test.HostOperatorNs, test.MemberOperatorNs, "secret1", "http://m1.com", status, false)
+	m2, sec2 := test.NewToolchainClusterWithEndpoint(t, "west", test.HostOperatorNs, test.MemberOperatorNs, "secret2", "http://m2.com", status, false)
```
fbm3307, line 79:
```suggestion
	m2, sec2 := test.NewToolchainClusterWithEndpoint(t, "west", test.HostOperatorNs, test.MemberOperatorNs, "secret2", "https://m2.com", status, false)
```
Should we change the above also to https?

fbm3307 (approved):
Looks Good .
Just a Minor question

PR #428, opened 2024-09-18
Don't use fields of ToolchainCluster spec

metlos:
This PR prepares `toolchain-common` for removal of most fields from the `ToolchainCluster` spec. The only field that will eventually remain is the secret reference. All the rest of the fields have either been moved to the status of the `ToolchainCluster` already or are part of the kubeconfig that is stored in the referenced secret.

NOTE: that this PR does NOT update the dependency to the API. That is to avoid breakage during the relatively long time it will take to update all the components with the necessary changes. Once we merge this PR, I will create a number of additional PRs in the rest of the components that will use the new version of `toolchain-common` and make necessary updates in the components to not use the outdated fields of `ToolchainCluster`. Only once those PRs are merged, we can go ahead and merge the PR to remove the fields from the API, followed by another set of PRs to update `toolchain-common` and subsequenlty all the other components to that new version.

Doing it this way will ensure that any changes that happen in `api` and/or `toolchain-common` in the meantime will not be affected by the breaking change.

### DIFF
--- a/pkg/test/cluster.go
+++ b/pkg/test/cluster.go
@@ -49,7 +49,6 @@ func NewToolchainClusterWithEndpoint(t *testing.T, name, tcNs, operatorNs, secNa
 				Name: secName,
 			},
 			APIEndpoint:            apiEndpoint,
-			CABundle:               "",
 			DisabledTLSValidations: []toolchainv1alpha1.TLSValidation{toolchainv1alpha1.TLSAll},
 		},
 		ObjectMeta: v1.ObjectMeta{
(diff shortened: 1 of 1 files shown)


MatousJobanek (approved):
thanks for the detailed explanation, but this is only a change in the test package, so you could easily go ahead and remove it from the api repo. 
Or do you think that there is gonna be more places where we rely on the fields in host/member/e2e/reg-service repos?

metlos:
That's a good point. There is a number of places in e2e and ksctl, but host operator, member operator and registration service don't see any changes but the update in versions. I will take a look at how the "PR train" could be simplified.

MatousJobanek:
Yeah, in this case, I wouldn't be worried so much. 
Feel free to drop it from the api right away and propagate the change through the set of PRs without any extra intermediate steps. :+1:

PR #431, opened 2024-10-01
fix make help and cleanup vendor references

mfrancisc:
Fix the make help command which was not showing the targets without comments, also remove the `vendor` support as not needed anymore.


fbm3307 (approved):
Thank you sorting this

alexeykazakov (approved):

PR #432, opened 2024-11-01
Update to the latest go-jose 2.x.

metlos:
`$TITLE`

Issue: https://issues.redhat.com/browse/SANDBOX-556


alexeykazakov (approved):
Looks good. But is the any context? I mean, sure, we want to update dependencies in general but was curious is there is something special in this case.

MatousJobanek (approved):
is this related to some CVE?

go.mod
```diff
@@ -53,14 +53,11 @@ require (
 	github.com/emicklei/go-restful/v3 v3.8.0 // indirect
 	github.com/evanphx/json-patch v5.6.0+incompatible // indirect
 	github.com/evanphx/json-patch/v5 v5.6.0 // indirect
-	github.com/fatih/color v1.12.0 // indirect
 	github.com/fsnotify/fsnotify v1.5.4 // indirect
-	github.com/go-bindata/go-bindata v3.1.2+incompatible // indirect
```
fbm3307, line 58: I see go-bin data being removed, we use go-bin data at many places , I am not sure if we can remove this directly, and if we can, does the places  using go-bin data get an alternative ?

metlos: Yep, this is just a consequence of updating `go-jose`. I just trusted `go mod` to do the right thing with the rest of the dep changes.

metlos:
> Looks good. But is the any context? I mean, sure, we want to update dependencies in general but was curious is there is something special in this case.

Good point, @alexeykazakov. I added the link to the corresponding issue to the PR description.

PR #435, opened 2024-11-21
wip generic predicates

metlos:
This a WIP of the proposed changes to unify the test assertions.


pkg/test/assertions.go
```diff
@@ -1,18 +1,16 @@
-package assertions
+package test
```
rsoaresd, line 1: We could also add `UpdateResource` in order to clean up our duplicated update functions on toolchain-e2e repo (ex. UpdateUserSignup, UpdateSpaceBinding, etc.). Something like this:

```
func UpdateResource[T any](
    t T,
    client client.Client,
    namespace string,
    resourceName string,
    status bool,
    modifyResource func(resource *T),
) (*T, error) {
    var resource *T
    err := wait.PollUntilContextTimeout(context.TODO(), time.Second*2, time.Minute*5, true, func(ctx context.Context) (done bool, err error) {
        // Fetch the resource from the cluster
        freshResource := new(T)
        if err := client.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: resourceName}, freshResource); err != nil {
            return true, err
        }

        // Get the resource type name
        resourceType := reflect.TypeOf(freshResource).Elem().Name()

        // Apply the modification function
        modifyResource(freshResource)

        // Update resource Spec or Status
        if status {
            if err := client.Status().Update(context.TODO(), freshResource); err != nil {
                t.Logf("Error updating %s Status '%s': %s. Will retry...", resourceType, resourceName, err.Error())
                return false, nil
            }
        } else {
            if err := client.Update(context.TODO(), freshResource); err != nil {
                t.Logf("Error updating %s Spec '%s': %s. Will retry...", resourceType, resourceName, err.Error())
                return false, nil
            }
        }

        resource = freshResource
        return true, nil
    })

    return resource, err
}
```

pkg/test/assertions_test.go
```diff
@@ -0,0 +1,126 @@
...
+			},
+		}
+
+		pred := ConditionThat(toolchainv1alpha1.ConditionType("test"), HasStatus(corev1.ConditionTrue))
+
+		expl := Explain(Has(pred), actual)
+
+		assert.True(t, strings.HasPrefix(expl, "predicate 'test.conditionsPredicate' didn't match the object because of the following"))
+	})
+}
```
rsoaresd, line 76: To avoid duplication of the test cases and to make the tests easier to read, we can use a table-driven approach, if you agree. Something like this:
```suggestion
func TestExplain(t *testing.T) {
	tests := []struct {
		name         string
		actual       any
		predicate    Predicate[any]
		expectedText string
		contains     []string
	}{
		{
			name:         "with diff",
			actual:       &corev1.Secret{Name: "actual"},
			predicate:    Has(Name("expected")),
			expectedText: "predicate 'test.named' didn't match the object because of the following differences",
			contains:     []string{"-", "\"expected\"", "+", "\"actual\""},
		},
		{
			name:         "without diff",
			actual:       &corev1.Secret{Name: "actual"},
			predicate:    Is[client.Object](&predicateWithoutFixing{}),
			expectedText: "predicate 'test.predicateWithoutFixing' didn't match the object",
			contains:     nil,
		},
		{
			name: "with a slice",
			actual: []int{1, 2, 3},
			predicate: MockPredicate[[]int]{
				MatchesFunc: func(v []int) bool { return false },
				FixToMatchFunc: func(v []int) []int {
					return []int{1, 2}
				},
			},
			expectedText: "predicate 'test.MockPredicate[[]int]' didn't match the object because of the following",
			contains:     nil,
		},
		{
			name: "with conditions",
			actual: []toolchainv1alpha1.Condition{
				{
					Type:   toolchainv1alpha1.ConditionType("test"),
					Status: corev1.ConditionFalse,
					Reason: "because",
				},
			},
			predicate: Has(ConditionThat(
				toolchainv1alpha1.ConditionType("test"),
				HasStatus(corev1.ConditionTrue),
			)),
			expectedText: "predicate 'test.conditionsPredicate' didn't match the object because of the following",
			contains:     nil,
		},
	}

	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			// when
			expl := Explain(tt.predicate, tt.actual)

			// then
			assert.True(t, strings.HasPrefix(expl, tt.expectedText))
			if tt.contains != nil {
				for _, part := range tt.contains {
					assert.Contains(t, expl, part)
				}
			}
		})
	}
}
```

rsoaresd (commented):
@metlos Amazing work! 🚀  Tysm for driving this 🏅

Just minor suggestions:

PR #436, opened 2024-12-03
fix GetBannedUser

rsoaresd:
# Description
We should return an error when no BannedUser is found


MatousJobanek (approved):

mfrancisc (approved):

mfrancisc:
Maybe you can return a specific type of error in this case, like:
```
return errors.NewNotFound(schema.GroupResource{}, key.Name)
```
and update the ksctl function to ignore this error.

rajivnathan:
+1 to @mfrancisc suggestion

PR #446, opened 2025-01-10
feat: new apply functionality

mfrancisc:
THIS IS EXPERIMENTAL and the PR might be closed! 

This is an attempt to leverage the kubectl apply package for applying objects on the clusters.

Unfortunately atm the package only supports applying from files and stdin, it doesn't support custom input. 
See: https://github.com/kubernetes/kubectl/issues/1670

wrt: [KUBESAW-139](https://issues.redhat.com/browse/KUBESAW-139)


pkg/client/kubectl_client.go
```diff
@@ -448,3 +466,31 @@ func (k *KubectlResourceOperations) authReconcile(obj *unstructured.Unstructured
...
+	orig *os.File
+}
+
+func pumpToStdin(data io.Reader) (*replacedStdin, error) {
+	stdinRead, stdinWrite, err := os.Pipe()
+	if err != nil {
+		return nil, err
+	}
+
+	f := &replacedStdin{orig: os.Stdin}
```
mfrancisc, line 480: why pointing os.File to os.Stdin ?

pkg/client/kubectl_client.go
```diff
@@ -448,3 +466,31 @@ func (k *KubectlResourceOperations) authReconcile(obj *unstructured.Unstructured
...
+
+	os.Stdin = stdinRead
```
mfrancisc, line 482: again , I'm not sure I follow why we need this . Could we point `stdinRead` to `*os.File` directly ? 

Also does this mean it's using a file for storing the data or it's using the stdin stream?

pkg/client/kubectl_client.go
```diff
@@ -448,3 +466,31 @@ func (k *KubectlResourceOperations) authReconcile(obj *unstructured.Unstructured
...
+
+	if _, err = io.Copy(stdinWrite, data); err != nil {
+		os.Stdin = f.orig
```
mfrancisc, line 485: why setting back os.Stdin to f only on error ?

pkg/client/kubectl_client.go
```diff
@@ -448,3 +466,31 @@ func (k *KubectlResourceOperations) authReconcile(obj *unstructured.Unstructured
 	}
 	return strings.Join(out, ". "), nil
 }
+
+type replacedStdin struct {
+	orig *os.File
+}
+
+func pumpToStdin(data io.Reader) (*replacedStdin, error) {
```
mfrancisc, line 474: could you please add a few comments explaining how the "trick" works ? I'm not 100% following all the swaps between the various variables ...

PR #450, opened 2025-01-20
KUBESAW-174: Print error output for repos where verification failed

fbm3307:
This PR is to update the verify dependency script such a way that it prints the error/failure in an understandable way

Similar PR
- API - https://github.com/codeready-toolchain/api/pull/458

### DIFF
--- a/scripts/verify-replace.sh
+++ b/scripts/verify-replace.sh
@@ -71,7 +71,7 @@ if [ ${#ERROR_REPO_LIST[@]} -ne 0 ]; then
         for std_out_file_name in ${STD_OUT_FILE_LIST[*]}
             do
                 if [[ ${std_out_file_name} =~ ${error_repo_name} ]]; then 
-                    cat "${std_out_file_name}" | grep -E ${GO_LINT_REGEX}|${ERROR_REGEX}
+                    cat "${std_out_file_name}" | grep  "${GO_LINT_REGEX}\|${ERROR_REGEX}"
                 fi
         done                                             
     done
--- a/scripts/verify-replace.sh
+++ b/scripts/verify-replace.sh
@@ -68,7 +68,7 @@ if [ ${#ERROR_REPO_LIST[@]} -ne 0 ]; then
         for std_out_file_name in ${STD_OUT_FILE_LIST[*]}
             do
                 if [[ ${std_out_file_name} =~ ${error_repo_name} ]]; then 
-                    cat "${std_out_file_name}" | grep -E ${GO_LINT_REGEX}|${ERROR_REGEX}|${FAIL_REGEX}
+                    cat "${std_out_file_name}" | grep -E ${GO_LINT_REGEX}|${ERROR_REGEX}
                 fi
         done                                             
     done
--- a/scripts/verify-replace.sh
+++ b/scripts/verify-replace.sh
@@ -9,7 +9,10 @@ ERROR_REPO_LIST=()
 ERROR_FILE_LIST=()
 STD_OUT_FILE_LIST=()
 GO_LINT_REGEX="[\s\w.\/]*:[0-9]*:[0-9]*:[\w\s)(*.\`]*"
-ERROR_REGEX="Error[:]*/|FAIL[:]*" #unit test or any other failure we log from our controllers other places goes into stdoutput, hence making that regex too, to fetch the error more precisely
+ERROR_REGEX="[E\|e][R\|r][R\|r][O\|o][R\|r][:]*\|[F\|f][A\|a][I\|i][l\|L][:]*\|expected[:]*\|actual[:]*" #unit test or any other failure we log from our controllers or other places goes into stdoutput 
+                                                            #(since we log it and its not a failure in running the command or dependency check), 
+                                                            #hence making that regex too, to fetch the error more precisely
+
 echo Initiating verify-replace on dependent repos
 for repo in "${REPOS[@]}"
 do
--- a/scripts/verify-replace.sh
+++ b/scripts/verify-replace.sh
@@ -9,7 +9,6 @@ ERRORREPOLIST=()
 ERRORFILELIST=()
 STDOUTFILELIST=()
 GOLINTREGEX="[\s\w.\/]*:[0-9]*:[0-9]*:[\w\s)(*.\`]*"
-LINTERERRORFILE=$(mktemp ${BASE_REPO_PATH}/LinterError.XXX)
 
 echo Initiating verify-replace on dependent repos
 for repo in "${REPOS[@]}"
@@ -70,10 +69,7 @@ if [ ${#ERRORREPOLIST[@]} -ne 0 ]; then
                 for stdoutfilename in ${STDOUTFILELIST[*]}
                 do
                     if [[ ${stdoutfilename} =~ ${errorreponame} ]]; then 
-                        cat "${stdoutfilename}" | grep -E ${GOLINTREGEX} > ${LINTERERRORFILE}
-                        if [ -s "${LINTERERRORFILE}" ]; then
-                            cat  "${LINTERERRORFILE}"   
-                        fi
+                        cat "${stdoutfilename}" | grep -E ${GOLINTREGEX} 
                     fi
                 done                                            
             fi
--- a/scripts/verify-replace.sh
+++ b/scripts/verify-replace.sh
@@ -63,9 +63,9 @@ if [ ${#ERRORREPOLIST[@]} -ne 0 ]; then
                 for s in ${STDOUTFILELIST[*]}
                 do
                     if [[ ${s} =~ ${e} ]]; then 
-                        cat "${s}" | grep -E ${GOLINTREGEX} > lintererror
-                        if ! [ -s "lintererror" ]; then
-                            cat  lintererror   
+                        cat "${s}" | grep -E ${GOLINTREGEX} > ${LINTERERRORFILE}
+                        if ! [ -s "${LINTERERRORFILE}" ]; then
+                            cat  "${LINTERERRORFILE}"   
                         fi
                     fi
                 done                                            
--- a/scripts/verify-replace.sh
+++ b/scripts/verify-replace.sh
@@ -5,7 +5,10 @@ GH_BASE_URL_KS=https://github.com/kubesaw/
 GH_BASE_URL_CRT=https://github.com/codeready-toolchain/
 declare -a REPOS=("${GH_BASE_URL_KS}ksctl" "${GH_BASE_URL_CRT}host-operator" "${GH_BASE_URL_CRT}member-operator" "${GH_BASE_URL_CRT}registration-service" "${GH_BASE_URL_CRT}toolchain-e2e")
 C_PATH=${PWD}
-ERRORLIST=()
+ERRORREPOLIST=()
+ERRORFILELIST=()
+GOLINTREGEX="[\s\w.\/]*:[0-9]*:[0-9]*:[\s\w]*[:]*[\s\w\">.-]*[\w\s)(*.]*"
+ERRORREGEX="Error[:]*"
 
 echo Initiating verify-replace on dependent repos
 for repo in "${REPOS[@]}"
@@ -16,26 +19,44 @@ do
     echo                                                                     
     echo =========================================================================================                                            
     repo_path=${BASE_REPO_PATH}/$(basename ${repo})
+    ERRFILE=$(mktemp ${TMP_DIR}$(basename ${repo}).XXX)
     echo "Cloning repo in /tmp"
     git clone --depth=1 ${repo} ${repo_path}
     echo "Repo cloned successfully"
     cd ${repo_path}
     if ! make pre-verify; then
-        ERRORLIST+="($(basename ${repo}))"
+        ERRORREPOLIST+="($(basename ${repo}))"
         continue
     fi
     echo "Initiating 'go mod replace' of current toolchain common version in dependent repos"
     go mod edit -replace github.com/codeready-toolchain/toolchain-common=${C_PATH}
-    make verify-dependencies || ERRORLIST+="($(basename ${repo}))"
+    make verify-dependencies &> ${ERRFILE} 
+    rc=$?
+    if [ ${rc} -ne 0 ]; then
+    ERRORREPOLIST+="($(basename ${repo}))" 
+    ERRORFILELIST+="${ERRFILE}  "
+    fi
     echo                                                          
     echo =========================================================================================
     echo                                                           
 done
-if [ ${#ERRORLIST[@]} -ne 0 ]; then
+echo "Summary"
+if [ ${#ERRORREPOLIST[@]} -ne 0 ]; then
     echo "Below are the repos with error: "
-    for e in ${ERRORLIST[*]}
+    for e in ${ERRORREPOLIST[*]}
     do
         echo "${e}"
+    done
+    for c in ${ERRORFILELIST[*]}
+    do  
+        echo                                                          
+        echo =========================================================================================
+        echo 
+        echo "${c} has the following errors "
+        echo                                                          
+        echo =========================================================================================
+        echo
+        cat "${c}" | grep "${GOLINTREGEX}\|${ERRORREGEX}"
     done
     exit 1
 else
--- a/scripts/verify-replace.sh
+++ b/scripts/verify-replace.sh
@@ -5,10 +5,10 @@ GH_BASE_URL_KS=https://github.com/kubesaw/
 GH_BASE_URL_CRT=https://github.com/codeready-toolchain/
 declare -a REPOS=("${GH_BASE_URL_KS}ksctl" "${GH_BASE_URL_CRT}host-operator" "${GH_BASE_URL_CRT}member-operator" "${GH_BASE_URL_CRT}registration-service" "${GH_BASE_URL_CRT}toolchain-e2e")
 C_PATH=${PWD}
-ERRORREPOLIST=()
-ERRORFILELIST=()
-STDOUTFILELIST=()
-GOLINTREGEX="[\s\w.\/]*:[0-9]*:[0-9]*:[\w\s)(*.\`]*"
+ERROR_REPO_LIST=()
+ERROR_FILE_LIST=()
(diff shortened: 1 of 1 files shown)


scripts/verify-replace.sh
```diff
@@ -16,27 +19,45 @@ do
     echo                                                                     
     echo =========================================================================================                                            
     repo_path=${BASE_REPO_PATH}/$(basename ${repo})
+    ERRFILE=$(mktemp ${TMP_DIR}$(basename ${repo}).XXX)
```
metlos, line 22: IMHO, the file extension could be something more descriptive, like `.errors`.

scripts/verify-replace.sh
```diff
@@ -16,27 +19,45 @@ do
...
+        ERRORREPOLIST+="($(basename ${repo}))"
         continue
     fi
     echo "Initiating 'go mod replace' of current toolchain common version in dependent repos"
     go mod edit -replace github.com/codeready-toolchain/toolchain-common=${C_PATH}
-    make verify-dependencies || ERRORLIST+="($(basename ${repo}))"
+    make verify-dependencies &> ${ERRFILE} 
+    rc=$?
+    if [ ${rc} -ne 0 ]; then
+    ERRORREPOLIST+="($(basename ${repo}))"
```
metlos, line 36: Should there be a trailing space here, too, as it is in the `ERRORFILELIST`?

metlos (commented):
Looks good. Only a couple of comments below..

scripts/verify-replace.sh
fbm3307: It would be okay if the exact same named files are created in CI because these are tmp files which are automatically deleted once the workspace is deleted after each ci run. 
But if this same script someone runs locally, its created in /tmp folder, and it gets deleted once the local env restarts. if someone does 2-3 runs locally without restarting the local, it will be a problem
I did think about deletion of files, but didn't want to deal with the risk of deleting something accidentally.
This was the reason to have the Random word generated as the file extension so that the file names are always unique.
But i do agree to have a self explanatory file  name and hence added -error in the file name rather than extension
eg.`(toolchain-common-error.b3m)`

scripts/verify-replace.sh
fbm3307: i added the trailing space now , i didn't do it previously, as my thought was to have all the repo names printed in a single line, but i guess its better to have them in 1 line each

scripts/verify-replace.sh
```diff
@@ -16,27 +19,45 @@ do
...
     cd ${repo_path}
     if ! make pre-verify; then
-        ERRORLIST+="($(basename ${repo}))"
+        ERRORREPOLIST+="$(basename ${repo}) "
         continue
     fi
     echo "Initiating 'go mod replace' of current toolchain common version in dependent repos"
     go mod edit -replace github.com/codeready-toolchain/toolchain-common=${C_PATH}
-    make verify-dependencies || ERRORLIST+="($(basename ${repo}))"
+    make verify-dependencies &> ${ERRFILE}
```
MatousJobanek, line 33: IIRC, you mentioned that the only process that doesn't print errors to a stderr is the linter, wouldn't it be better to redirect the outputs to separate files and:
* print the whole content of the file with stderr
* and parse the stdin with the `GOLINTREGEX`?

In this way, you could limit the dependency on the error output formatting.

scripts/verify-replace.sh
```diff
@@ -16,27 +19,45 @@ do
...
+    for c in ${ERRORFILELIST[*]}
+    do  
+        echo                                                          
+        echo =========================================================================================
+        echo 
+        echo                       "${c} has the following errors "
+        echo                                                          
+        echo =========================================================================================
+        echo
+        cat "${c}" | grep "${GOLINTREGEX}\|${ERRORREGEX}"
```
MatousJobanek, line 59: since you use regexps, shouldn't it has `grep -E`?

scripts/verify-replace.sh
```diff
@@ -5,7 +5,10 @@ GH_BASE_URL_KS=https://github.com/kubesaw/
 GH_BASE_URL_CRT=https://github.com/codeready-toolchain/
 declare -a REPOS=("${GH_BASE_URL_KS}ksctl" "${GH_BASE_URL_CRT}host-operator" "${GH_BASE_URL_CRT}member-operator" "${GH_BASE_URL_CRT}registration-service" "${GH_BASE_URL_CRT}toolchain-e2e")
 C_PATH=${PWD}
-ERRORLIST=()
+ERRORREPOLIST=()
+ERRORFILELIST=()
+GOLINTREGEX="[\s\w.\/]*:[0-9]*:[0-9]*:[\w\s)(*.\`]*"
+ERRORREGEX="Error[:]*"
```
MatousJobanek, line 11: Could you please clarify what the `ERRORREGEX` is used for?

scripts/verify-replace.sh
fbm3307: Yes, I like your idea of reducing the dependency, I have changed the logic now. PTAL

scripts/verify-replace.sh
fbm3307: I used quotes instead of `-E`, which did the job.
but i have now used `-E` in the command

scripts/verify-replace.sh
fbm3307: I used it to get the error sentences like below and also any other error which would be there
```
make[1]: *** [vet] Error 1
make: *** [verify-replace-run] Error 1
```
but since i have changed the logic, it no longer required.

scripts/verify-replace.sh
```diff
@@ -16,26 +20,57 @@ do
...
     fi
     echo "Initiating 'go mod replace' of current toolchain common version in dependent repos"
     go mod edit -replace github.com/codeready-toolchain/toolchain-common=${C_PATH}
-    make verify-dependencies || ERRORLIST+="($(basename ${repo}))"
+    make verify-dependencies 2> ${ERRFILE} 1> ${STDOUTFILE}
+    rc=$?
+    if [ ${rc} -ne 0 ]; then
+    ERRORREPOLIST+="$(basename ${repo}) " 
+    ERRORFILELIST+="${ERRFILE}  "
+    STDOUTFILELIST="${STDOUTFILE} "
```
MatousJobanek, line 40: isn't it missing the plus `+`?
```suggestion
    STDOUTFILELIST+="${STDOUTFILE} "
```

scripts/verify-replace.sh
```diff
@@ -16,26 +20,57 @@ do
...
+echo                "Summary"
+if [ ${#ERRORREPOLIST[@]} -ne 0 ]; then
     echo "Below are the repos with error: "
-    for e in ${ERRORLIST[*]}
+    for e in ${ERRORREPOLIST[*]}
     do
-        echo "${e}"
+        for c in ${ERRORFILELIST[*]}
+        do
+            if [[ ${c} =~ ${e} ]]; then
```
MatousJobanek, line 53: not using just a single chars, bust whole words for the variable would significantly improve the reading of the code

scripts/verify-replace.sh
```diff
@@ -16,26 +20,57 @@ do
...
     cd ${repo_path}
     if ! make pre-verify; then
-        ERRORLIST+="($(basename ${repo}))"
+        ERRORREPOLIST+="$(basename ${repo}) "
         continue
     fi
     echo "Initiating 'go mod replace' of current toolchain common version in dependent repos"
     go mod edit -replace github.com/codeready-toolchain/toolchain-common=${C_PATH}
-    make verify-dependencies || ERRORLIST+="($(basename ${repo}))"
+    make verify-dependencies 2> ${ERRFILE} 1> ${STDOUTFILE}
```
MatousJobanek, line 35: we need to print the copy of the output also to the terminal. There are two reasons:
1. important for local execution to follow on the progress of the execution. Otherwise, it just stays only with `Initiating 'go mod replace' of current toolchain common version in dependent repos` and stays silent
2. To see the whole context of the potential erros. If the regex selection of the errors doesn't include everything, then it's good to scroll up and see the whole output.

scripts/verify-replace.sh
```diff
@@ -16,26 +20,57 @@ do
     echo                                                                     
     echo =========================================================================================                                            
     repo_path=${BASE_REPO_PATH}/$(basename ${repo})
+    ERRFILE=$(mktemp ${TMP_DIR}$(basename ${repo})-error.XXX)
```
MatousJobanek, line 23: let's keep some consistency with the case of the variables. it's fine to keep the local variables lower-case, but let's apply the rule to all of them.

scripts/verify-replace.sh
```diff
@@ -16,26 +20,57 @@ do
...
+    STDOUTFILE=$(mktemp ${TMP_DIR}$(basename ${repo})-output.XXX)
```
MatousJobanek, line 24: grouping them under the `BASE_REPO_PATH` would make more sense. and let's also print them. Something like this:
```suggestion
    ERRFILE=$(mktemp ${BASE_REPO_PATH}/$(basename ${repo})-error.XXX)
    STDOUTFILE=$(mktemp ${BASE_REPO_PATH}/$(basename ${repo})-output.XXX)
    echo "Std output file ${STDOUTFILE}"
    echo "Error output file ${ERRFILE}"
```

scripts/verify-replace.sh
```diff
@@ -5,7 +5,11 @@ GH_BASE_URL_KS=https://github.com/kubesaw/
 GH_BASE_URL_CRT=https://github.com/codeready-toolchain/
 declare -a REPOS=("${GH_BASE_URL_KS}ksctl" "${GH_BASE_URL_CRT}host-operator" "${GH_BASE_URL_CRT}member-operator" "${GH_BASE_URL_CRT}registration-service" "${GH_BASE_URL_CRT}toolchain-e2e")
 C_PATH=${PWD}
-ERRORLIST=()
+ERRORREPOLIST=()
+ERRORFILELIST=()
+STDOUTFILELIST=()
+GOLINTREGEX="[\s\w.\/]*:[0-9]*:[0-9]*:[\w\s)(*.\`]*"
+LINTERERRORFILE=$(mktemp ${TMP_DIR}LinterError.XXX)
```
MatousJobanek, line 12: the same with grouping the files under the same directory
```suggestion
LINTERERRORFILE=$(mktemp ${BASE_REPO_PATH}/LinterError.XXX)
```

scripts/verify-replace.sh
```diff
@@ -16,26 +20,57 @@ do
...
+            else
+                for s in ${STDOUTFILELIST[*]}
+                do
+                    if [[ ${s} =~ ${e} ]]; then 
+                        cat "${s}" | grep -E ${GOLINTREGEX} > ${LINTERERRORFILE}
+                        if ! [ -s "${LINTERERRORFILE}" ]; then
+                            cat  "${LINTERERRORFILE}"   
+                        fi
+                    fi
+                done
```
MatousJobanek, line 71: why is this inside of the `else` statement? can you add some comments explaining that?

scripts/verify-replace.sh
```diff
@@ -16,26 +20,57 @@ do
...
+                echo                                                          
+                echo =========================================================================================
+                echo 
+                cat "${c}"
+            else
+                for s in ${STDOUTFILELIST[*]}
+                do
+                    if [[ ${s} =~ ${e} ]]; then 
+                        cat "${s}" | grep -E ${GOLINTREGEX} > ${LINTERERRORFILE}
+                        if ! [ -s "${LINTERERRORFILE}" ]; then
```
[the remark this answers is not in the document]
fbm3307: i just got mixed up in 0 & 1 and thought the other way around, you are right, i have updated it now

scripts/verify-replace.sh
fbm3307: updated it with comments

scripts/verify-replace.sh
fbm3307: okay yes i agree , updated and grouped it

scripts/verify-replace.sh
fbm3307: done

scripts/verify-replace.sh
fbm3307: ahh , yes changed it now

scripts/verify-replace.sh
fbm3307: okay i agree, updated it to better names

scripts/verify-replace.sh
fbm3307: ideally yes, i just didn't see it and don't know how it worked without the plus sign, probably i tested wrong.
changed it now

scripts/verify-replace.sh
```diff
@@ -16,26 +19,61 @@ do
...
+            # Meaning if the other checks have passed then only it wil proceed to check golint and we would parse 
+            # golint errors from std files if any.
+                for stdoutfilename in ${STDOUTFILELIST[*]}
+                do
+                    if [[ ${stdoutfilename} =~ ${errorreponame} ]]; then 
+                        cat "${stdoutfilename}" | grep -E ${GOLINTREGEX} 
+                    fi
+                done                                            
+            fi
+        done
```
MatousJobanek, line 76: There is something weird with these nested for cycles.
There is one stderr file and one stdout file per repo, right?
The first for cycle goes through all error file names and tries to find the corresponding one for the failed repo. For every file whose name doesn't match the expected repo the code goes into the `else` part and starts a loop for all stdout file names trying to find matching one also there. 
But why doing it for every single time when the `errorfilename` doesn't match `errorreponame`? Why having the for loop nested inside the first one and as part of the `else` section? why not having the second loop at the same level as the first one?

scripts/verify-replace.sh
```diff
@@ -5,7 +5,10 @@ GH_BASE_URL_KS=https://github.com/kubesaw/
 GH_BASE_URL_CRT=https://github.com/codeready-toolchain/
 declare -a REPOS=("${GH_BASE_URL_KS}ksctl" "${GH_BASE_URL_CRT}host-operator" "${GH_BASE_URL_CRT}member-operator" "${GH_BASE_URL_CRT}registration-service" "${GH_BASE_URL_CRT}toolchain-e2e")
 C_PATH=${PWD}
-ERRORLIST=()
+ERRORREPOLIST=()
+ERRORFILELIST=()
+STDOUTFILELIST=()
+GOLINTREGEX="[\s\w.\/]*:[0-9]*:[0-9]*:[\w\s)(*.\`]*"
```
MatousJobanek, line 11: Let's keep consistency:
```suggestion
ERROR_REPO_LIST=()
ERROR_FILE_LIST=()
STDOUT_FILE_LIST=()
GO_LINT_REGEX="[\s\w.\/]*:[0-9]*:[0-9]*:[\w\s)(*.\`]*"
```

scripts/verify-replace.sh
```diff
@@ -16,26 +19,61 @@ do
...
 done
-if [ ${#ERRORLIST[@]} -ne 0 ]; then
+echo                "Summary"
+if [ ${#ERRORREPOLIST[@]} -ne 0 ]; then
     echo "Below are the repos with error: "
-    for e in ${ERRORLIST[*]}
+    for errorreponame in ${ERRORREPOLIST[*]}
     do
-        echo "${e}"
+        for errorfilename in ${ERRORFILELIST[*]}
```
MatousJobanek, line 52: let's keep consistency in the format of the variables in your script
```go
    for error_repo_name in ${ERROR_REPO_LIST[*]}
    do
        for error_file_name in ${ERROR_FILE_LIST[*]}
```

scripts/verify-replace.sh
fbm3307: Okay the logic you suggested can also be one option, 
what i thought and as i wrote in the comments that, we get `go lint error` only if it hasn't failed in the previous checks (go-lint check is always last check) , and since i had to anyways go through a separate for loop i thought of checking only if there is no error file in the previous list then only go and check in the stdout file for go-lint errors

fbm3307: Okay as we discussed, i would move it to the same level

MatousJobanek (approved):
Looks good, thanks for addressing my comments

scripts/verify-replace.sh
```diff
@@ -16,26 +22,58 @@ do
...
+        for error_file_name in ${ERROR_FILE_LIST[*]}
+        do
+            if [[ ${error_file_name} =~ ${error_repo_name} ]]; then 
+                echo                                                          
+                echo =========================================================================================
+                echo 
+                echo                       "${error_repo_name} has the following errors "
+                echo                                                          
+                echo =========================================================================================
+                echo
```
MatousJobanek, line 67: it would make more sense to move the error banner out of the for loop
```suggestion
        echo                                                          
        echo =========================================================================================
        echo 
        echo                       "${error_repo_name} has the following errors "
        echo                                                          
        echo =========================================================================================
        echo 
        for error_file_name in ${ERROR_FILE_LIST[*]}
        do
            if [[ ${error_file_name} =~ ${error_repo_name} ]]; then 
```

MatousJobanek (approved):
last suggestion

scripts/verify-replace.sh
```diff
@@ -16,26 +22,58 @@ do
...
+                echo                                                          
+                echo =========================================================================================
+                echo 
+                cat "${error_file_name}"
+            fi
+        done
+        for std_out_file_name in ${STD_OUT_FILE_LIST[*]}
+            do
+                if [[ ${std_out_file_name} =~ ${error_repo_name} ]]; then 
+                    cat "${std_out_file_name}" | grep  "${GO_LINT_REGEX}\|${ERROR_REGEX}"
```
MatousJobanek, line 74: from the last failure (in member-operator), it's apparent that the selection doesn't include everything - eg it's missing the line of the failed test (which is really missing to understand the problem)
```
                        	            				/tmp/replace-verify.OGM/member-operator/controllers/idler/idler_controller_test.go:259
```
https://github.com/codeready-toolchain/toolchain-common/actions/runs/13068368416/job/36464570446?pr=450#step:5:662

I would add `-C 5` to make sure that the output contains the whole context
```suggestion
                    cat "${std_out_file_name}" | grep -C 5  "${GO_LINT_REGEX}\|${ERROR_REGEX}"
```

PR #452, opened 2025-01-24
move EncodeUserIdentifier & signup helper functions

MatousJobanek:
move EncodeUserIdentifier function here and add a few UserSignup helper function. The most important one is the `WithEncodedName` one - to have an easy way of using the encoded value as the name of the UserSignup and the original value as the preffered username.

[KUBESAW-254](https://issues.redhat.com/browse/KUBESAW-254)


mfrancisc (approved):
Nice 👍

alexeykazakov (approved):
:+1:

PR #453, opened 2025-02-04
test: metric assertions and coomparisons

xcoulon:
new helper funcs

### DIFF
--- a/pkg/test/metrics/metric.go
+++ b/pkg/test/metrics/metric.go
@@ -12,6 +12,14 @@ func AssertMetricsCounterEquals(t *testing.T, expected int, c prometheus.Counter
 	assert.InDelta(t, float64(expected), promtestutil.ToFloat64(c), 0.01)
 }
 
+func AssertCounterEqualsInt(t *testing.T, expected int, c prometheus.Counter) {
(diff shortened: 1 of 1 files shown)


MatousJobanek (approved):

alexeykazakov (approved):

pkg/test/metrics/metric.go
```diff
@@ -12,6 +12,14 @@ func AssertMetricsCounterEquals(t *testing.T, expected int, c prometheus.Counter
 	assert.InDelta(t, float64(expected), promtestutil.ToFloat64(c), 0.01)
 }
 
+func AssertCounterEqualsInt(t *testing.T, expected int, c prometheus.Counter) {
+	assert.Equal(t, expected, int(promtestutil.ToFloat64(c)))
```
mfrancisc, line 16: is it actually required to convert to float64 and than to int? I believe so, but just curious.

mfrancisc (approved):

pkg/test/metrics/metric.go
xcoulon: yes, if you want to compare to an `int` in the `assert.Equal` func.

PR #454, opened 2025-02-05
KUBESAW-147: Space controller updates tier-hash label based on the NSTemplateTier.Status.Revisions field

fbm3307:
This PR is to update the space controller to compute the tier-hash label value using the NSTemplateTier.Status.Revisions field so that it can match outdated spaces using the new revisions mechanism.

Jira-Link - [KUBESAW-147](https://issues.redhat.com/browse/KUBESAW-147)

Related PR
- Host-Operator - https://github.com/codeready-toolchain/host-operator/pull/1135
- Toolchain-e2e - https://github.com/codeready-toolchain/toolchain-e2e/pull/1109


pkg/hash/hash.go
```diff
@@ -12,14 +12,20 @@ func TemplateTierHashLabelKey(tierName string) string {
...
+// to calculate hash
 func ComputeHashForNSTemplateTier(tier *toolchainv1alpha1.NSTemplateTier) (string, error) {
 	refs := []string{}
 	for _, ns := range tier.Spec.Namespaces {
-		refs = append(refs, ns.TemplateRef)
+		refs = append(refs, tier.Status.Revisions[ns.TemplateRef])
 	}
 	if tier.Spec.ClusterResources != nil {
-		refs = append(refs, tier.Spec.ClusterResources.TemplateRef)
+		refs = append(refs, tier.Status.Revisions[tier.Spec.ClusterResources.TemplateRef])
```
MatousJobanek, line 28: As it is mentioned in the Jira and also in the design document, we should use the whole revisions map to collect the refs and compute the hash, not only specific references.
It was also mentioned in the other threads that the current logic has a bug that it doesn't include the refs of the role templates - this should be fixed by collecting all revisions from the map

fbm3307: Currently we have extra values  and we do not have the logic of delete/clean up of revisions field, thats the reason i did this https://github.com/codeready-toolchain/toolchain-common/pull/454/files#diff-1da09bec2a4b26b257932b06c7ea839cc44492e089bd8276e4ae348bc6772b78R20

if i dont do it and if there are extra values, it will mess with the hash , thats my understanding .
Thats the reason i did the specific check for now and then probably change it once we have the clean-up logic for revisions.
WDYT?

pkg/hash/hash_test.go
```diff
@@ -82,7 +101,13 @@ func TestTierHashMatches(t *testing.T) {
...
 					TemplateRef: "base1ns-clusterresources-e0e1f34-e0e1f34",
 				},
+				SpaceRoles: map[string]toolchainv1alpha1.NSTemplateTierSpaceRole{
+					"admin": {
+						TemplateRef: "base1ns-admin-123456abc",
+					},
+				},
 			},
+			Status: Status,
 		}
```
MatousJobanek, line 111: for the purpose of the unit-test, there is no need to define the spec of the NSTemplateTier, rather the opposite - not defining it it will ensure that the spec is not used for the calculation of the hash:
```go
		tier := &toolchainv1alpha1.NSTemplateTier{
			Status: toolchainv1alpha1.NSTemplateTierStatus{
				Revisions: map[string]string{
					"base1ns-dev-aeb78eb-aeb78eb":              "base1ns-dev-aeb78eb-aeb78eb",
					"base1ns-clusterresources-e0e1f34-e0e1f34": "base1ns-clusterresources-e0e1f34-e0e1f34",
					"base1ns-admin-123456abc":                  "base1ns-admin-123456abc",
				},
			},
		}
```

pkg/hash/hash_test.go
```diff
@@ -113,6 +143,7 @@ func TestTierHashMatches(t *testing.T) {
 					TemplateRef: "base1ns-clusterresources-e0e1f34-e0e1f34",
 				},
 			},
+			Status: Status,
```
MatousJobanek, line 146: the same, let's drop the spec part

pkg/hash/hash_test.go
```diff
@@ -143,6 +174,7 @@ func TestTierHashMatches(t *testing.T) {
...
```
MatousJobanek, line 177: here too

pkg/test/tier/templateRefs.go
```diff
@@ -8,14 +8,11 @@ import (
...
-	for _, ns := range tier.Spec.Namespaces {
-		refs = append(refs, ns.TemplateRef)
-	}
-	if tier.Spec.ClusterResources != nil {
-		refs = append(refs, tier.Spec.ClusterResources.TemplateRef)
+	for _, rev := range tier.Status.Revisions {
+		refs = append(refs, rev)
 	}
 	sort.Strings(refs)
 	m, err := json.Marshal(templateRefs{Refs: refs})
```
MatousJobanek, line 18: I know that you didn't introduce these functions, but I'm wondering if there is any difference compared to the functions that are in the `pkg/hash`. Couldn't we remove them?

fbm3307: I did notice it, was waiting to remove the refernece first from e2e and then remove it from here.. but as discussed, i have removed it now .

pkg/hash/hash_test.go
fbm3307: you are right, removed the spec part

MatousJobanek (approved):

mfrancisc (approved):
Nice Job 👍

PR #463, opened 2025-03-12
Adding the api replace/dependency in dependent repos

fbm3307:
This change is to add the api repo replace along with toolchain common in the dependency check to cover the case where there are changes to api and to toolchain common both, so the verify script can check it correctly

### DIFF
--- a/scripts/verify-replace.sh
+++ b/scripts/verify-replace.sh
@@ -5,6 +5,7 @@ GH_BASE_URL_KS=https://github.com/kubesaw/
 GH_BASE_URL_CRT=https://github.com/codeready-toolchain/
 declare -a REPOS=("${GH_BASE_URL_KS}ksctl" "${GH_BASE_URL_CRT}host-operator" "${GH_BASE_URL_CRT}member-operator" "${GH_BASE_URL_CRT}registration-service" "${GH_BASE_URL_CRT}toolchain-e2e")
 C_PATH=${PWD}
+API_REPLACE=$(grep 'replace' go.mod)
 ERROR_REPO_LIST=()
 ERROR_FILE_LIST=()
 STD_OUT_FILE_LIST=()
@@ -39,6 +40,7 @@ do
     fi
     echo "Initiating 'go mod replace' of current toolchain common version in dependent repos"
     go mod edit -replace github.com/codeready-toolchain/toolchain-common=${C_PATH}
+    echo ${API_REPLACE} >> go.mod
     make verify-dependencies 2> >(tee ${err_file}) 1> >(tee ${std_out_file})
     rc=$?
     if [ ${rc} -ne 0 ]; then


scripts/verify-replace.sh
```diff
@@ -39,6 +40,7 @@ do
     fi
     echo "Initiating 'go mod replace' of current toolchain common version in dependent repos"
     go mod edit -replace github.com/codeready-toolchain/toolchain-common=${C_PATH}
+    echo ${API_REPLACE} >> go.mod
```
metlos, line 43: I am no expert on `go.mod`, but how does this accomplish to add the api to the list of replaced modules?

scripts/verify-replace.sh
```diff
@@ -5,6 +5,7 @@ GH_BASE_URL_KS=https://github.com/kubesaw/
 GH_BASE_URL_CRT=https://github.com/codeready-toolchain/
 declare -a REPOS=("${GH_BASE_URL_KS}ksctl" "${GH_BASE_URL_CRT}host-operator" "${GH_BASE_URL_CRT}member-operator" "${GH_BASE_URL_CRT}registration-service" "${GH_BASE_URL_CRT}toolchain-e2e")
 C_PATH=${PWD}
+API_REPLACE=$(grep 'replace' go.mod)
```
MatousJobanek, line 8: You cannot simply grep `go.mod` for any replace clause - there could be multiple of them and in two formats.
During our 1-1, I pointed you to the way of checking replace clauses in toolchain-cicd repo:
https://github.com/codeready-toolchain/toolchain-cicd/blob/d5e8b3cdde0dd75c1de1735c22d2fb8adf2d37f1/gomod-check/action.yml#L24
is there a reason why you didn't use it?

drpaneas:
```
go mod edit -json | jq '.Replace'
```

thank me later :D

fbm3307: @MatousJobanek i did try using that, the main concern for me was to construct a replace of api via script and add it to other repos after fetching it through the above command. somehow i thought of simplifying it and i just forgot about 2nd type of format of replace . i presumed, in toolchain-common, we will have only api replace clause atleast in the PRs where we require this check.🤦🏻‍♀️

fbm3307: @drpaneas thanks for this, checking if i can use that 👍

scripts/verify-replace.sh
drpaneas: You need to think of ways this might break. e.g. what you might hold in API_REPLACE is gibberish, so a check with `go list -f` (or sth like this) is mandatory to ensure you are going to add a valid entry in go.mod. Or this line might already be there, so don't add it twice. I would prefer using Go tooling for this, that parses the file, rather than treating is a plain txt. Your solution would be fine if you have plain text files, but there you have a format, a valid lexer and parser. Adding things with echo in the very end, it's a no-no ;) I mean, it's good enough for testing your idea, but the implementation needs some love :)

fbm3307: i have updated the logic now PTAL

drpaneas (changes requested):
Thanks for addressing the comments. I left some more based on the new version you pushed ;)

scripts/verify-replace.sh
```diff
@@ -5,6 +5,8 @@ GH_BASE_URL_KS=https://github.com/kubesaw/
 GH_BASE_URL_CRT=https://github.com/codeready-toolchain/
 declare -a REPOS=("${GH_BASE_URL_KS}ksctl" "${GH_BASE_URL_CRT}host-operator" "${GH_BASE_URL_CRT}member-operator" "${GH_BASE_URL_CRT}registration-service" "${GH_BASE_URL_CRT}toolchain-e2e")
 C_PATH=${PWD}
+API_PATH=$(go mod edit -json | jq '.Replace[]?|select(.Old.Path=="github.com/codeready-toolchain/api")|.New.Path' -r)
```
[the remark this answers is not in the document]
metlos: We list the two in `toolchain-e2e/required_tools.adoc`. Granted, that's not the ideal place to do that.

I agree that we should check for the presence of those tools before using them but the common practice across the codebase is not to. `jq` is used in `toolchain-e2e`, `member-operator`, is part of the GH actions (the go.mod check), is being added to the CI image and is required to upload coverage to codecov.

So IMHO, checking for the presence of the prerequisites is a good thing but lies outside of the scope of this PR. I took the liberty to open a new issue for that: https://issues.redhat.com/browse/KUBESAW-270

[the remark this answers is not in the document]
MatousJobanek: yeah, we rely on `jq` and `go` at many places, let's keep it outside of this PR and script

scripts/verify-replace.sh
```diff
@@ -39,6 +41,9 @@ do
     fi
     echo "Initiating 'go mod replace' of current toolchain common version in dependent repos"
     go mod edit -replace github.com/codeready-toolchain/toolchain-common=${C_PATH}
+    if [[ -n "${API_PATH}" ]]; then #check if there is any api replace in toolchain common - propogate the same to other repos along with tc
```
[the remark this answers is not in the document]
fbm3307: Updated it

scripts/verify-replace.sh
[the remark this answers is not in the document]
drpaneas: My "context" is only this script, so if you say this is handled elsewhere, before this script runs, then it's fine, you can disregard my comment on this ;)

drpaneas (commented):
/lgtm thanks for addressing my comments

Just a cosmetic thing, if you want your scripts to look polished, I would suggest you to do `# comment` vs `#comment`. This is not a veto request, just my OCD kicks in (see Google Bash Script guide: https://google.github.io/styleguide/shellguide.html#s4.3-implementation-comments) 

Can you provide a smal 1,2,3 steps how to test this so I can check it in my laptop?

fbm3307:
> Just a cosmetic thing, if you want your scripts to look polished, I would suggest you to do `# comment` vs `#comment`. This is not a veto request, just my OCD kicks in (see Google Bash Script guide: https://google.github.io/styleguide/shellguide.html#s4.3-implementation-comments)
> 
> Can you provide a smal 1,2,3 steps how to test this so I can check it in my laptop?

@drpaneas Thanks for reviewing i have updated it with comments and also added the steps or rather how to run it, in the script itself.
I would just repeat it here, you can have something changed in tc of your local and just run the script/make command `make verify-replace-run`. rest the script will take care and return with the result. 
basically it replaces tc to all the other dependent repos and runs go mod tidy, go vet, go build, go test, go-lint etc and display the result, so if error is there in any one repo , it will display the repo name and error and this will tell us if that repo needs changes.

Also you would need to approve the PR .

scripts/verify-replace.sh
```diff
@@ -1,17 +1,29 @@
+# This script is written to check whether any changes to toolchain-common, 
+# related API changes need update/changes in other related repos or not
+# To run this script there is make command 'make verify-replace-run' or 
+# you can directly run this script.
+# if you have changes in toolchain common, run this scipt and it will
```
rsoaresd, line 5:
```suggestion
# if you have changes in toolchain-common, run this script and it will
```

scripts/verify-replace.sh
```diff
@@ -38,7 +50,14 @@ do
         continue
     fi
     echo "Initiating 'go mod replace' of current toolchain common version in dependent repos"
-    go mod edit -replace github.com/codeready-toolchain/toolchain-common=${C_PATH}
+    go mod edit -replace ${TC_PATH}=${C_PATH}
+    # we are only fetching api replace and hence 
+    # check if there is any api replace in toolchain common - if there is
+    # propogate the same to other repos along with toolchaincommon replace
```
rsoaresd, line 56:
```suggestion
    # propogate the same to other repos along with toolchain-common replace
```

rsoaresd (commented):
Nice work 🚀 Just minor comment suggestions:

drpaneas:
> Also you would need to approve the PR .

I would need to test it first, and to do that I would require a lot of hand-holding, since I am not into how you guys work with the repositories and what's in and what's not. That means unless you give me a detailed guide (e.g. Step 1: Clone this repo, Step 2: do this, Step 3: you should see that, etc) then I cannot approve it. My intention was to flag some dangerous code out, I will pass the approver's torch to the people who take part in this marathon :)

scripts/verify-replace.sh
fbm3307: Updated it Thanks

drpaneas (commented):
Very nice! I would only suggest to ensure that both `API_REPLACE_PATH` and `API_REPLACE_VERSION` are valid and non-empty before attempting the API replace. This can prevent potential issues when the replacement directive isn’t found.

metlos (approved):

PR #468, opened 2025-03-14
Implement a new apply client based on SSA.

metlos:
This is not used anywhere currently. We will start to "roll it out" in various places one by one.

### DIFF
--- a/pkg/client/ssa_client.go
+++ b/pkg/client/ssa_client.go
@@ -1,0 +1,162 @@
+package client
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ApplyClient the client to use when creating or updating objects
+type SsaApplyClient struct {
+	client.Client
+}
+
+// NewSsaApplyClient returns a new ApplyClient
+func NewSsaApplyClient(cl client.Client) *SsaApplyClient {
+	return &SsaApplyClient{
+		Client: cl,
+	}
+}
+
+type ssaApplyObjectConfiguration struct {
+	owner           metav1.Object
+	newLabels       map[string]string
+	determineUpdate bool
+	skipIf          func(client.Object) bool
+}
+
+func newSsaApplyObjectConfiguration(options ...SsaApplyObjectOption) ssaApplyObjectConfiguration {
+	config := ssaApplyObjectConfiguration{}
+	for _, apply := range options {
+		apply(&config)
+	}
+	return config
+}
+
+// SsaApplyObjectOption an option when creating or updating a resource
+type SsaApplyObjectOption func(*ssaApplyObjectConfiguration)
+
+// SsaSetOwner sets the owner of the resource (default: `nil`)
+func SsaSetOwner(owner metav1.Object) SsaApplyObjectOption {
+	return func(config *ssaApplyObjectConfiguration) {
+		config.owner = owner
+	}
+}
+
+// SkipIf will cause the apply function skip the update of the object if
+// the provided function returns true. The supplied object is guaranteed to
+// have its GVK set.
+func SkipIf(test func(client.Object) bool) SsaApplyObjectOption {
+	return func(config *ssaApplyObjectConfiguration) {
+		config.skipIf = test
+	}
+}
+
+// DetermineUpdate instructs the ApplyObject function to truly test whether the object
+// was updated or not during the apply operation. By default, the function always returns
+// true on success which is more efficient and sufficient in most circumstances.
+func DetermineUpdate(value bool) SsaApplyObjectOption {
+	return func(config *ssaApplyObjectConfiguration) {
+		config.determineUpdate = value
+	}
+}
+
+// EnsureLabels makes sure that the provided labels are applied to the object even if
+// the supplied object doesn't have them set.
+func EnsureLabels(labels map[string]string) SsaApplyObjectOption {
+	return func(config *ssaApplyObjectConfiguration) {
+		config.newLabels = labels
+	}
+}
+
+// Configure sets the owner reference and merges the labels. Other options modify the logic
+// of apply function and therefore need to be checked manually.
+func (c *ssaApplyObjectConfiguration) Configure(obj client.Object, s *runtime.Scheme) error {
+	if c.owner != nil {
+		if err := controllerutil.SetControllerReference(c.owner, obj, s); err != nil {
+			return err
+		}
+	}
+	MergeLabels(obj, c.newLabels)
+
+	return nil
+}
+
+// ApplyObject creates the object if is missing and if the owner object is provided, then it's set as a controller reference.
+// If the objects exists then when the spec content has changed (based on the content of the annotation in the original object) then it
+// is automatically updated. If it looks to be same then based on the value of forceUpdate param it updates the object or not.
+// The return boolean says if the object was either created or updated (`true`). If nothing changed (ie, the generation was not
+// incremented by the server), then it returns `false`.
+//
+// NOTE: the return boolean IS ALWAYS TRUE on success by default. This is much more efficient. If you truly need
+// to determine whether the object has been updated or not (which is not the case anywhere in the codebase but the tests),
+// you can use the DetermineUpdate() option.
+func (c SsaApplyClient) ApplyObject(ctx context.Context, obj client.Object, options ...SsaApplyObjectOption) (bool, error) {
+	config := newSsaApplyObjectConfiguration(options...)
+	if err := config.Configure(obj, c.Scheme()); err != nil {
+		return false, err
+	}
+
+	if err := prepareForSSA(obj, c.Scheme()); err != nil {
+		return false, err
+	}
+
+	if config.skipIf != nil && config.skipIf(obj) {
+		return false, nil
+	}
+
+	updated := true
+	var orig client.Object
+	if config.determineUpdate {
+		orig = obj.DeepCopyObject().(client.Object)
+		if err := c.Get(ctx, client.ObjectKeyFromObject(obj), orig); err != nil && !apierrors.IsNotFound(err) {
+			return false, err
+		}
+	}
+
+	if err := c.Patch(ctx, obj, client.Apply, client.FieldOwner("kubesaw"), client.ForceOwnership); err != nil {
+		return false, fmt.Errorf("unable to patch '%s' called '%s' in namespace '%s': %w", obj.GetObjectKind().GroupVersionKind(), obj.GetName(), obj.GetNamespace(), err)
+	}
+
+	if config.determineUpdate {
+		updated = obj.GetGeneration() != orig.GetGeneration()
+	}
+
+	return updated, nil
+}
+
+func prepareForSSA(obj client.Object, scheme *runtime.Scheme) error {
+	gvk, err := apiutil.GVKForObject(obj, scheme)
+	if err != nil {
+		return err
+	}
+	obj.GetObjectKind().SetGroupVersionKind(gvk)
+	obj.SetManagedFields(nil)
+
+	return nil
+}
+
+// Apply applies the objects, ie, creates or updates them on the cluster
+// returns `true, nil` if at least one of the objects was created or modified,
+// `false, nil` if nothing changed, and `false, err` if an error occurred
+//
+// NOTE: this is only used in tests
+func (c SsaApplyClient) Apply(ctx context.Context, toolchainObjects []client.Object, opts ...SsaApplyObjectOption) (bool, error) {
+	createdOrUpdated := false
+	for _, toolchainObject := range toolchainObjects {
+		result, err := c.ApplyObject(ctx, toolchainObject, opts...)
+		if err != nil {
+			return createdOrUpdated, errors.Wrapf(err, "unable to create resource of kind: %s, version: %s", toolchainObject.GetObjectKind().GroupVersionKind().Kind, toolchainObject.GetObjectKind().GroupVersionKind().Version)
+		}
+		createdOrUpdated = createdOrUpdated || result
+	}
+	return createdOrUpdated, nil
+}
--- a/pkg/client/ssa_client_test.go
+++ b/pkg/client/ssa_client_test.go
@@ -1,0 +1,175 @@
+package client_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/codeready-toolchain/toolchain-common/pkg/client"
+	"github.com/codeready-toolchain/toolchain-common/pkg/test"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func TestSsaClient(t *testing.T) {
+	t.Run("ApplyObject", func(t *testing.T) {
+		t.Run("creates", func(t *testing.T) {
+			// given
+			cl, acl := NewTestSsaApplyClient(t)
+			obj := &corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "obj",
+					Namespace: "default",
+				},
+			}
+
+			// when
+			_, err := acl.ApplyObject(context.TODO(), obj)
+			require.NoError(t, err)
+
+			// then
+			inCluster := &corev1.ConfigMap{}
+			require.NoError(t, cl.Get(context.TODO(), runtimeclient.ObjectKeyFromObject(obj), inCluster))
+		})
+		t.Run("updates", func(t *testing.T) {
+			// given
+			obj := &corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "obj",
+					Namespace: "default",
+				},
+				Data: map[string]string{"a": "b"},
+			}
+			cl, acl := NewTestSsaApplyClient(t, obj)
+
+			updated := obj.DeepCopy()
+			updated.Data["a"] = "c"
+
+			// when
+			_, err := acl.ApplyObject(context.TODO(), updated)
+			require.NoError(t, err)
+			inCluster := &corev1.ConfigMap{}
+			require.NoError(t, cl.Get(context.TODO(), runtimeclient.ObjectKeyFromObject(obj), inCluster))
+
+			// then
+			assert.Equal(t, "c", inCluster.Data["a"])
+		})
+		t.Run("SetOwner", func(t *testing.T) {
+			// given
+			owner := &corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "owner",
+					Namespace: "default",
+				},
+			}
+
+			obj := &corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "owned",
+					Namespace: "default",
+				},
+			}
+			cl, acl := NewTestSsaApplyClient(t, owner, obj)
+
+			// when
+			_, err := acl.ApplyObject(context.TODO(), obj, client.SsaSetOwner(owner))
+			require.NoError(t, err)
+			inCluster := &corev1.ConfigMap{}
+			require.NoError(t, cl.Get(context.TODO(), runtimeclient.ObjectKeyFromObject(obj), inCluster))
+
+			// then
+			require.Len(t, inCluster.OwnerReferences, 1)
+			assert.Equal(t, "ConfigMap", inCluster.OwnerReferences[0].Kind)
+			assert.Equal(t, "owner", inCluster.OwnerReferences[0].Name)
+		})
+		t.Run("EnsureLabels", func(t *testing.T) {
+			// given
+			obj := &corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "obj",
+					Namespace: "default",
+				},
+			}
+			cl, acl := NewTestSsaApplyClient(t, obj)
+
+			// when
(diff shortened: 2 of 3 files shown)


pkg/client/ssa_client.go
```diff
@@ -0,0 +1,218 @@
...
+
+	return nil
+}
+
+// ApplyObject creates the object if is missing and if the owner object is provided, then it's set as a controller reference.
+// If the objects exists then when the spec content has changed (based on the content of the annotation in the original object) then it
+// is automatically updated. If it looks to be same then based on the value of forceUpdate param it updates the object or not.
+// The return boolean says if the object was either created or updated (`true`). If nothing changed (ie, the generation was not
+// incremented by the server), then it returns `false`.
+//
```
MatousJobanek, line 126: this looks like a copy-paste of the already existing ApplyObject method

pkg/client/ssa_client.go
```diff
@@ -0,0 +1,218 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ApplyClient the client to use when creating or updating objects. It uses SSA to apply the objects
+// to the cluster and takes care of migrating the objects from ordinary "CRUD" flow to SSA.
+type SsaApplyClient struct {
```
MatousJobanek, line 23: I haven't found any fancy "create" function (or anything similar) that could be called without constructing the object "manually".  Something similar to the already existing `NewApplyClient` or `ApplyUnstructuredObjectsWithNewLabels`

pkg/client/ssa_client.go
```diff
@@ -0,0 +1,218 @@
...
+
+	return updated, nil
+}
+
+func prepareForSSA(obj client.Object, scheme *runtime.Scheme) error {
+	gvk, err := apiutil.GVKForObject(obj, scheme)
+	if err != nil {
+		return err
+	}
+	obj.GetObjectKind().SetGroupVersionKind(gvk)
```
MatousJobanek, line 179: what if the GVK of the object is not in the scheme? We cannot expect that all GVKs of all resources defined in the templates would be present in the scheme (eg currently the appstudio API).
TBH, in most cases (when the object is taken from a template) the GVK should be already present in the object. Couldn't we simply check if it's already set or not, and try to set it from the scheme only if it's not set in the object already?

pkg/client/ssa_client.go
```diff
@@ -0,0 +1,218 @@
...
+	if config.skipIf != nil && config.skipIf(obj) {
+		return false, nil
+	}
+
+	updated := true
+
+	// NOTE: once the SSA migration is not needed anymore, read the orig conditionally only when config.determineUpdate is true.
+
+	// this cannot fail - we would have failed previously already because the scheme wouldn't have known the object's kind
+	origRo, _ := c.Client.Scheme().New(obj.GetObjectKind().GroupVersionKind())
```
MatousJobanek, line 149: The same thing again - we shouldn't rely on the fact that he object's kind will be always present in the scheme.
I would rather rely on Unstructured type in this case

pkg/client/ssa_client.go
```diff
@@ -0,0 +1,218 @@
...
+	if err := csaupgrade.UpgradeManagedFields(obj, sets.New(c.NonSSAFieldOwner), c.FieldOwner); err != nil {
+		return err
+	}
+
+	return c.Client.Update(ctx, obj)
+}
+
+func isSsaMigrationNeeded(obj client.Object, nonSsaOwner string) bool {
+	for _, mf := range obj.GetManagedFields() {
+		if mf.Manager == nonSsaOwner {
```
MatousJobanek, line 212: I'm wondering how this would be different compared to a case when a user deletes the managedFields completely. If it's the same, then we would need to keep the migration forever.

pkg/client/ssa_client.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	// NOTE: remove this once this code is used for all "apply-style" functionality and has updated all the objects in the cluster.
+	if isSsaMigrationNeeded(orig, c.NonSSAFieldOwner) {
+		if err := c.MigrateToSSA(ctx, orig); err != nil {
+			return false, err
+		}
+	}
+
+	if err := c.Client.Patch(ctx, obj, client.Apply, client.FieldOwner(c.FieldOwner), client.ForceOwnership); err != nil {
```
MatousJobanek, line 163: do I understand it correctly that in case of migration, we would need to do update and patch the resource immediately after each other?
Could the Patch call cause any conflicts? I see that it's not taking the resourceVersion of the updated object.

pkg/client/ssa_client.go
```diff
@@ -0,0 +1,218 @@
...
+		return false, fmt.Errorf("unable to patch '%s' called '%s' in namespace '%s': %w", obj.GetObjectKind().GroupVersionKind(), obj.GetName(), obj.GetNamespace(), err)
+	}
+
+	if config.determineUpdate {
+		updated = obj.GetGeneration() != orig.GetGeneration()
+	}
```
MatousJobanek, line 169: we could also check labels & annotations

MatousJobanek (commented):
Looks great

pkg/client/ssa_client.go
MatousJobanek: never mind, I misread the code. it actually checks if the nonSsaOwner was used before or not.

pkg/client/ssa_client.go
```diff
@@ -0,0 +1,218 @@
...
+		result, err := c.ApplyObject(ctx, toolchainObject, opts...)
+		if err != nil {
+			return createdOrUpdated, errors.Wrapf(err, "unable to create resource of kind: %s, version: %s", toolchainObject.GetObjectKind().GroupVersionKind().Kind, toolchainObject.GetObjectKind().GroupVersionKind().Version)
+		}
+		createdOrUpdated = createdOrUpdated || result
+	}
+	return createdOrUpdated, nil
+}
+
+func (c *SsaApplyClient) MigrateToSSA(ctx context.Context, obj client.Object) error {
```
MatousJobanek, line 202: btw, couldn't we (as part of the migration) also drop the `toolchain.dev.openshift.com/last-applied-configuration` annotation?

pkg/client/ssa_client.go
```diff
@@ -0,0 +1,218 @@
...
+	if err := csaupgrade.UpgradeManagedFields(obj, sets.New(c.NonSSAFieldOwner), c.FieldOwner); err != nil {
```
MatousJobanek, line 203: I'm wondering, would we actually need to do the migration when we make sure that the new `FieldOwner` is not the same as the `NonSSAFieldOwner`? Currently, the owner is something like `host-operator`, `member-operator`, or `ksctl`. What if we modify the owner by adding `kubesaw-` prefix there?

pkg/client/ssa_client.go
metlos: If the GVK is not in the scheme, the call to client.Patch, Create, Get, ... will fail regardless as far as I am aware. I think we would have to work with a dynamic client to support that and then it would become problem of determining the GVR (group version resource) from the object itself, which is not possible in the general sense because the resource string is different from kind (usually, its just its plural, but we've seen with kubevirt and their sub-resources that stuff can get complex).

pkg/client/ssa_client.go
metlos: IMHO, the kind needs to be in the scheme even for the unstructured objects. But let me experiment with that.

pkg/client/ssa_client.go
metlos: I'm hoping we will drop this check altogether and will just not bother trying to determine if there was an update or not. But I kept this here to lower the impact on the tests that do work with this value. In all other places that I am aware, the result of this is just used for logging purposes.

pkg/client/ssa_client.go
metlos: It is dropped automagically because we first do the migration and then apply the patch (that doesn't contain it).

pkg/client/ssa_client.go
metlos: We need to do the migration not because of a different owner, but because the operations that update the objects are recorded as `Update` in the managed fields. For all the "magic" of SSA to work, we need to only have `Apply` operations done by "our" owners. That's what the migration is doing, apart from the update to the owner.

I wanted to update the owner to a value that is hardcoded in the code rather than deduced from the name of the binary that contains the code.

pkg/client/ssa_client.go
metlos: yes, and it's intentional here, because that will remove any additional "cruft" from the resource that we applied to it using the old logic, like the `last-applied-configuration`.

pkg/client/ssa_client.go
metlos: All the fields need to be set on the struct, so I didn't see the need for the function actually.

pkg/client/ssa_client.go
metlos: Indeed, I will reword it.

pkg/client/ssa_client.go
MatousJobanek: > For all the "magic" of SSA to work, we need to only have Apply operations done by "our" owners.

but if we start forcing ownership of all fields with a new owner/manager with the apply logic, then it should just work.
What confuses me is how this would be different compared to a user updating the object (and the fields that are owned by us) on their own. As a result, the resource would have a new manager owning "our" fields, but our next patch (when we force the ownership of our fields) should override that. 

What I'm proposing is treating the already existing managers in the managedFields  (`host-operator`, `member-operator`, `ksctl`) as 3rd party entities we don't care about, and for all patches using SSA use new managers `kubesaw-host-operator`, `kubesaw-member-operator`, `kubesaw-ksctl` (or with `toolchain-*` prefix to match the api group). It would be very similar to the case when someone uses the `Update` operation with a random manager name.

pkg/client/ssa_client.go
metlos: Ok, I learned something today :) Fix incoming :)

pkg/client/ssa_client.go
MatousJobanek: I like the construct functions more as it's easier to use and it's clear what parameters needs to be set. It could technically also call `GetDefaultFieldOwner` automatically. Or do we expect that we would set something else for this field?

pkg/client/ssa_client.go
```diff
@@ -0,0 +1,218 @@
...
+
+// GetDefaultFieldOwner returns the default field owner that is applied if no explicit field owner is set.
+// This can be used to determine the field owner used by the non-SSA CRUD operations performed by
+// the kubernetes client.
+//
+// This value is obtained from the user agent header defined in the provided config, or, if it is not set,
+// from the default kubernetes user agent string.
+//
+// If the provided config is nil, the default kubernetes user agent is returned
+func GetDefaultFieldOwner(cfg *rest.Config) string {
```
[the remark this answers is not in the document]
MatousJobanek: never mind. I was wondering if the returned value could change over the time. Let's say that today it returns `host-operator`, but tomorrow (based on some changes in the config) it would return `toolchain-host-operator` or something like that. But I guess that the probability of such a change is almost equal to zero.

pkg/client/ssa_client.go
MatousJobanek: Yeah, that's what I wanted to propose actually, if there is no real usage of this returned value, then I would simply drop it.

pkg/client/ssa_client.go
metlos: will be fixed

pkg/client/ssa_client.go
metlos: Patch will not cause a conflict. Resource version doesn't have to be set - that's the general case with patches actually - they're supposed to work without getting the object first and therefore without knowing its resource version.

pkg/client/ssa_client.go
metlos: I supplied a function to initialize the client from a controller manager, maybe that'll become useful.

pkg/client/ssa_client.go
metlos: So imagine this situation (on a configmap data for simplicity):

The current state of the cluster is this:
```
key1 = value1 # owned by host-operator, using update
key2 = value2 # owned by host-operator, using update
```

Now, in the code, we SHOULD take into account the situation where the template changed and we're doing the migration.

The template contains this (notice the absence of `key2`):

```
key1 = value3
```

If we merely forced the patch with the new template, we'd end up in this situation:

```
key1 = value3 # owned by kube-saw-host-operator, using apply
key2 = value2 # owned by host-operator, using update
```

If we first do the migration, the data will first transfer to this:
```
key1 = value1 # owned by kubesaw-host-operator, using apply
key2 = value2 # owned by kubesaw-host-operator, using apply
```

and when then we apply the patch (that doesn't contain `key2`), we will see the correct result:

```
key1 = value3 # owned by kubesaw-host-operator, using apply
```

metlos: In another words, IMHO we cannot treat the legacy owners the same as 3rd party owners, because it was the legacy owners that applied our data. To ensure the consistency, we need to treat them specially.

This situation may be most prominent, IMHO, with `ksctl install` that can theoretically update long-outdated operators and chances are we've removed some field in the deployments in the meantime.

Which is something I didn't realize before (the long timeframes we have to operate with), so I think the question of when to remove the migration logic might not be that simple to answer.

pkg/client/ssa_client.go
metlos: Makes sense, let's just bite the bullet and keep the production code as simple as possible.

MatousJobanek: :+1:

pkg/test/client.go
```diff
@@ -31,6 +33,64 @@ func NewFakeClient(t T, initObjs ...client.Object) *FakeClient {
...
+			var err error
+			shouldUpdateGeneration, err = isGenerationChangeNeeded(orig, dryRunObj)
+			if err != nil {
+				return err
+			}
+		}
+
+		if shouldUpdateGeneration {
+			obj.SetGeneration(orig.GetGeneration() + 1)
+		}
```
mfrancisc, line 88: maybe I'm wrong , but it looks like we could move this inside the if statement above, at line 83. That's the only place where the `shouldUpdateGeneration` can become `true`, unless I'm missing something.

pkg/client/ssa_client_test.go
```diff
@@ -0,0 +1,200 @@
...
+			updated := obj.DeepCopy()
+			updated.Data["a"] = "c"
+
+			// when
+			require.NoError(t, acl.ApplyObject(context.TODO(), updated))
+			inCluster := &corev1.ConfigMap{}
+			require.NoError(t, cl.Get(context.TODO(), runtimeclient.ObjectKeyFromObject(obj), inCluster))
+
+			// then
+			assert.Equal(t, "c", inCluster.Data["a"])
```
mfrancisc, line 58: not sure I'm following how we are testing updates - I mean the obj is not actually being first created ( with the fake client)  and than updated. You're just updating the "local" version  of the object. But I'm sure I'm missing something.

mfrancisc (commented):
Great Job 👏 

I have only few minor comments.

pkg/client/ssa_client.go
metlos: We will not see any conflicts, but theoretically should see obsolete fields still present in the cluster. But as discussed privately, this is very very unlikely - we will not face a situation described above, where we'd have a template deployed to a cluster first, then a template change and then apply using SSA.

There will not be a template/deployment change that would not be applied to any cluster prior to us introducing SSA.

pkg/client/ssa_client.go
```diff
@@ -0,0 +1,212 @@
...
+		return err
+	}
+
+	if err := prepareForSSA(obj, c.Client.Scheme()); err != nil {
+		return err
+	}
+
+	if config.skipIf != nil && config.skipIf(obj) {
+		return nil
+	}
```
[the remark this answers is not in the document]
metlos: This is intentional, so that the object is fully ready for SSA when we do the check - i.e. the object is ensured to have the GVK set on it. It is mentioned in the `SkipIf` doc comment.

This is so that the users of "skip if" can make decisions based on the kind of the object (and not have to care whether the supplied `client.Object` is unstructured or typed).

pkg/client/ssa_client_test.go
mfrancisc: ok, I see. I didn't saw that passing the obj to the constructor actually adds it to the fake client. 
I was expecting something like:
- I apply one object with some data in one test
- In a subtest I modify the object and apply it again and than I check that the object has all the changes

pkg/client/ssa_client.go
```diff
@@ -0,0 +1,212 @@
...
+
+	return nil
+}
+
+// Apply is a utility function that just calls `ApplyObject` in a loop on all the supplied objects.
+func (c *SsaApplyClient) Apply(ctx context.Context, toolchainObjects []client.Object, opts ...SsaApplyObjectOption) error {
+	for _, toolchainObject := range toolchainObjects {
+		if err := c.ApplyObject(ctx, toolchainObject, opts...); err != nil {
+			return errors.Wrapf(err, "unable to create resource of kind: %s, version: %s", toolchainObject.GetObjectKind().GroupVersionKind().Kind, toolchainObject.GetObjectKind().GroupVersionKind().Version)
+		}
```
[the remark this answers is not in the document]
mfrancisc: Makes sense. Just one idea. The current logic would stop at the first issue. But something we can address later if needed.

pkg/test/client.go
```diff
@@ -31,6 +33,63 @@ func NewFakeClient(t T, initObjs ...client.Object) *FakeClient {
...
+			// the fake client actively complains if it sees an SSA patch...
+			patch = client.Merge
+		}
+
+		if found {
+			dryRunOpts := make([]client.PatchOption, len(opts)+1)
+			copy(dryRunOpts, opts)
+			dryRunOpts[len(opts)] = client.DryRunAll
+			dryRunObj := obj.DeepCopyObject().(client.Object)
+			if err := fakeClient.Client.Patch(ctx, dryRunObj, patch, dryRunOpts...); err != nil {
```
MatousJobanek, line 74: can you explain why this is needed?

pkg/test/client.go
```diff
@@ -31,6 +33,63 @@ func NewFakeClient(t T, initObjs ...client.Object) *FakeClient {
...
+
+			var err error
+			shouldUpdateGeneration, err := isGenerationChangeNeeded(orig, dryRunObj)
+			if err != nil {
+				return err
+			}
+
+			if shouldUpdateGeneration {
+				obj.SetGeneration(orig.GetGeneration() + 1)
+			}
```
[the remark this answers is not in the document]
metlos: This is still needed because we expect the generation to change if the patch updates the spec of the object. The patch may not know that it did that but it should still happen (and the tests check for it).

pkg/client/ssa_client_test.go
```diff
@@ -0,0 +1,169 @@
...
+			assert.Equal(t, "owner", inCluster.OwnerReferences[0].Name)
+		})
+		t.Run("EnsureLabels", func(t *testing.T) {
+			// given
+			obj := &corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "obj",
+					Namespace: "default",
+				},
+			}
```
MatousJobanek, line 94: can you add a few already-existing labels and verify that they are merged properly?

pkg/test/client.go
```diff
@@ -31,6 +33,63 @@ func NewFakeClient(t T, initObjs ...client.Object) *FakeClient {
 	return &FakeClient{Client: cl, T: t}
 }
 
+func FakeSSA(fakeClient *FakeClient) {
+	fakeClient.MockPatch = FakeSSAPatchFunction(fakeClient)
```
MatousJobanek, line 37: why not running this FakeSSAPatchFunction by default as part of the default FakeClient implementation? In the same way as it is currently done for the `Update` function.

pkg/client/ssa_client.go
```diff
@@ -0,0 +1,178 @@
...
+	if err := prepareForSSA(obj, c.Client.Scheme()); err != nil {
+		return composeError(obj, fmt.Errorf("failed to prepare the object for SSA: %w", err))
+	}
+
+	if config.skipIf != nil && config.skipIf(obj) {
+		return nil
+	}
+
+	if err := c.Client.Patch(ctx, obj, client.Apply, client.FieldOwner(c.FieldOwner), client.ForceOwnership); err != nil {
+		return composeError(obj, err)
```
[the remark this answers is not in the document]
metlos: This will work. `errors.IsNotFound()` will work on the composed error because we wrap the error and the `k8s.io/apimachinery/pkg/api/errors` searches the "causes" to find the error originating in kubernetes.

pkg/test/client.go
```diff
@@ -31,6 +33,63 @@ func NewFakeClient(t T, initObjs ...client.Object) *FakeClient {
...
+}
+
+func FakeSSAPatchFunction(fakeClient *FakeClient) func(ctx context.Context, obj client.Object, patch client.Patch, opts ...client.PatchOption) error {
+	return func(ctx context.Context, obj client.Object, patch client.Patch, opts ...client.PatchOption) error {
+		// fake client doesn't support SSA yet, so we have to be creative here and try to mock it out. Hopefully, SSA will be merged soon and we will
+		// be able to remove this.
```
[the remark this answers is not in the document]
metlos: I added a link to the issue and PR in the controller-runtime to a comment in this function.

pkg/test/client.go
metlos: I added a comment explaning why this is needed.

pkg/client/ssa_client_test.go
metlos: Added a test for merging as well as adding to empty labels.

pkg/client/ssa_client.go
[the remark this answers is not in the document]
MatousJobanek: OK. But that reminds me that I haven't seen any unit tests for the error cases in your PR. Can you please add them and also verify that `errors.IsNotFound()` works as expected?

pkg/client/ssa_client.go
```diff
@@ -0,0 +1,149 @@
...
+		config.owner = owner
+	}
+}
+
+// SkipIf will cause the apply function skip the update of the object if
+// the provided function returns true. The supplied object is guaranteed to
+// have its GVK set.
+func SkipIf(test func(client.Object) bool) SSAApplyObjectOption {
+	return func(config *ssaApplyObjectConfiguration) {
+		config.skipIf = test
```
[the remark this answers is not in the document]
metlos: I disagree, the provided function tests whether the object should be skipped, so the name seems appropriate to me.

MatousJobanek (approved):
cool, thanks for the additional changes :+1: :rocket: :1st_place_medal:

PR #475, opened 2025-03-28
update the golangci-lint version in the verify-dependencies GH action

metlos:
update the golangci-lint version in the verify-dependencies GH action to be in line with the version in linters.yaml


mfrancisc (approved):
Thank you!

MatousJobanek (approved):

PR #479, opened 2025-04-25
Add SSA migration and exclusive ownership capabilities to the SSA client

metlos:
The migration from CSA (client-side-apply) to SSA is in the end needed (we first thought we could do without it), because the "dangling" update entries in the managed fields could prevent the automatic field removal in some future update (that could happen many operator versions later, as long as the object still survives from the times where our operators didn't use SSA for creating/updating objects in the cluster).

Additionally, this PR introduces the ability to ensure the exclusive ownership of all the fields present on the object template. This is only required for objects that the end users have write-access to. Having write-access to the object enables the user to fiddle with the managed fields, too, using which they could "prepare" the managed fields to a state such that any patches using the normal SSA could leave behind fields that should no longer be there. This is off by default, because it is not needed for objects that should only be writable to operators (either ours or other) which are implicitly trusted.

### DIFF
--- a/pkg/client/ssa_client.go
+++ b/pkg/client/ssa_client.go
@@ -4,9 +4,14 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strings"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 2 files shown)


metlos:
I'm closing this and will open 2 separate PRs for migration and exclusive ownership.

metlos:
https://github.com/codeready-toolchain/toolchain-common/pull/480 is just the SSA migration. I will reopen the PR for the exclusive ownership if and when we decide to use it.

PR #480, opened 2025-04-28
Add migration support to the SSA client

metlos:
The migration from CSA (client-side-apply) to SSA is in the end needed (we first thought we could do without it), because the "dangling" update entries in the managed fields could prevent the automatic field removal in some future update (that could happen many operator versions later, as long as the object still survives from the times where our operators didn't use SSA for creating/updating objects in the cluster).

### DIFF
--- a/pkg/client/ssa_client.go
+++ b/pkg/client/ssa_client.go
@@ -4,9 +4,14 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strings"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/util/csaupgrade"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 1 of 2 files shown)


pkg/client/ssa_client.go
```diff
@@ -24,21 +29,59 @@ type SSAApplyClient struct {
...
+	MigrateSSAByDefault bool
+
+	// NonSSAFieldOwner should be set to the same value as the user agent used by the provided Kubernetes client
+	// or to the value of the explicit field owner that the calling code used to use with the normal CRUD operations
+	// (highly unlikely and not the case in our codebase).
+	//
+	// The user agent can be obtained from the REST config from which the client is constructed.
+	//
+	// The user agent in the REST config is usually empty, so there's no need to set it here either in that case.
+	NonSSAFieldOwner string
```
MatousJobanek, line 57: in your code in the previous PR, you originally had a code that was able to construct the owner name we should migrate from - couldn't we use that?
https://github.com/metlos/toolchain-common/blob/6ff87dfbb853a19475dff63b22a0e5a2f51300d9/pkg/client/ssa_client.go#L50-L62

pkg/client/ssa_client.go
```diff
@@ -111,6 +173,32 @@ func (c *SSAApplyClient) ApplyObject(ctx context.Context, obj client.Object, opt
...
+	}
+
+	if orig != nil {
+		oldFieldOwner := c.NonSSAFieldOwner
+		if len(oldFieldOwner) == 0 {
+			// this is how the kubernetes api server determines the default owner from the user agent
+			// The default user agent has the form of "name-of-binary/version information etc.".
+			// The owner is the first part of the UA unless explicitly specified in the request URI.
+			oldFieldOwner = strings.Split(rest.DefaultKubernetesUserAgent(), "/")[0]
+		}
```
MatousJobanek, line 192: related to my previous comment, do you think that we would need to se the owner name explicitly? Couldn't we rely on the default behavior only?

MatousJobanek (approved):
Nice :+1:

pkg/client/ssa_client.go
metlos: I wanted to have this option available though in case I missed something and we for example set the user agent of the kubernetes client to an explicit value somewhere in the code.

In that case, we'd need to set this to an explicit value, too.

pkg/client/ssa_client.go
metlos: We could, but the default behavior makes assumptions that I was not 100% sure we satisfy everywhere in the codebase, so I wanted to have an explicit option available for the case we find a place where the defaults would not be applicable.

The default behavior assumes that:
1. There is no user agent set explicitly in the config of any the kubernetes client.
2. We don't explicitly set the field owner with any of the existing CRUD operations.

MatousJobanek: >There is no user agent set explicitly in the config of any kubernetes client.
>We don't explicitly set the field owner with any of the existing CRUD operations with the kubernetes clients.

I'm not aware of any place that would change any of these explicitly, so everything should comply. 

Anyway, thanks for the explanation. It's fine to keep it there if not required to be set explicitly by default (only as a backup solution for some corner-cases).

PR #485, opened 2025-06-12
introduce a simple abstraction for easier handling of finalizers

metlos:
This can be used for simplification of the finalizer handling in the controllers.


metlos:
We will no go ahead with this. It's not bringing enough benefit and doesn't fit all the usecases.

PR #540, opened 2026-07-31
Extract owner idle actions into pkg/idler from member operator

alexeykazakov:
- Introduce `pkg/idler` as the shared library for idling Sandbox workload owners (scale/stop/patch/delete by kind).
- Expose `IdleOwner` for a single known owner and on-demand `IdleFromPod` (up to two known owners, no timeout policy, no user notifications).
- Parameterize ServingRuntime InferenceService cleanup via `Options.TimeoutSeconds` (no Idler CR dependency in common).

Related PR: https://github.com/codeready-toolchain/member-operator/pull/761

## Summary by CodeRabbit

* **New Features**
  * Added workload idling for supported resources, including scaling controllers to zero and stopping virtual machines.
  * Added cleanup for eligible, outdated inference services using configurable timeouts.
  * Added DeploymentConfig replica management and specialized handling for application, serving, and virtual machine resources.
  * Added pod-based ownership traversal with configurable secondary-owner policies.
  * Added optional deletion skipping, unsupported-resource handling, and aggregated action errors.
* **Tests**
  * Added comprehensive coverage for idling, ownership traversal, timeout filtering, errors, and cleanup behavior.


MatousJobanek (approved):

alexeykazakov:
@MatousJobanek more code is extracted from the member operator now with additional configuration options. Please take a look.

pkg/idler/idler.go
```diff
@@ -0,0 +1,244 @@
...
+
+	if opts.SecondOwnerPolicy == SecondOwnerAfterTimeout && shouldRecommendPodDelete(pod, opts.PodTimeoutSeconds, attempted) {
+		// Nothing remains attempted (e.g. top owner idled but remaining owners are deleting).
+		// Clear owner so the caller can delete the pod as a last resort.
+		return Result{PodDeleteRecommended: true}, errToReturn
+	}
+
+	if topOwnerKind == "" && fetchErr != nil {
+		return Result{}, fetchErr
+	}
```
MatousJobanek, line 195: this is slightly different compared to the previous logic - if nothing was idled and there was an error, the original code tried to delete the pod, which wouldn't do in this particular case if I'm not mistaken

alexeykazakov: Right. Previously (current logic) if we failed to fetch the owner of the pod due to some errors then we would delete the pod as a standalone pod. It has some pros and cons. One con is that it could hide some bugs in our code or if it was a temp error in the cluster then we would basically reset the pod idling timeout. With the new logic we always retry in case of an error while fetching the owner(s). I would keep the new logic. But if we see a legit case when an error should be treated as a standalone pod then we can always restore the old logic. WDYT?

